PR #6, opened 2021-09-06
Fixed ignoring error of zypper on the build for windows

### LINKED ISSUE #5: [Bug] Build for Windows fails on cmake phase due wxwidgets is not installed
Now we use zypper for installing some packages for cross compiling from http://download.opensuse.org/repositories/home:e9925248:mingw/openSUSE_Tumbleweed/home:e9925248:mingw.repo

Sometimes zypper fails and the `prepare-ubuntu-20.sh` ignores this error. It causes a dependency failure on the `cmake` phase that makes troubleshooting more complex.

It is better not to ignore this error and to fail on the `Install Dependencies` phase

oleg68:
Resolves #5 

This change
1. Disabled ignoring an error
2. Splitted one zypper command to two ones for decreasing probability of errors (zypper usually fails when downloading a large amount of files)


larspalo (approved):

rousseldenis (approved):
So, Agile and people are aware of changes. I like it

PR #18, opened 2021-09-08
[WIP] Sourceforge migration

rousseldenis:


rousseldenis:
@larspalo A first script.

Authentication is working, but not a first query yet (I discover graphql too).

oleg68:
1. This script is necessary only once. Should we place in the source tree?

    Earlier I added [scripts for converting svn history to the CHANGELOG.md](https://github.com/GrandOrgue/grandorgue/tree/master/build-scripts/for-changelog), because I was going to use it each time when new commits would appear in svn.  But now this script is not necessary and may be deleted.

2. Even so we deside to place this script, it's better not to create a new top-level subdirectory and to put it somewhere in https://github.com/oleg68/GrandOrgue/tree/master/build-scripts

rousseldenis:
> This script is necessary only once. Should we place it at all in the source tree?

This way we can work on it all together. But, indeed, it can disappear after migrated.

> 2\. Even so we deside to place this script, it's better not to create a new top-level subdirectory and to put it somewhere in https://github.com/oleg68/GrandOrgue/tree/master/build-scripts

I don't mind where to put it. It's a work in progress for now, but isn't it weird to put it 'build' ?

oleg68:
@rousseldenis OK, lets work with this PR together and then close this PR without merge.

oleg68:
Hi, @rousseldenis!

Did you managed to import SF discussions to your test repo?

larspalo:
I'm not at all familiar with python, but I've experimented a bit with the GraphQL API and found out a few things that we might need to think about.

- when we import the discussion comments we'll loose the user who actually made that comment, we might need to add a note in the body of the discussion/comment stating from whom the original content come.
- the Sourceforge forum can have multiple levels of responses nested, it seems that the GitHub forum cannot - from the original discussion content you can have comments and then replies in one level to the comments, or of course comment on the original discussion. This means that we need to flatten out the original Sourceforge content to match the GitHub format, which is no big loss in my opinion.
- it's not documented in the GrapQL API for discussions, https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions, how the attachments are handled. As the Sourceforge .json file contain links to the attachment on the Sourceforge, we could, if we don't find a way to import the attachment files, just append the link to the body.
- as the discussion ID and comment ID is set at time of creation, the script we create must actually fetch the relevant ID for the created discussion/comment to be used, no big problem as that can easily be returned by the mutation Payload.

There are likely more things to consider, but this is a start.

larspalo:
@rousseldenis @oleg68 Ok. I've taken another approach and used php to create and send the GraphQL needed with curl. I've got all the basics done, it's just the parsing logic of the Sourceforge .json that's left to do. For this my previous comment becomes quite urgent. Could you have a look at that, please. I could pretty soon have a working import script ready, I think.

rousseldenis:
> @rousseldenis @oleg68 Ok. I've taken another approach and used php to create and send the GraphQL needed with curl. I've got all the basics done, it's just the parsing logic of the Sourceforge .json that's left to do. For this my previous comment becomes quite urgent. Could you have a look at that, please. I could pretty soon have a working import script ready, I think.

Hi guys, sorry but past days were a little bit busy for me.

If you succeeded in migrating with php, that's cool.

> when we import the discussion comments we'll loose the user who actually made that comment, we might need to add a note in the body of the discussion/comment stating from whom the original content come.

Agree for the note (maybe 'Commented in previous Sourceforge forum by ...')

> This means that we need to flatten out the original Sourceforge content to match the GitHub format, which is no big loss in my opinion.

Yes, no problem to flatten answers.

> it's not documented in the GrapQL API for discussions, https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions, how the attachments are handled. As the Sourceforge .json file contain links to the attachment on the Sourceforge, we could, if we don't find a way to import the attachment files, just append the link to the body.

The link is the easiest to do it. Either, you need to upload attachment in github (it generates a unique url at that moment) before creating that link.


We can maybe create a private repo to test it and compare then the result ?

oleg68:
@larspalo  Some SourceForge users present in Github too. It would be nice to map them

larspalo:
> @larspalo Some SourceForge users present in Github too. It would be nice to map them

@oleg68 How can that be done easiest? From the .json backup file we get the author of the post as a string.

rousseldenis:
> @oleg68 How can that be done easiest? From the .json backup file we get the author of the post as a string.

The way I did the svn branch.

But you need to do it in two steps.

First, generate a text file with the list of authors (with unique entries) with their github login (first, you just make a duplicate of key/value  e.g: 'larspalo = <larspalo>'

Then, you have to modify by hand (that's the only way) the authors value if you find them with their github login with email adress (I think github needs it) in the <>.

Then, do your script with the text file as source for authors.

larspalo:
If I just add an @ before the author name, there's no insurance that it would map correctly... In my case and for Oleg it will, but many others might not exist or even be false == another person.

Let's be honest, I'm not thrilled by the idea of looking for users and verifying they are the right ones...

rousseldenis:
> If I just add an @ before the author name, there's no insurance that it would map correctly... In my case and for Oleg it will, but many others might not exist or even be false == another person.
> 
> Let's be honest, I'm not thrilled by the idea of looking for users and verifying they are the right ones...

Yes, I understand, that will be harsh to do.

As I said, maybe just the comment as suggested above should be sufficent for unknown users

oleg68:
@larspalo I suggest to create a dictionary and include there known users. Any other users will not be mapped and will rest in comments.

Known users:

    - Denis Roussel: rousseldenis
    - Jean Gabriel:  Jean-Gabriel1982
    - Laércio de Sousa:  lbssousa
    - Lars Palo:  larspalo
    - Mark Bugeja: Bumblebee001
    - Martin Koegler: e9925248
    - Nick Appleton: nickappleton
    - Oleg Samarin: oleg68
    - Paolo Gilardi: paolo2504

larspalo:
I've encountered a few problems which have delayed me, mostly the issue seems to have to do with some special characters (more specifically the double and single quotes in subject=title and text=body) and the json.

As the script (kind of) works, you can preview the imported forum at https://github.com/larspalo/grandorgue/discussions but there are problems, most likely with the quotes mentioned earlier, which causes some discussions/comments to disappear (as the json produced obviously causes errors when trying to send them).

If anyone is interested I can of course share the php script so you can have a look at it, but I have to warn you that it's very messy in it's current state...

larspalo:
If one writes [oleg68](https://github.com/oleg68) `[oleg68](https://github.com/oleg68)` one gets a link to the user. So, it should be possible to examine the author string against the list above and link to the user by changing the text to use. That would not be too much work to implement in my script.

larspalo:
I'll test a new version of the import script, which means that I first must delete all the discussions I imported earlier. So, for a while the discussion forum will not be as it should...

larspalo:
My most recent version of the import script seems to be able to import all the forum posts. It unfortunately mess up some of the quotes still - it rewrites them to html characters which seems to not (at least in the discussion titles) be recognized for what they are on GitHub... Still you can have a look at the result in the discussion forum of my fork. Mapping the known users work fine with the rewrite. If someone is really interested in chasing down more users I attach a text file with the usernames from those who posted in the GrandOrgue forum on SourceForge. [GOforum_authors.txt](https://github.com/GrandOrgue/grandorgue/files/7164024/GOforum_authors.txt)

oleg68:
@larspalo I've look the test discussions. My notes:
1. All messages are authored by larspalo. May it be changed?
2. All messages were posted one hour ago, so sorting by last message doesn't work.

Does GrapQL API allow to override the technical fields: the author and the posting date?

larspalo:
>     1. All messages are authored by larspalo. May it be changed?

Not really, that's why the first paragraph is added stating when and whom made the original post.
>     2. All messages were posted one hour ago, so sorting by last message doesn't work.
The posts are sorted by the last actual date any comment was made in that discussion - that is, they have the same sort order as the SourceForge forum.

> 
> Does GrapQL API allow to override the technical fields: the author and the posting date?
Not at creation time with the mutations, anyway. I can only find them in a query type, and then I can only view them.

larspalo:
So, what do you think? Do we need to polish the import more or can we live with it as it is? If I eventually need to make another import I perhaps should correct a few at mentions in the original text that seem to annoy some people here on GitHub.

oleg68:
Except is no way of modifying the author and the posting date,  all others look fine and we may live with it.

I'm going to transfer the topics here from my fork just after your import.

rousseldenis:
Let's go for this

larspalo:
The old posts in the Sourceforge forum are now imported to GrandOrgue on GitHub. I hereby close this PR.

PR #20, opened 2021-09-09
Removed scripts for parsing svn history

oleg68:
When GO was on SourceForge I used these scripts for parsing svn history.

After migration to GitHub they are not necessary


rousseldenis (approved):

rousseldenis:
@oleg68 FYI, I used git svn tool that has advantage of be able to match user commit in svn and github user.

larspalo (approved):

PR #682, opened 2021-09-16
Increasing the limit of number of ranks from 400 to 999

oleg68:
Discussed at https://github.com/GrandOrgue/grandorgue/discussions/680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Increasing the limit of number of ranks from 400 to 999 https://github.com/GrandOrgue/grandorgue/discussions/680
 # 0.3.1.2342 (2021-09-16)
 - Fixed build for windows on github https://github.com/GrandOrgue/grandorgue/issues/29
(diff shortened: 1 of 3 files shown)


larspalo (approved):
Nice touch updating the help too, thanks!

rousseldenis (approved):

oleg68:
@rousseldenis because Lars wants to recruit @Jean-Gabriel1982, is it possible to add him as a reviewer? I couldn't find this option.

oleg68:
Agree. I think we should to wtite CONTRIBUTORS.md with a step-by-step instruction that would start with very easy steps and would allow to be involved.

rousseldenis:
> @rousseldenis because Lars wants to recruit @Jean-Gabriel1982, is it possible to add him as a reviewer? I couldn't find this option.

I think everybody can review.

Apart of that, I think we can add some teams with different rights (not so many to begin simple)

Jean-Gabriel1982 (approved):

PR #718, opened 2021-09-21
[IMP] Improve licensing root files

rousseldenis:


rousseldenis:
> I think that the start header of the LICENSE file should be kept (from the old licence.txt). Partly it's because of the controversial ancient code heritage, but also it's the lifeline to acceptably be able to include ASIO support in the released builds.

I've read that it's better to keep GPL license text as is. But, my idea was to include in the short header of code files particular things like that where it's needed.

Moreover, we can explain in AUTHORS.txt (as it is currently) those paricularities.

rousseldenis:
And I think we could embed AUTHORS file there too.

https://github.com/GrandOrgue/grandorgue/pull/718/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR210

larspalo:
So where do you put:

  This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version. 

  This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License below for more details. 

  Linking GrandOrgue statically or dynamically with other modules is making a combined work based on GrandOrgue. Thus, the terms and conditions of the GNU General Public License cover the whole combination.

  In addition, as a special exception, the Kloria Publishing LLC and GrandOrgue Team gives permission to combine GrandOrgue with any ASIO SDK and/or VST SDK produced by Steinberg Media Technologies GmbH and to distribute the linked executables. You must obey the GNU General Public License in all respects for all of the code used other than the ASIO SDK and/or VST SDK.

  Note that entities making modified versions of GrandOrgue are not obligated to grant this special exception for their modified versions; it is their choice whether to do so. The GNU General Public License gives permission to release a modified version without this exception; this exception also makes it possible to release a modified version which carries forward this exception.

that are part of the original license.txt file?

rousseldenis:
> So where do you put:
> 
> This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version.
> 
> This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License below for more details.
> 
> Linking GrandOrgue statically or dynamically with other modules is making a combined work based on GrandOrgue. Thus, the terms and conditions of the GNU General Public License cover the whole combination.
> 
> In addition, as a special exception, the Kloria Publishing LLC and GrandOrgue Team gives permission to combine GrandOrgue with any ASIO SDK and/or VST SDK produced by Steinberg Media Technologies GmbH and to distribute the linked executables. You must obey the GNU General Public License in all respects for all of the code used other than the ASIO SDK and/or VST SDK.
> 
> Note that entities making modified versions of GrandOrgue are not obligated to grant this special exception for their modified versions; it is their choice whether to do so. The GNU General Public License gives permission to release a modified version without this exception; this exception also makes it possible to release a modified version which carries forward this exception.
> 
> that are part of the original license.txt file?

Indeed. I've read the particular things related here https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs

That correspond at first sight to the text you point.

As I understand, it should be included in the 7th paragraph. If you understand it like me

larspalo:
Only for GPLv3. For GPLv2 it's not explicitly stated where to put it. The meaning of the line:

> In either case, you should put this text in each file to which you are granting this permission.

is not entirely clear. Do we need to have that in every header of the program or if it's enough to just include at the top of the license file or some other file?

rousseldenis:
> Only for GPLv3. For GPLv2 it's not explicitly stated where to put it. The meaning of the line:
> 
> > In either case, you should put this text in each file to which you are granting this permission.
> 
> is not entirely clear. Do we need to have that in every header of the program or if it's enough to just include at the top of the license file or some other file?

As I've understood and from my day-to-day practice, I would say the entire license file should put on root and shorten ones on all files headers.

As either you write code that use these libraries, that written code is under GPL.

We just ensure that user is warned that some embedded libraries are under other licenses. BUT, for ASIO specifically, I've read the agreement and there is nothing that is mandatory about licensing marks in code.

The only thing is we need to mention ASIO logo and a text in the About dialog and sign agreement.

larspalo:
I suggest that we 
1) include the section I mentioned above in the root license file like it was earlier but adjust the formatting to make more sense
2) shorten and update the headers of all other source code files (that actually belong to GO and doesn't come from other sources - which we shouldn't modify at all)
3) modify and update the AUTHORS file with the new contributors

rousseldenis:
> or (at your option) any later version.

@larspalo If I read correctly, the GPL version is at least GPL2 (which is deprecated) or later. So, shouldn't we adapt to GPL v3 ?

rousseldenis:
I've re-added the former part at the beginning of the LICENSE root file and modified the sources files headers.

larspalo:
> > or (at your option) any later version.
> 
> @larspalo If I read correctly, the GPL version is at least GPL2 (which is deprecated) or later. So, shouldn't we adapt to GPL v3 ?

Yes, on closer inspection we could. See https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.html#VersionTwoOrLater

That's why it's important to keep that part, and eventually when/if we go to v3 it should be “Version 3 of the GPL or any later version”

larspalo (commented):
Add the first section too to the license file that makes it possible to distribute under GPL v2 or later, but change the initial wording to "GrandOrgue is free software".. Also adjust the line lenghts in all the (five) top sections of the license file so that they are not too long. 

Update the dates in all the shortened headers to 2009-2021 and change to GPL-2.0-or-later instead of only GPL-2.0

Please add Oleg, yourself and other new contributors from Olegs fork to the AUTHORS file

oleg68:
@rousseldenis why copyright is ``2009-2019`` instead of ``2009-2021``?

rousseldenis:
> @rousseldenis why is copyright `2009-2019` instead of `2009-2021`?

Cause I took the original ones, but I can change :sweat_smile:

rousseldenis:
> @rousseldenis why is copyright `2009-2019` instead of `2009-2021`?

But, IMHO, last part of year range should be updated only when the concerned file is updated too. No ?

rousseldenis:
> I suggest that we
> 
> 1. include the section I mentioned above in the root license file like it was earlier but adjust the formatting to make more sense
> 2. shorten and update the headers of all other source code files (that actually belong to GO and doesn't come from other sources - which we shouldn't modify at all)
> 3. modify and update the AUTHORS file with the new contributors

@larspalo Done in separate commits. I'm not sure all headers have been done, but that can be done through future devs, the important ones are done (/src).

rousseldenis:
Do we try to integrate ASIO icon ? Should I sign the agreement in the name of Organization to Steinberg ?

oleg68:
@rousseldenis not yet. I suggest to separate asio-related things to a separate issue.

rousseldenis:
> @rousseldenis not yet. I suggest to separate asio-related things to a separate issue.

Ok, so if you agree, this one seems to be ready

larspalo:
No, it's not ready as it is now. Still the top paragraph is missing in the LICENSE file and the formatting of the lines should be improved (shorter lines).

I think the years should be updated. From https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

>  When you add the new year, it is not required to keep track of which files have seen significant changes in the new year and which have not. It is recommended and simpler to add the new year to all files in the package, and be done with it for the rest of the year.

Non trivial changes to the GO package has been made continuously since 2019 it's just that noone has bothered with updating the dates.

rousseldenis:
> No, it's not ready as it is now. Still the top paragraph is missing in the LICENSE file and the formatting of the lines should be improved (shorter lines).

It looks :

image

rousseldenis:
> I think the years should be updated. From https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html
> 
> > When you add the new year, it is not required to keep track of which files have seen significant changes in the new year and which have not. It is recommended and simpler to add the new year to all files in the package, and be done with it for the rest of the year.
> 
> Non trivial changes to the GO package has been made continuously since 2019 it's just that noone has bothered with updating the dates.

Ok, do it now

larspalo:
Ok, strange, it still looked wrong when I last looked a few minutes ago...

rousseldenis:
> Ok, strange, it still looked wrong when I last looked a few minutes ago... Yes, now I can see it. Great!

When, you get back on your browser, github pages are usually in cache with previous version. A page refresh solves that.

rousseldenis:
2021 done on all headers.

larspalo (approved):

rousseldenis:
@oleg68

oleg68 (approved):

Issue #712, opened 2021-09-19
[Bug] Panel menu is too narrow


oleg68:
When opening a Panel menu for the first time, the menu rectangle is narrow; the pannel names do not fit to it

rousseldenis:
Indeed, +1

larspalo:
Actually, I think it's not really populated until an organ is opened/loaded.

larspalo:
But yes, even after an organ is opened/loaded the first opening of the Panel menu is too narrow the first time so it's a bug.

Jean-Gabriel1982:
+1

oleg68:
The reason of this bug is that GOrgueFrame::UpdatePanelMenu, that populates this menu, is called a response to wxEVT_MENU_OPEN event. But the size of menu is defined before wxEVT_MENU_OPEN.

The solution may be to populate PanelMenu in Advance, when an organ is loaded.

PR #724, opened 2021-09-23
Fix Panel menu sizing issue #712

larspalo:
Updating panel menu when organ is loaded

### DIFF
--- a/src/grandorgue/GrandOrgueFrame.cpp
+++ b/src/grandorgue/GrandOrgueFrame.cpp
@@ -476,6 +476,7 @@ void GOrgueFrame::Open(const GOrgueOrgan& organ)
 	GOrgueProgressDialog dlg;
 	m_doc = new GOrgueDocument(&m_Sound);
 	m_doc->Load(&dlg, organ);
+	UpdatePanelMenu();
 }
 
 GOrgueDocument* GOrgueFrame::GetDocument()
--- a/src/grandorgue/GrandOrgueFrame.cpp
+++ b/src/grandorgue/GrandOrgueFrame.cpp
@@ -463,6 +463,7 @@ bool GOrgueFrame::DoClose()
 		return false;
 	delete m_doc;
 	m_doc = NULL;
+	UpdatePanelMenu();
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/grandorgue/GrandOrgueFrame.cpp
```diff
@@ -476,6 +476,7 @@ void GOrgueFrame::Open(const GOrgueOrgan& organ)
 	GOrgueProgressDialog dlg;
 	m_doc = new GOrgueDocument(&m_Sound);
 	m_doc->Load(&dlg, organ);
+	UpdatePanelMenu();
 }
```
oleg68, line 481: 1. When current organ is closed without opening the new one, the menu also should be updated (GOrgueFrame::DoClose() ?).
2. Because the panel menu is changing only when opening/closing an organ, calling UpdatePanelMenu() is no more necessary in GOrgueFrame::OnMenuOpen

oleg68 (changes requested):

src/grandorgue/GrandOrgueFrame.cpp
larspalo: 1. Perhaps, to be consistent - but it's not as problematic for an user as it is now either since it's empty anyway, large or small
2. No, it's still necessary to provide the ticks next to what panel(s) is/are open.

larspalo: 1. Fixed
2. Won't fix, as this handles tick on/off for active or not active panels

oleg68: 1. I see you have added it
2. Could you explain more?

    > to provide the ticks next to what panel(s) is/are open

    According to the code, UpdatePanelMenu fully destroys the old panel menu and constructs a new one and does not do anything more. This action is necessary when the list of pannels can be changed, but looks unnecessary on each opening the menu.

larspalo: When you have an organ opened the Panel menu is filled and each open panel will have a tick to the left of it to indicate that the panel is open. That tick is placed by the line
`item->Check(doc->WindowExists(GOrgueDocument::PANEL, panel) ? true : false);`
in the UpdatePaneMenu.
If the panel is not open there won't be a tick next to it. All that is taken care of in the UpdatePanelMenu function.

oleg68: Yes, you are right.

The best solution would be to split UpdatePaneMenu to two parts: the highweight constructing the menu - from Open and DoClose and the lighweigth item->Check for each panel - from GOrgueFrame::OnMenuOpen. But this refactoring is more than just bugfixing.

oleg68 (approved):

src/grandorgue/GrandOrgueFrame.cpp
larspalo: That's my thoughts, exactly!

rousseldenis (approved):
LGTM

Issue #14, opened 2021-09-08
Release numbering is outdated


larspalo:
The numbering of the releases should reflect what revision they come from. That was earlier set with cmake with the -DVERSION_REVISION= at build time. I suggest we return to that release numbering instead of the non meaningful current version of keeping an outdated revision number with dash something.

rousseldenis:
> The numbering of the releases should reflect what revision they come from. That was earlier set with cmake with the -DVERSION_REVISION= at build time. I suggest we return to that release numbering instead of the non meaningful current version of keeping an outdated revision number with dash something.

I ++ for this.

oleg68:
Now it can be set  with  -DVERSION and -DBUILD_VERSION parameters at the build time.

I'd like to discuss changing version numbering. I have some thouths.
But I think we may to return here after the first oficcial release `0.3.1.2342-1`

Issue #711, opened 2021-09-19
Automate issuing of releases


oleg68:
Now the process of issuing of releases is the following

1. Create a release tag and push it to master
2. Wait when the release auto build creates the release on github and update its description manually with the CHANGELOG.md contents

(1) was acceptable when GO was hosted in a private repository. But working with organisation means no direct access to the GO repositories (I disable it in my git settings for prudence) and making all changes in own fork with creating a pull request. But Github does not allow PR on tags. It requires to enable direct access to the target repo temporary.

Moreover issuing releases should have aproval from other members. Tagging does not allow this.

(2) requires manual actions and may be done incorrectly.

My approach of the new process is the following:

1. Add the new release label on top of CHANGELOG.md
2. Make a ususa PR with the only change in this file
3. After approval and merging this PR all futher steps should be made authomatically. 

@larspalo @rousseldenis how do you find the new process?

rousseldenis:
In the processes I'm used to, people that have write access will create releases.

Usually, updating the changelog and using tags. IMHO, the use of tags should be mandatory as they allow to freeze a version. 

Concerning the approval of other members, maybe yes or maybe not, I'm not sure this will help in the flow.

PS: A little tool that helps releases is bumpversion. It allows to modify several files at once with --patch/minor/major commands.

rousseldenis:
> Usually, updating the changelog and using tags. IMHO, the use of tags should be mandatory as they allow to freeze a version

And the main aim too is that allows to get back to that version if master is in a mess.

oleg68:
> IMHO, the use of tags should be mandatory as they allow to freeze a version.

I suggest the build script to read the first line of CHANGELOG.md and to put the tag authomatically.  Not to put the version tag manually.

rousseldenis:
> > IMHO, the use of tags should be mandatory as they allow to freeze a version.
> 
> I suggest the build script to read the first line of CHANGELOG.md and to put the tag authomatically. Not to put the version tag manually.

Of course, if we can make our path in order to get them automatically, let's go

larspalo:
I like the process Oleg suggests.

PR #726, opened 2021-09-24
Renewed a default GO version during cmake build

oleg68:
Resolves: #14

Earlier cmake used `0.3.1` as the base and concatenated the release version had been taken from svn.

Now we assign the release version manually.

I've introduced the file `version.txt` where the current release version is stored.

Now cmake by default takes the version from `version.txt` and uses 0.YYMMDD as a build version.

During automatic builds on github VERSION and BUILD_VERSION are calculated from git tags and commit history.

In the future  (#711) I'm going to implement automatic git tagging when version.txt is being changed.


larspalo:
I see you slightly adjusted the build numbering from what we discussed earlier and what your previous solution was. I guess that we still will set the build type number to -1 and adjust the release number when we do the actual releases, right, or did we just drop that part completely and have only a date based last part?

Also, what happens if intermediate builds are triggered the same day?

oleg68:
> I see you slightly adjusted the build numbering from what we discussed earlier and what your previous solution was.

This PR does not affect automatic builds on github, because they go with -DVERSION and -DBUILD_VERSION passed. They still have 3.2.0-0.<small number> for intermediate builds and 3.2.0-1 for the release.

But what version to use with local build, when you run cmake or build-on-linux.sh manually?

My approach is 3.2.0-0.<something>, where <something> differs from any github autobuilds. I've made YYMMDD, but maybe to make 'local' (3.2.0-0.local) is better?

larspalo (approved):

oleg68:
@larspalo, may be change to `local`?

larspalo:
@oleg68 I can live with whichever, a .local ending is fine for me. If someone wants to change it they can override it at build time. Personally I can imagine overriding it to whatever branch (name) I'm testing at the moment just to know what the build is supposed to do.

oleg68:
> @oleg68 I can live with whichever, a .local ending is fine for me. If someone wants to change it they can override it at build time. Personally I can imagine overriding it to whatever branch (name) I'm testing at the moment just to know what the build is supposed to do.

Chanded to `0.local`. I like your idea to use the branch name as a part of BUILD_VERSION

larspalo (approved):

oleg68:
@rousseldenis  could you approve this pr?

rousseldenis (approved):

PR #738, opened 2021-09-28
Update GrandOrgue.appdata.xml to match latest release

lbssousa:


oleg68 (approved):
I'm approving this issue as a temporary solution, but I think that this file

1. Should be generated automatically on a release
2. Should be placed among release assets, not in the sourcecode.

rousseldenis:
> 1. Should be generated automatically on a release

Indeed, that's why I proposed before to use bumpversion as it allows to maintain versions texts on several files at once.

[](https://github.com/peritus/bumpversion)

rousseldenis (approved):

larspalo (approved):

PR #746, opened 2021-09-30
Fix scrolling in Organsettings dialog resolves #735

### LINKED ISSUE #735: Organsettings dialog is not scrollable
The right part of the organsettings dialog can be invisible to users with a lower resolution screen. It should be made scrollable.

larspalo:


oleg68:
@larspalo Your cod is not compilable for Windows. Look at https://github.com/larspalo/grandorgue/runs/3758701722?check_suite_focus=true

oleg68:
@larspalo Could you register your change in CHANGELOG.md?

oleg68:
@larspalo  I couldn't find any scrollbars with your intermediate build:

OrganSettings

larspalo:
What platform? Seems like the dialog is too small too on your screenshot.

larspalo:
Actually, as I solved it, I only added scrolling to the right part of the dialog and only in vertical direction since that was the part that was hidden.

Is it desirable that the dialog should function in another way?

larspalo:
Tested on windows and the scrollbars show up there. But the default sizing of the dialog is not as good on windows as on Ubuntu. Maybe SetSizeHints will work better than Fit?

oleg68:
> What platform? Seems like the dialog is too small too on your screenshot. Is that the default size or have you resized the dialog?

Fedora/Gnome

I tested it again. The vertical scrollbar is not visible, but when I move mouse to the right margin of the dialog, the scrollbar appears. So I think It is OK
OrganSettings2

oleg68:
> Actually, as I solved it, I only added scrolling to the right part of the dialog and only in vertical direction since that was the part that was hidden.

> Is it desirable that the dialog should function in another way?

I think yes, it is. This behavior is right.

oleg68:
> Tested on windows and the scrollbars show up there. But the default sizing of the dialog is not as good on windows as on Ubuntu. Maybe SetSizeHints will work better than Fit?

wx 3.0.2...

Earlier I had problems with scrolling and the default size. I did not test SetSizeHints. wxWindow has both `BestSize` and `MinimalSize`, but, unfortunally BestSize doesn't work and `Fit` forces all children to become to their `MinimalSize`, not to `BestSize`.

oleg68:
@larspalo Please rebase your PR on top of the new master because the conflict occurs.

oleg68 (approved):

rousseldenis:
@larspalo it's usually better to do a rebase on master to solve conflicts than a merge from master to your branch.

That avoids 'merge' commits.

larspalo:
> @larspalo it's usually better to do a rebase on master to solve conflicts than a merge from master to your branch.
> 
> That avoids 'merge' commits.

@rousseldenis I'm happy that I can manage to get git going at all... ;-)

src/grandorgue/OrganDialog.cpp
```diff
@@ -262,6 +262,7 @@ OrganDialog::OrganDialog (GOrgueDocumentBase* doc, wxWindow* parent, GrandOrgueF
 	Load();
 	
 	topSizer->Fit(this);
+	topSizer->SetSizeHints(this);
```
oleg68, line 265: After this changes I cann't resize this dialog vertically at all: in opens with the whole wcreen height and resizing works only vertically.

oleg68: The behaviour on windows has not changed
GrandOrgueWin

oleg68: I suggest to remove SetSizeHints, but to specify a minimum size 640 in OrganDialog::OrganDialog:

```
(GOrgueDocumentBase* doc, wxWindow* parent, GrandOrgueFile* organfile) :
	wxDialog(parent, wxID_ANY, _("Organ settings"), wxDefaultPosition, wxSize(wxDefaultCoord, 640), wxDEFAULT_DIALOG_STYLE | wxRESIZE_BORDER),
```

oleg68 (changes requested):

src/grandorgue/OrganDialog.cpp
larspalo: How strange. It did on my windows computer... No, I don't think we should specify a fixed minimum size that way, it should be dictated by the contents of the sizers. And yes, it was my intention to disallow resizing the dialog too much horizontally, so that one wouldn't run into the issue with your first screenshot above.

oleg68: > No, I don't think we should specify a fixed minimum size that way, it should be dictated by the contents of the sizers.

By what algorithm?

- If all contents must fit the minimal size then scrolling is not necessary
- If allow the resizing behind the ''all fit" size then how to consider the contents?

> And yes, it was my intention to disallow resizing the dialog too much horizontally

horizontally - ok, but I could not reduce the size vertically

> so that one wouldn't run into the issue with your first screenshot above.

This issue doesn't depend on the horizontal size.

larspalo: Scrolling will be necessary if the screen is too small for the content, but if the screen is big enough for the content of the dialog then no scrolling of course is needed and shouldn't be available. And yes, in theory the dialog could be too big horizontally for a certain screen resolution but that's quite unlikely to ever happen.

Well, the original issue is solved, I just found some other side effects that perhaps should be addressed too. I have a new test going on right now.

oleg68: Now it is too big vertically and does not allow resizing. The behaviour without SetSizeHints was better
OrganSettings69

larspalo: Ok, you might be right about that. Seems like since the behaviour is not consistent on the different platforms we perhaps should be content with just fixing the original issue and leave the rest for now.

oleg68 (approved):

rousseldenis (approved):

rousseldenis:
> > @larspalo it's usually better to do a rebase on master to solve conflicts than a merge from master to your branch.
> > That avoids 'merge' commits.
> 
> @rousseldenis I'm happy that I can manage to get git going at all... ;-)

Hehe!

The problem with that is that we have 'noise' commits in master now : 

image


The good command steps to solve conflicts is to rebase your branch on the target branch:

```
1) git fetch
2) git rebase <GrandOrgue>/master   where <GrandOrgue> is the GrandOrgue remote url - check your .git/config file
```

3) If you have conflicts (you see concerned files in log), solve them, then:

```
4) git add <your file(s)>
5) git rebase --continue
```

Redo 3-4-5 as many conflicts you have per commit history you get from master.

Then, do a :

```
git push origin HEAD -f
```
where origin is your repository.

larspalo:
> 
> The problem with that is that we have 'noise' commits in master now :
> 

@rousseldenis Understood, I'll try to get my lack of git experience fixed eventually...

> The good command steps to solve conflicts is to rebase your branch on the target branch:
> 
> ```
> 1) git fetch

On my issue branch or on the master of my fork? As I understand it I must first get my fork master in sync with the origin - the Official Grandorgue repository master. The easiest way I can think of is to never actually do any work on my master - I just fetch and merge any changes from the official right here on GitHub, then I can use the command "git pull" on my local fork master to bring it in sync with GitHub. All my local programming work is then done in a specific branch that I create, right, and I might need to get that branch in sync with the master (that's always supposed to be an exact copy of the official master). That's where this "rebase" comes in I guess...

> 2) git rebase <GrandOrgue>/master   where <GrandOrgue> is the GrandOrgue remote url - check your .git/config file
> ```
> 

That's my fork master then, right? Because I don't think I've setup any upstream master on my local git - that's just a clone of my fork.

>     1. If you have conflicts (you see concerned files in log), solve them, then:
> 
> 
> ```
> 4) git add <your file(s)>

If I changed any to solve the conflict, right? When I've been working on something locally, I've so far only staged and commited in one go (git commit -a -m 'Message').

> 5) git rebase --continue
> ```
> 
> Redo 3-4-5 as many conflicts you have per commit history you get from master.
> 
> Then, do a :
> 
> ```
> git push origin HEAD -f
> ```
> 
> where origin is your repository.

I've always just used "git push" on the current local branch when I've made any changes there that I want to send to my GitHub fork.

Also, looking at the commit history above, it would perhaps look better if I had all the commits fixed before even opening a PR, but... To be honest, I'd like more to spend time working on the code itself than worrying about how the commit history will look like.

We should maybe have a detailed developer instruction manual on how to do all the things necessary with git and GitHub. Personally, I think that the actual programming work is the easier part of development...

Issue #741, opened 2021-09-29
Installing/Upgrading GO on Linux with graphic installers


oleg68:
The [Install guide](https://github.com/GrandOrgue/grandorgue/blob/master/INSTALL.md) suggests to use a command line package manager (dnf, zypper, yum, apt-get) for installing GO on linux. But lots of users do not read this guide and try to install it just with double clicking on the rpm or deb file. The OS complains about unsatisfied dependencies and refuses to install.

The root cause is there are two packages `grandorgue` and `grandorgue-resources` that depend on each other. It requires installing them at the same time that double-click installers do not support.

My approach is

1. To break dependency grandorgue-resources -> grandorgue. It will allow to install grandorgue-resources before grandorgue
2. To let loose the dependency from grandorgue -> grandorgue-resources. Now it requires to be the same version, but in most cases `great or equal` will be sufficient. It will allow to upgrade grandorgue-resources with a double-click installer befor upgrading `grandorgue`

It will be still necessary to read some guide for knowing that resuurces should be installed first.

Another approach - not to make separate packages for grandorgue and grandorgue resources.

Jean-Gabriel1982:
Yes at the beginning it is a bit confusing ..
I've downloaded
Grandorgue, grandorgue-ressources, and grandorgue-demo.

First, we need to install Ressources, then GrandOrgue, and finish by Demo by double-clicking on the .deb file.

Is it necessary to have 3 packages instead of only 1 ?

oleg68:
> I've downloaded Grandorgue, grandorgue-ressources, and grandorgue-demo.

Lots of users download only grandorgue and try to install them.

> First, we need to install Ressources, then GrandOrgue, and finish by Demo by double-clicking on the .deb file.

For now, it is not possible.

> Is it necessary to have 3 packages instead of only 1 ?

For now, yes, it is.

oleg68:
@larspalo How do you find an idea to make a singleton package `grandorgue` for linux witout separating resources and demo i own packages?

oleg68:
@paolo2504 You like when everything is optimised. How do you find the idea when GrandOrgue Demo Organ is always installed without capability to install GrandOrgue without the demo,

paolo2504:
Generally speaking I personally prefer not to have separate package installation for Go and GO resources. I mean, in the future I’d like to have one single install that takes care first of resources then of GO main. 

[...]

PR #747, opened 2021-09-30
Making the single grandorgue linux package

oleg68:
Resolves: #741

It makes the grandorgue .rpm/.deb file installable with double-click

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ if (WIN32 OR APPLE)
 else ()
    option(INSTALL_DEPEND      "Copy dependencies (wxWidgets libraries and Translations) on installation" OFF)
 endif ()
+option(GO_SEPARATE_LINUX_PACKAGES "Generate separate linux packages for resources and demo" OFF)
 
 # only use options supported by the compiler
 include(CheckIncludeFileCXX)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -304,13 +305,17 @@ elseif(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   set(CPACK_RPM_DEMO_PACKAGE_SUMMARY "${CPACK_COMPONENT_DEMO_DISPLAY_NAME}")
   set(CPACK_RPM_DEMO_PACKAGE_ARCHITECTURE noarch)
   set(CPACK_RPM_DEMO_FILE_NAME RPM-DEFAULT)
+  if(NOT GO_SEPARATE_LINUX_PACKAGES)
```
rousseldenis, line 308: If I understand correctly (I'm not aware of packaging processes), is it really needed to keep option to manage separate packages? Why not keeping it simple?

oleg68: I kept this option for possible future integration with OBS and making native packages for various distributions.

larspalo (approved):

CMakeLists.txt
rousseldenis: Ok, understand

rousseldenis (approved):

PR #756, opened 2021-10-05
Update LINGUAS

Romek1234:
Unfortunately, this compilation https://github.com/GrandOrgue/grandorgue/actions/runs/1307218898 does not support Polish.

I wonder if pl is needed in the file / po / LINGUAS?

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,1 +1,1 @@
-de es fr it nl sv
+de es fr it nl sv pl


oleg68:
@Romek1234 Please check your fork builds at https://github.com/Romek1234/grandorgue/actions Is Polish working?

oleg68:
https://github.com/GrandOrgue/grandorgue/blob/master/src/grandorgue/SettingsOption.cpp should also be updated for supporting a new language.

oleg68:
This file is not used

Issue #762, opened 2021-10-09
Panic is not able to reset sound system if some thread is waiting for a mutex.


oleg68:
After researching #702 it found out that Panic cann't process when some sound thread is waiting for a mutex.

Whatever the root cause of locking the mutex `Panic` should reset the sound system even if something goes wrong.

oleg68:
References: #702

PR #765, opened 2021-10-09
Bugfix/moving classes and files

oleg68:
For working with #762 I did some beautification of the code without changing the behaviour.

I've tried to achieve
- Each of threading classes has separated header (.h) and implementation (.cpp) file
- The name of the files are the same as the names of the classes
- All the files are in the separate directory src/core/threading

I made a separate PR for a part of changes because a lot of files had been affected with trivial changes (mostly the path of include file). It would be difficult to review them with essencial changes.

### DIFF
--- a/src/core/threading/GOMutexLocker.h
+++ b/src/core/threading/GOMutexLocker.h
@@ -4,7 +4,7 @@
 * License GPL-2.0 or later (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
 */
 
-#ifndef GOUTEXLOCKER_H
+#ifndef GOMUTEXLOCKER_H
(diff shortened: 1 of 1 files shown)


larspalo (commented):
The copyright header should be added to all the .cpp files too, right?

In src/core/threading/GOMutexLocker.h you have a typo for the #ifndef on line 7

oleg68:
> The copyright header should be added to all the .cpp files too, right?
> 
> In src/core/threading/GOMutexLocker.h you have a typo for the #ifndef on line 7

Fixed

larspalo (approved):

oleg68:
@rousseldenis ?

rousseldenis (approved):

PR #772, opened 2021-10-13
Support of stopping the thread in GOSoundWorkItem::Run

oleg68:
This is the second part of changes caused by #762

The change is quite simple: GOSoundWorkItem::Run gets an additional parameter a GOSoundThread where the item is running now.

GOSoundThread::Entry passes the parameter to GOSoundWorkItem::Run.

It is intended GOSoundWorkItem::Run to check GOSoundThread::ShouldStop and to break its work if the thread is being stopped.

Because there are lots of derived classes from GOSoundWorkItem, their files are also changed.
None of them are using this new parameter now so the behaviour does not change yet. But I make a separate PR for it because it would difficult to review a lot of trivial changes at the same time as a change of behavior

### DIFF
--- a/src/core/GOSoundTouchWorkItem.cpp
+++ b/src/core/GOSoundTouchWorkItem.cpp
@@ -30,7 +30,7 @@ bool GOSoundTouchWorkItem::GetRepeat()
 	return false;
 }
 
-void GOSoundTouchWorkItem::Run()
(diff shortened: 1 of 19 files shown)


oleg68:

rousseldenis (approved):
LGTM

PR #774, opened 2021-10-14
[Enhancement] New Per-Pipe Release Truncation Mechanism

pgcrowe:
As promised, I have completed the process of adding my new per-pipe release truncation mechanism to the latest source code for GrandOrgue. All code has been revised to incorporate recent changes in the GrandOrgue source, notably revised wxWidgets GUI code, After debugging for compiling errors and GUI bugs, I'm happy to report that the result works beautifully on macOS Big Sur 11.6.

### Instructions for Use:
- To set Release Truncation Length in a **_Definition File_**, use **_Pipe999ReleaseTruncationLength_**
- To set Release Truncation Length values in the **_Organ Settings_** panel, select one or more pipes, enter your desired release length (ms) in the new  _**Rel. Truncation (MS)**_ box, then click apply.
- The default value for release truncation is  _**0**_, which **disables** release truncation. 

### Important Design Notes (Developers, Please Read):
- The **Pipe999ReleaseTruncationLength** is designed to function similar to the **Pipe999ReleaseCrossfadeLength** mechanism. Hence, _**Pipe999ReleaseTruncationLength**_ must be set by pipe in the definition file.
- Unlike _**Pipe999ReleaseCrossfadeLength**_, _**Pipe999ReleaseTruncationLength**_ values are adjustable in the **Organ Settings Panel.**
- Enabling the organ-wide truncation mechanism in the toolbar OVERRIDES all Pipe999ReleaseTruncationLength settings. If desired, code could be altered to make the Organ-Wide Truncation mechanism truncate only when a **Pipe999ReleaseTruncationLength** value is greater than an **Organ-Wide Truncation Length** value.
- Currently, the **_Organ Settings Panel_** displays the **_Rel. Truncation (ms)_** box on all tree items. At present, any change to an _**upper tree**_ item overrides all _**sub-tree**_ items (EG: A Pipe is overridden by a Rank, which is overridden by a Division). _A more intuitive implementation would be to program upper tree items to automatically change the numeric values listed in subtree item **Rel. Truncation (ms)**  boxes._ If anyone reading this knows an easy way to program this behavior, please do so!

My contribution here aside, I am thrilled to see the revitalized life has been brought to the GrandOrgue project in recent months, by the newest GrandOrgue contributors Oleg and Dennis! With the submission of this feature, I am honored to join in their revitalization efforts, and the tireless work of the founding developers who laid GrandOrgue's foundation years ago, and CONTINUE to innovate and enhance this "grand" project. 

This PR is Developed, Fully Tested, and Debugged

Vivat, GrandOrgue!
Pat


oleg68:
> Please, remove **all** comments 

@larspalo I think commenting the code is a good style. So I'd keep the comments and I'd add more commebts in the GrandOrgue source code in the future.

larspalo:
Sorry, but the really needed documentation is lacking in this PR - in the GO help! No new user features should be added without it being documented in the help - user documentation should not be in the source code.

The very worst thing (except for that most of this is already self-explaining) is the mixing of styles in the comments and where they are placed in the code. Presently, in this PR, it clutters the code way beyond any possible benefits.

pgcrowe:
> > Please, remove **all** comments
> 
> @larspalo I think commenting the code is a good style. I want to thank @pgcrowe for adding comments to the existing code.
> 
> So I'd keep the comments and I'd add more commebts in the GrandOrgue source code in the future.
> 
> I agree with removing only comments like
> 
> ```
> // TRUNCATION SETTINGS � 11-23-20 - Release Sample Truncation Settings
> ```
> 
> and
> 
> ```
> // END TRUNCATION SETTINGS � 11-23-20 - Release Sample Truncation Settings
> ```
> 
> because they double the standard git `Blame` feature.
My pleasure! I'll clean them out.

larspalo:
> It is : https://github.com/GrandOrgue/grandorgue/tree/master/help

Right, that was exactly what I was saying earlier! That's where the documentation should be, not as comments scattered in the code.

Now, as for the code, comments can of course be done where needed. But...

1. They should be sensible and serve a pupose that's not just stating the obvious like `// Added for Per-Pipe Release Truncation Mechanism` The wonderfully constructed variable name just say the same. Good choice of variable names say what's needed to know.
2. Don't put "pretty" versioning information into the comments like `                    // BEGIN TRUNCATION SETTINGS - 11-23-20 - Release Sample Truncation Settings
                     /* ***************************************** *\
                         *              IF STATEMENT:                *
                         * IF: release truncation values exist from  *
                         * (Pipe999ReleaseTruncationLength)...       *
                        \* ***************************************** */`
3. Don't state obvious things like where a bracket ends. Like `/* *************** END: if (m_ScaledReleases) *************** */`, that's what the indentation itself is for.

It's a real problem (in my opinion) when a PR, that's actually very good in it's intention and deserves to be recognized as just that, contain more distracting comments than productive code to review.

So, @pgcrowe realize that I'm not attacking your contribution in it's essence - I'm rather positive about it. But there are also some things to refine before this PR is "production" ready.

johnroper100:
@larspalo @pgcrowe This is a good resource: https://stackoverflow.blog/2021/07/05/best-practices-for-writing-code-comments/

larspalo (changes requested):
You're getting closer! I think the following things need correction

1. You have versioning and date information present in comments in the code. Please remove them. (everything with `// 11-20-20 - New Pipe Release Truncation Mechanism` and the likes)
2. All the variable comments that say the same thing as the name of the variable are just not needed (like `	unsigned m_HarmonicNumber; /* Harmonic Number */`) and loads of similar cases like `hash.Update(m_ReleaseInfo[i].sample_group); // Update Hash Release Sample Group`
3. Check the character encoding that you're using. Lots of characters now look wrong.
4. I don't like where you've inserted comments into statements that breaks the flow of code. Choose one way of inserting the comments, not all over the place before, in and after

I think that one of the big problems here lies in the fact that the release tail trunctation was always considered an User preference - not a sampleset creators - thus no such thing as allowing the sampleset creator, through the ODF, to influence what a user should have for settings was considered a good praxis. In the spirit of that I actually begin to doubt if we at all should have such settings available in the odf as such - but instead in a user settings file for the organ in question.

Anyway, the general way to proceed with the hierarchy of settings would be to from the parent populate the children if, and only if, they doesn't have any value specified for them.

pgcrowe:
Well that's just it, Lars. My philosophy gives the **user** more control, as truncation would not be saved permanently into a wave file, and all ODF values would remain alterable by the **user**.

larspalo (changes requested):
Date in void GOSoundProviderWave::SetAmplitude(float fixed_amplitude, float gain).

Look at the files in the PR on GitHub at [https://github.com/GrandOrgue/grandorgue/pull/774/files](https://github.com/GrandOrgue/grandorgue/pull/774/files) and compare the red and green lines in grandorgue.xml. Red is right encoding, green is wrong.

Excess comments in: `hash.Update(m_ReleaseInfo[i].sample_group); // Update Hash Release Sample Group`

Otherwise, I think you've done a great job!

pgcrowe:
Lars, you must consider the need for flexibility not the uses you can imagine, but the ones that you cannot imagine. For example, if I want to produce a "semidry" ODF for a sample set (with necessary custom release shaping on each pipe) as an alternative to a "wet" ODF, your method forces me to create an entirely separate sample set with truncation processed into the WAV file. Using the toolbar truncation menu produces inadequate results compared to custom settings for each pipe. Ultimately, it's a collective decision, which is not yours, and certainly not mine. I'm just offering forth this volunteered code at the suggestion of Oleg.

pgcrowe:
Yes, I am aware of the .XML. I pushed other changes, as noted in the last commit.

src/grandorgue/GOrgueSettings.cpp
```diff
@@ -117,10 +117,12 @@ GOrgueSettings::GOrgueSettings(wxString instance) :
 	ReverbFile(this, wxT("Reverb"), wxT("ReverbFile"), wxEmptyString),
 	MemoryLimit(this, wxT("General"), wxT("MemoryLimit"), 0, 1024 * 1024, GOrgueMemoryPool::GetSystemMemoryLimit()),
 	SamplesPerBuffer(this, wxT("General"), wxT("SamplesPerBuffer"), 1, MAX_FRAME_SIZE, 1024),
-	SampleRate(this, wxT("General"), wxT("SampleRate"), 1000, 100000, 44100),
-	Volume(this, wxT("General"), wxT("Volume"), -120, 20, -15),
+    /*  SAMPLE RATE — Enabled 192000 by changing 100000 to 200000.
```
oleg68, line 120: This change seems doesn't belong to the Release Length / Truncation. Please remove it from here and make a separate PR for it.

oleg68 (changes requested):

src/grandorgue/GOrgueSettings.cpp
pgcrowe: > This change seems doesn't belong to the Release Length / Truncation. Please remove it from here and make a separate PR for it.

Yes, that was an accidental push on my part, from my private build copy, sorry! It does work though and is worth implementing. If need be, I have 192 kHz samples you can test with.

oleg68:
@larspalo There are intermediate people between sample set producers and organ players: the ones, who adopt alien sample sets (mostly from Hauptwerk) for others need or who make composite sample sets.

For example, If I already have a relativelly dry sample set and I'm adding some stops from another (wet) sample set to the first one, I need a capability to limit release length of the new stops in ODF, not in the organ settings.

So I'd add this capability to ODF.

oleg68:
@pgcrowe The encoding in the hlelp file is still odd. Please use UTF-8.

help/grandorgue.xml
```diff
@@ -9962,6 +9962,15 @@ release (or other attacks). 0 means automatic selection.
           </listitem>
         </varlistentry>
       </variablelist>
+      <varlistentry>
+          <term>Pipe999ReleaseTruncationLength</term>
+          <listitem>
+            <para>
+(int 0 - 10000, default: 0) Release truncation length in ms, to shorten the length of releases.
```
oleg68, line 9969: 1. This solution does not allow to disable the release at all. Maybe -1 is better?
2. How to set the release truncation on the Rank level? Maybe ReleaseTruncationLenght?

src/grandorgue/GOSoundEngine.cpp
```diff
@@ -576,27 +596,39 @@ void GOSoundEngine::CreateReleaseSampler(GO_SAMPLER* handle)
...
+                    gain_decay_length = m_ReleaseLength;
+            }
+
+            /* If truncation values exist and release scaling has not been enabled, gain_decay_length = truncation value. */
+            else if (truncation_fade_len > 0 && gain_decay_length == 0)
+            { 
+                gain_decay_length = truncation_fade_len;
+            }
+
+            if (gain_decay_length > 0)
```
oleg68, line 620: The lines from here have not been changed. Please leave them in the original state (maybe it is a tab stuff)

src/grandorgue/GOSoundEngine.cpp
```diff
@@ -576,27 +596,39 @@ void GOSoundEngine::CreateReleaseSampler(GO_SAMPLER* handle)
...
-			if (m_ReleaseAlignmentEnabled && release_section->SupportsStreamAlignment())
-			{
-				release_section->InitAlignedStream(&new_sampler->stream, &handle->stream);
-			}
-			else
-			{
-				release_section->InitStream(&m_ResamplerCoefs, &new_sampler->stream, this_pipe->GetTuning() / (float)m_SampleRate);
-			}
-			new_sampler->is_release = true;
+            /*  Algorithm to Choose Which Release Truncation or Scaling Method to activate. */
```
oleg68, line 601: It is more clear to move this comment and assignment to truncation_fade_len below calling NewAttacking

src/grandorgue/GOSoundEngine.cpp
```diff
@@ -576,27 +596,39 @@ void GOSoundEngine::CreateReleaseSampler(GO_SAMPLER* handle)
...
+			
+			/* If Release Length Set in GO GUI toolbar is larger than 0 */
+            if (m_ReleaseLength > 0)
+            {
+                /* If Gain Decay Length (gain_decay_length) is greater than toolbar value (m_ReleaseLength), or equal to 0, gain_decay_length = toolbar value. */
+                if ( m_ReleaseLength < gain_decay_length || gain_decay_length == 0 )
+                    gain_decay_length = m_ReleaseLength;
+            }
+
+            /* If truncation values exist and release scaling has not been enabled, gain_decay_length = truncation value. */
```
oleg68, line 614: This behaviour seems illogic: If the toolbar value is activated then we totally ignore ReleaseTruncationLength.

Taking the lowest value among m_ReleaseLength, gain_decay_length and truncation_fade_len would be better

src/grandorgue/GrandOrgueFile.cpp
```diff
@@ -893,6 +893,11 @@ void GrandOrgueFile::UpdateAmplitude()
 {
 }
 
+// Update Release Truncation Length Added 12-9-20
```
oleg68, line 896: Remove this comment

src/grandorgue/OrganDialog.cpp
```diff
@@ -705,6 +754,15 @@ void OrganDialog::OnEventApply(wxCommandEvent &e)
 		return;
 	}
 
+	    // TRUNCATION SETTINGS — Release Sample Truncation Settings
+    if (!m_ReleaseTruncationLength->GetValue().ToLong(&truncation) &&
+        (m_ReleaseTruncationLength->IsModified() &&
+         (truncation < 0 || truncation > 10000)))
```
oleg68, line 760:
```suggestion
         (truncation < 0 || truncation > 100000)))
```

oleg68 (changes requested):

oleg68:
@pgcrowe First of all, please correct your tabulation settings. The standard ones, compatible with every text editors is ``1 tab = 8 spaces``. It will remove lots of line from diff.

help/grandorgue.xml
larspalo: Normally -1 is indicating a default, unchanged value.

larspalo:
> > @larspalo There are intermediate people between sample set producers and organ players: the ones, who adopt alien sample sets (mostly from Hauptwerk) for others needs or who make composite sample sets.
> > For example, If I already have a relativelly dry sample set and I'm adding some stops from another (wet) sample set to the first one, I need a capability to limit release length of the new stops in ODF, not in the organ settings.
> > So I'd add this capability to ODF.
> 
> Yes, thank you, Oleg, I couldn't agree more! You've explained this need so much better than I can.

Then, by my standard, you're no sampleset producer.

One of the reasons for having the trunctation as an user setting (for now mainly in the .cmb, not in the .odf) is mainly that if you want to have multiple versions for an organ - you can just import the settings without reloading the whole organ with all the samples.

For the users wishing to make composite organs I think we should adopt another long term solution. This is perhaps for another topic really, but anyway... It's already now possible to create organs that depend on other organs - that is, they use resources from other organs. Perhaps we should expand on this approach and make it easier for users to pick from the existing organs on their system and compose a "new" organ of their liking. Then configure it with appropriate user settings.

My thoughts are: the user should never need to bother with writing odfs and learning it's specs, that's the sampleset producers burden. GrandOrgue as a software should make it easy for the user to use existing resources in whatever way they like.

Do you see the beauty of it?

help/grandorgue.xml
oleg68: I'd withdrow my suggestion (1). All other related settings have the default value `0` that means not to limit the release length. So I agree with the current solution to have `0` by default. Otherwise we should change the old settings as well.

oleg68:
> My thoughts are: the user should never need to bother with writing odfs and learning it's specs, that's the sampleset producers burden. GrandOrgue as a software should make it easy for the user to use existing resources in whatever way they like.
> 
> Do you see the beauty of it?

You do not consider composite set creators as sampleset producers and consider them as users. It is your right.

But in this case I'd differenciate between two kinds of users: ordinar users - organ players and users who makes organs for other users or convert them from Hauptwerk format.

I agree with you that learning and modifying ODFs is not necessary for the first kind of user. But for now it is the only possible way for the second kind of users. And we must provide them ODF capabilities until we suggest them other acceptable way to do their work.

pgcrowe:
> Dear all,
> 
> I have followed the whole discussion thread with interest, and even if I did not understand everything (English is not my mother language), I allow myself a comment. My deepest apologies in advance if i am off topic.
> 
> To integrate a possibility of homogenization of the sounds in GrandOrgue, in the case where an organ is created from samples of different organs, seems to me an excellent thing and a tremendous progress for the program.
> 
> afterwards, as I read in the thread, the organ installed will sound different, whether it is in a living room, a church, etc. it indeed seems important to me that the end user can have control over the final rendering.
> 
> homogenization between samples in the ODF file should therefore be done "at the minimum"
> 
> * sound level so that disparate stops have the same sound level wherever they come from
> * homogeneous reverb
> * homogeneous temperament
> 
> afterwards, it would be very interesting and a real "plus" if GrandOrgue allowed the user to harmonize the stops note by note, via a mixer for example, where the user could interact himself on the harmony of the stops , under its own responsibility.
> 
> forgive me the length of the message and the possible off topic ...

I'm so happy to hear your kind comments, Jean-Gabriel, thank you!

src/grandorgue/GOSoundEngine.cpp
pgcrowe: I mentioned that as an optional implementation in the first post of this thread. I'll change it.

pgcrowe:
@oleg68  I just compiled and tested your new algorithm that automatically determines release length truncation value.  You're a genius!, it works flawlessly! 

The remaining issue is to find a way to overwrite/auto populate parent truncation values into child values in the organ settings panel. At present, while setting a parent values overrides child values, the parent does not change child values in the GUI. I suspect this must be done by altering GOrguePipeConfigNode.cpp or OrganDialog.cpp with the, but I'm stumped as to how to achieve it.

pgcrowe:
As a provisional measure for pipe999release999TruncationLength, I have added description comments to GOrgueSoundingPipe.cpp, to label the purpose of each pipe999Parameter999 section at a glance. Even just adding "pipe999release999" and "pipe999attack999" etc in written comments makes it so much easier to find their origin code.
Ergo:
```
	/* This Section defines pipe999release999 parameters in ODFs.
	 * In this section, wxT("Value") sets the pipe999release999Value ODF call names. */
```

larspalo:
May I ask that you hold your horses on this a little and think things through. In the organ settings dialog there are only the main pipe available in the tree. It was discussed a long time ago (regarding pitch info, at that time) if it was necessary to have it on a per attack/release level or not. The choice was not - the main reason was that every attack/release was supposed to have been recorded at the same time (and place) so the needed info could be read from just the main sample, not from every version of attack/release.

If you now are talking about needing to match releases from so different pipe recordings within the same pipe that you need to trunctate them differently for reverb matching reasons, I believe you're now thinking about doing something very wrong indeed! ;-)

However, this train of thoughts, is now approaching a completely different subject - the release tail shaping as feature within GO depending on playing time (a feature to imitate shorter reverb tail on staccato playing) known as "Realease sample scaling". Make sure we're differentiating between the features and know exactly what their purposes are. Generally - the release sample scaling is a "poor mans version" of (not) having multiple releases available of different lengths - and was basically not supposed to be used when multiple releases was available.

pgcrowe:
> If you now are talking about needing to match releases from so different pipe recordings within the same pipe that you need to trunctate them differently for reverb matching reasons, I believe you're now thinking about doing something very wrong indeed! ;-)
> 
Heavens no. Besides, it would take far more than per-release controls to attacks and releases from two different ranks!

> May I ask that you hold your horses on this a little and think things through. In the organ settings dialog there are only the main pipe available in the tree. It was discussed a long time ago (regarding pitch info, at that time) if it was necessary to have it on a per attack/release level or not. The choice was not - the main reason was that every attack/release was supposed to have been recorded at the same time (and place) so the needed info could be read from just the main sample, not from every version of attack/release.
> 
In my opinion, pipe999release controls are probably best left in the ODF only. If added to the GUI, I fear it would add clutter and confusion for the end user, for a function that is almost solely developer-centric.

You have to realize, while important, pipe999release999 controls would cope with incredibly subtle issues. Issues so subtle, that  the majority of users typically would miss the detail unless trained to listen for it. pipe999release999 aren't something something that even I would heavily utilize. But for the problems could solve, it would be an invaluable addition.

One application for pipe999release999 adjustable crossfade can be found with certain forms of clarinet (such as Hope Jones, E.M. Skinner) and Cornopean (such as Æolian) produce a louder release with short sustain than long sustain. Adjustable crossfade per release would allow adequate adjustment for both, rather than having to set a compromise. Another application for adjustable crossfade is compensation for acoustical anomalies such as slapback, that can intrude upon one release length, but not another. 

> However, this train of thoughts, is now approaching a completely different subject - the release tail shaping as feature within GO depending on playing time (a feature to imitate shorter reverb tail on staccato playing) known as "Realease sample scaling". Make sure we're differentiating between the features and know exactly what their purposes are. Generally - the release sample scaling is a "poor mans version" of (not) having multiple releases available of different lengths - and was basically not supposed to be used when multiple releases was available.> 

It would be great if there was a way to integrate adjustable Release Sample Scaling control triggers into the ODF. Actually,, I think it would be a game changer to integrate an enhanced release sample scaling mechanism designed to work with multi release — Ergo: a fade algorithm to help bridge a seamless transition between the small gaps in reverb length between short, mid, and long releases.

pgcrowe:
> While testing this I realized that the naming is rather counter intuitive. Rel. Truncation (ms) is in effect a decision on how long the release tail is at maximum, not how much it's shortened. The naming should thus be something more like "Max. Release Length (ms)". In that context it would make sense to have a (blank) default at -1, which equals no modification and a setting of 0 is a release lenght of 0.
> 
I totally agree with your logic here, and am in favor of this change. I used "truncation" simply for sake of familiarity, because of Hauptwerk's existing usage of the term, which, admittedly, is likewise counterintuitive. However, even with that, I still have worried that the word "truncation"  is not intuitive, or familiar enough among native English speakers, or otherwise. Other non-organ samplers use the term "Release Length, "Max Release Length," or similar.

> Also there are a few compiler warnings that should be addressed, both on comparing unsigned vs signed int and some indentations close to if statements.
> 
> `/src/grandorgue/GOSoundEngine.cpp:577:57: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare] 577 | if (truncation_fade_len > 0 && truncation_fade_len < time_to_full_reverb)`
> 
> `/src/grandorgue/OrganDialog.cpp: In member function ‘void OrganDialog::Load()’: /src/grandorgue/OrganDialog.cpp:498:4: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation] 498 | if (entries.size() == 1) | ^~ /src/grandorgue/OrganDialog.cpp:500:4: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’ 500 | m_ReleaseTruncationLengthSpin->SetValue(truncation);`>

Thank you so much for bringing these compiler errors to our attention! What was the the target OS for that compile?
Neither warning appeared while compiling on macOS 11 Big Sur with macOS target, so I'll likely need to change target to debug.

A number of other warnings did appear, unrelated to changes in this commit. **See the attached file for macOS Terminal output on compile.** Extraneous folder paths have been "truncated hahaha" for clarity. Most warnings pertain to PortAudio attributes that have been depreciated in macOS. Since much of the source in /src/portaudio is over 7 years old, I'm guessing that most issues would be resolved by updating src/portaudio to the latest stable version, 19.7.0 (April 6, 2021). Yes, you read that right, the first new stable release in 5 years!
[Terminal Saved Output.txt](https://github.com/GrandOrgue/grandorgue/files/7393407/Terminal.Saved.Output.txt)

pgcrowe:
**Update: ** I did a compile test with the latest stable version of Port Audio (19.7.0) on Mac, with Mac target. The source compiled without errors, and the number of warnings reduced from 116 to 81. If someone can test compiling for Linux and Windows without issues, I'd recommend upgrading the Port Audio library to 19.7.0 in a separate commit.

rousseldenis:
@pgcrowe pre-commit has been enabled now. You can install it on your local git repo (pre-commit install), then, at each commit you do, pre-commit is run and correct your changes syntaxically.

oleg68:
Duplicated by #1380

PR #778, opened 2021-10-15
Fix id/linkend typo in help file

larspalo:
Fix error that crept up while compiling the source from within Eclipse.


oleg68:
@rousseldenis Lets not keep pending PRs for long time.

rousseldenis (approved):
15 hours is not a long time ;-)

Issue #702, opened 2021-08-10
GO breaks sound after high playing


oleg68:
This issue exists for a lot of time since very old versions.

When playing GO with lots of stops and notes and  high reveberation to a jack output (Rt:jack or Jack:native) sometimes sound stops. Trying to do Panic or to exit GO causes that GO hangs.

oleg68:
After adding a diagnostic warning messages in #762, when the sound is brocken, there warning messages appeared in the log window:

```
Вс 17 окт 2021 18:23:51: 18:23:51: Warning: GOMutex: timeout when locking a mutex; currentLocker= newLocker=GOSoundGroupWorkItem::Run.beforeProcess
Вс 17 окт 2021 18:23:51: 18:23:51: Warning: GOMutex: timeout when locking a mutex; currentLocker= newLocker=GOSoundGroupWorkItem::Finish
Вс 17 окт 2021 18:23:51: 18:23:51: Warning: GOMutex: timeout when locking a mutex; currentLocker= newLocker=GOSoundGroupWorkItem::Run.beforeProcess
Вс 17 окт 2021 18:23:51: 18:23:51: Warning: GOMutex: timeout when locking a mutex; currentLocker= newLocker=GOSoundGroupWorkItem::Finish
```

The mutex could not be obtained in 1 second, but there were no any clear reason: there was no current locker. Moreover, the problem has been changed: earlier the sound stopped totally, now the sound returns after one second. It means that the mutex is obtained successfuly at the second attempt.

Both these facts suggest that the root cause of this problem is wrong implementation of GOMutex. Sometimes it cann't be locked when it is not locked by any other thread.

Now there are two alternative implementation of GOMutex and GOCondition in the GO source code: with the wxWirgets classes wxMutex, wxCondition and with GOWaitQueue. The implementations are switched by `WX_MUTEX` compilation variable.

GOWaitQueue is not a standard primitive and itself has two alternative implementations: with wxSemaphore and with std::mutex (has been switched to std::timed_mutex as part of #784).

Now by default GOMutex uses GOWaitQueue that uses wxSemaphore.

This code is too complex and it is hard to troubleshoot. Most likely it has a bug, that causes this issue.

When GO was created, there were no any standard primitives std::mutex and std::condition_variable. But [they appeared in `C++11`](https://en.cppreference.com/w/cpp/thread).

[...]

PR #784, opened 2021-10-17
Fixed hang on Panic button after the sound had been suddenly stopped

### LINKED ISSUE #762: Panic is not able to reset sound system if some thread is waiting for a mutex.

oleg68:
Resolves #762

1. Added new functions of threading primitives GOMutex::LockOrStop and GOCondition::WaitOrStop. They wait not infinitely, but check the current thread every second: whether it has been marked for stop (with Panic button)
2. GOSoundThread, GOSoundGroupWorkItem and GOSoundOutputWorkItem started using the new functions. It allows them to finish when `Panic` button has been pressed
3. Added writing a warning to the log window with the informations about the locker and the waiter, if the mutex cann't be obtained or the condition cann't be satisfied in one second. It will help to diagnose #702 

PS. This change has already helped me to localise #702. https://github.com/GrandOrgue/grandorgue/issues/702#issuecomment-945163575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
+- Fixed hang on Panic button after the sound had been suddenly stopped https://github.com/GrandOrgue/grandorgue/issues/762
+- Make the Paths settings in the Options tab expandable
 - Added build for linux appimage https://github.com/GrandOrgue/grandorgue/issues/698
 - Fixed bug preventing building with -DGO_USE_JACK=OFF
 - Moved the netbeans project files to the ide-projects/NetBeans12 subdirectory https://github.com/GrandOrgue/grandorgue/discussions/771
-- Make the Paths settings in the Options tab expandable
 # 3.3.0 (2021-10-08)
 - Added automated release tagging on GitHub https://github.com/GrandOrgue/grandorgue/issues/711
 - Added polish language support https://github.com/GrandOrgue/grandorgue/discussions/743
--- a/src/core/threading/GOCondition.cpp
+++ b/src/core/threading/GOCondition.cpp
@@ -89,7 +89,7 @@ unsigned GOCondition::DoWait(bool isWithTimeout, const char* waiterInfo, GOrgueT
     // a timeout occured
     if (isFirstTime && waiterInfo)
     {
-      wxLogWarning("GOCondition::Wait: unable to restore mutex lock for %s", wxString(waiterInfo));
+      wxLogWarning("GOCondition::Wait: unable to restore lock on the condition mutex %p:%p for %s", this, &m_Mutex, wxString(waiterInfo));
       isFirstTime = false;
     }
   } while (pThread == NULL || !pThread->ShouldStop());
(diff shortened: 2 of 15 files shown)


larspalo (commented):
Just a few things to think about.

Naming of the header file `#include "threading_impl.h"`. Perhaps all the files should carry the GO identifying start, and if we continue the trend it might be GOThreadImpl.h or something similar.

English spelling of infinitely (void GOWaitQueue::WaitInfinitelly()) perhaps should be corrected.

When we're talking commenting, cases where not obvious operations are performed might benefit from comments explaining what's actually done (and why) like with the bitwise OR in `unsigned compose_result(bool isSignalReceived, bool isMutexLocked)` especially when used in conjunction with ternary operator. This is definitely not newbie friendly stuff, while very compact and concise indeed.

When you have situations with just one line after a condition and the code continues in blocks it's not immediately obvious how you intend the comment. For instance in:
`if (isMutexLocked)
      break;
    // a timeout occured
    if (isFirstTime && waiterInfo)
    {`

I can read from indention the comment belong to upcoming condition, but wouldn't it be nicer with a blank line in between? Many times I now personally prefer to put brackets after conditions even when there's only one line following, but I don't mind other peoples practices... But I remember that Nick Appleton once told me that:

> Whitespace is your friend.

You have a very nice example in unsigned `GOCondition::WaitOrStop(const char* waiterInfo, GOrgueThread* pThread)` of excellent use of whitespace to make the code more readable.

None of my comments are of course vital for the functionality of this PR.

oleg68:
> Naming of the header file `#include "threading_impl.h"`. Perhaps all the files should carry the GO identifying start, and if we continue the trend it might be GOThreadImpl.h or something similar.

As I understand the name GOThreadImpl.h is used when there is a **class** GOThreadImpl defined in the files. Otherwise the name I think should be another.

I use lowercase name for a file without any classes. If you like, I can add the `go_` preffix to it: `go_thread_impl.h`. But I wouldn't like name it `GOThreadImpl.h` because there is no class GOThreadImpl inside.

> English spelling of infinitely (void GOWaitQueue::WaitInfinitelly()) perhaps should be corrected.

Will be corrected. I hope this class GOWaitQueue [will be got rid soon](https://github.com/GrandOrgue/grandorgue/issues/702#issuecomment-945163575)

> cases where not obvious operations are performed might benefit from comments explaining what's actually done (and why) like with the bitwise OR in unsigned compose_result

A comment will be added

larspalo (approved):

oleg68:
@larspalo @rousseldenis I added some changes **after** you had approved this PR. Could you have a look?

oleg68:
> This is interesting. You say that it happens when exactly what audio output group/channel routing is done? Multiple groups send to same channels?

No. When one channel of a group is routed to several audio output channel.

The simple such configuration includes only one audio group. Like thi
BuggyAudioDistribution
s

PR #811, opened 2021-10-29
Changed log error to warning when importing combinations.

### LINKED ISSUE #809: Extra error: "The ODF does not match the combination file"
When I try to import a combination saved from a previous version of ODF, the combination is loaded successfully, but the error message "The ODF does not match the combination file" appears in the log window.

1. I import a combination into the organ, not the organ into the combination. So this message is not precise. ``The combination file does not match the current ODF`` sounds more precisely.
2. Because the combination is loaded anyway, the message should be a warning, not an error.

larspalo:
Resolves #809

### DIFF
--- a/src/grandorgue/GrandOrgueFile.cpp
+++ b/src/grandorgue/GrandOrgueFile.cpp
@@ -612,12 +612,12 @@ void GrandOrgueFile::LoadCombination(const wxString& file)
 
 		wxString church_name = cfg.ReadString(CMBSetting, wxT("Organ"), wxT("ChurchName"));
 		if (church_name != m_ChurchName)
-			throw wxString::Format(_("File belongs to a different organ: %s"), church_name.c_str());
+			wxLogWarning(_("This combination file was originally made for: %s"), church_name.c_str());
 		wxString hash = odf_ini_file.getEntry(wxT("Organ"), wxT("ODFHash"));
 		if (hash != wxEmptyString)
 			if (hash != m_ODFHash)
 			{
-				wxLogError(_("The ODF does not match the combination file."));
+				wxLogWarning(_("The combination file does not exactly match the current ODF."));
 			}
 		/* skip informational items */
 		cfg.ReadString(CMBSetting, wxT("Organ"), wxT("ChurchAddress"), false);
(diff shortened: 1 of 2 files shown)


larspalo:
As I tested importing only combinations to changed odfs I only had issues when I had removed a stop from the odf - in such a situation I get an error log message and the combinations won't import. New stops that I had added behaved properly - they were simply not set. Existing but with changes had no issues either.

To correct the import for an odf which has a removed stop, I think we have to change the parsing behaviour (to just ignore a cmb section when it's not present in the odf). But that's perhaps another issue?

Interestingly enough, I didn't have any issues when importing just the settings (with #810 of course) so the problem with removed stops are limited to if existing combination entries are present in the frame generals.

oleg68:
> To correct the import for an odf which has a removed stop, I think we have to change the parsing behaviour (to just ignore a cmb section when it's not present in the odf). But that's perhaps another issue?

Yes, indeed, another issue.

src/grandorgue/GrandOrgueFile.cpp
```diff
@@ -619,13 +619,16 @@ void GrandOrgueFile::LoadCombination(const wxString& file)
 		if (hash != wxEmptyString)
 			if (hash != m_ODFHash)
 			{
-				wxLogWarning(_("The combination file does not exactly match the current ODF."));
+				if (wxMessageBox(_("The combination file does not exactly match the current ODF. Importing it can cause various problems. Should it really be imported?"), _("Import"), wxYES_NO, NULL) == wxYES)
```
oleg68, line 622: Clicking every time to OK is not good.

Can you display the message box only if the churche name has changed (probably this combinations are from a totally another organ)? 

If the churche name is the same but hash is different then only to log warning?

larspalo: Sure, it's certainly possible. I'll fix that tomorrow!

larspalo: I've added a new solution that works better.

oleg68 (approved):

rousseldenis (approved):

PR #815, opened 2021-10-30
Fixed restoring the size and the position of panels

### LINKED ISSUE #787: Organ panels position not memorised
Hello,
I don't know if it's a grandorgue bug or a bug in my computer, but when I start grandorgue, the position of my panels is never remembered.

I am using Linux Mint with grandorgue version 3.2.0.1, and two touch screens. When the computer starts up, the two panels with stops (right jamb and left jamb) are placed on my right screen, while the main menu of grandorgue is placed on the left screen.

Systematically, I move my left panel to the left screen then I save the settings, but when restarting it is identical the two panels go back to the right screen ...

oleg68:
Resolves #787

1. Eliminated restoring the original aspect ratio. Now the saved size is used
2. Saving display numbers and maximizing flags. Using them on restore
3. Renamed Size to Rect somewhere

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed restoring the size and the position of panels when they did not fit their displays https://github.com/GrandOrgue/grandorgue/issues/787
 - Removed translation of GrandOrgue midi ports https://github.com/GrandOrgue/grandorgue/issues/791
 - Switched the build for windows to use wxWidgets 3.1.5 https://github.com/GrandOrgue/grandorgue/issues/792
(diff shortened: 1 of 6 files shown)


larspalo:
The ReadFromCfg naming is a bit vague and general in my opinion, especially since its reading is quite specific and only used for one purpose. I'd consider renaming it to something like ReadSizeInfoFromCfg or anything more descriptive.

Good idea to store the maximized flag.

It seems to work as expected, including that the up-scaling is very poor.

larspalo (approved):

rousseldenis (approved):

PR #837, opened 2021-11-04
Added implementation of the threading primitives with standard C++11 ones

### LINKED ISSUE #817: Implement GO threading primitives over the C++11 standard ones
Now there are two alternative implementation of GOMutex and GOCondition in the GO source code: with the wxWirgets classes wxMutex, wxCondition and with GOWaitQueue. The implementations are switched by WX_MUTEX compilation variable.

GOWaitQueue is not a standard primitive and itself has two alternative implementations: with wxSemaphore and with std::mutex (has been switched to std::timed_mutex as part of #784).

Now by default GOMutex uses GOWaitQueue that uses wxSemaphore.

This code is too complex and it is hard to troubleshoot. Most likely it has a bug, that causes #702

When GO was created, there were no any standard primitives std::mutex and std::condition_variable. But they appeared in C++11.

So seems the best solution is to rewrite the current implementation of GOMutex and GOCondition with std::timed_mutex and std::condition_variable without capability of switching to another impementations.

oleg68:
Resolves #817 

The new implementation is quite simple.

I keep the capability of switching to the old implementstion by undefining GO_STD_MUTEX in cmake

Unfortunally I cann't test it with high playing and jack until the Sunday.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Added implementation of the threading primitives with standard C++11 ones https://github.com/GrandOrgue/grandorgue/issues/817
 # 3.4.1 (2021-11-02)
 - Changed log error to warning when importing combinations https://github.com/GrandOrgue/grandorgue/issues/809
 - Eased restrictions on importing .cmb settings https://github.com/GrandOrgue/grandorgue/issues/807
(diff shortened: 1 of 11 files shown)


oleg68:
I've tested https://github.com/oleg68/GrandOrgue-oficial/actions/runs/1418196304 with playing about 15 min. No hangs or any other issues appeared.

This evening I will be able to do more tests either on both linux and windows.

oleg68:
I tested this on windows with rt/asio sound port an on linux with jack:Native Output sound port in a time about one hour. GO worked fine in both cases. No problems described in #702 occured.

oleg68:
@larspalo @rousseldenis could you approve this. I've tested it and it seems to work.

larspalo (approved):

rousseldenis (approved):

Issue #816, opened 2021-10-31
Rename classes and source files to have the `GO` prefix


oleg68:
Aggording to #793 it has decided, that all GrandOrgue classes should have the single `GO` prefix.

1. All classes and files started with `GOrgue` should be renamed to start with `GO` (remove `rgue`).
2. All classes and files started with `GrandOrgue` should be renamed to start with `GO`.
3. The files should be renamed to correspond to the class names
4. The names of header files with these prefixes not containing any classes should be converted to lowercase with the prefix `go_`
5. Not to forget to rename include protection variables: 

    ```
    #ifndef GORGUESTOP_H
    #define GORGUESTOP_H
    ```

    to

    ```
    #ifndef GOSTOP_H
    #define GOSTOP_H
    ```

Issue #834, opened 2021-11-03
Redistribute GO source files across directories.


oleg68:
As discussed in #793, the source code of GO should be relocated to the following directory structure:

- src
    - core
        - config: `*Config*`
        - gui: `*GUI*`
        - settings `*Settings*`
        - sound
            - scheduler: `GOSoudTouchWorkItem.cpp`, `GOSoudTouchWorkItem.h` and `GOSoudWorkItem.h`
    - grandorgue
        - gui: `*GUI*`
        - midi: `*MIDI*`
        - settings: `*Settings*`
        - sound: `*Sound*`
            - scheduler: `GOSoundScheduler`, `GOSoundThread` and all descendents from `GOSoudWorkItem` 
            - ports: `*Sound*Port*`
    - tools: GrandOrgueTool, perftest

The `#include` directives and the `CMakeLists.txt` files should be modified according to the new file layout.

PR #839, opened 2021-11-06
Renaming GOrgue classes and files

oleg68:
Resolves: #816

Ther are lots of changes but they are quite simple: only renaming classes and files.


larspalo:
This was a hard one to read! Around file 106 the diff messes up with some of the files in what file actually changed name to what and that definitely makes more difficult than it usually is to follow along in the changes...

larspalo:
There are (quite) a few files that have no prefix, mainly the ones for parts of different dialogs. They are kind of grouped together as they are already, but what are your thoughts about them eventually?

Also, quite understandably you have not yet implemented the folder re-structuring either... But that comes next, right?

oleg68:
> This was a hard one to read! 

Yes, agree, That is why I didn't make any extra changes exept renaming.

> Around file 106 the diff messes up with some of the files in what file actually changed name to what and that definitely makes more difficult than it usually is to follow along in the changes...

Yes, but renaming classes requires changing in all files where they are used.

Sometimes GitHub does not catch the renaming and shows add/remove instead. I do not know how to fix it.

> There are (quite) a few files that have no prefix, mainly the ones for parts of different dialogs. They are kind of grouped together as they are already, but what are your thoughts about them eventually?

I agree with adding the prefix to the files and classes that do not have it, but it would make this PR harder to review. So I made only changes in the scope of #816 and nothing else. I'd prefer to add the prefix in a separate issue.

The one exception is the `GrandOrgueFile`. Because another class `GOrgueFile` already existed in the `core` subdirectory, I couldn't rename both classes to `GOFile`. So I renamed `GOrgueFile` to `GOFile` and `GrandOrgueFile` to `GODefinitionFile`.

> Also, quite understandably you have not yet implemented the folder re-structuring either... But that comes next, right?

Yes, I haven't it. I wait for merging this PR and then I'll take #834 in work.

oleg68:
@larspalo @rousseldenis Could you approve this? Keeping this PR unmerged will raise a lot of conflicts in the future.

larspalo (approved):

oleg68:
@rousseldenis please, approve this PR.

rousseldenis (approved):

PR #844, opened 2021-11-11
Restructure source code directories

oleg68:
Resolves: #834


oleg68:
@larspalo @rousseldenis 

I understand that changing lots of files is hard to check and review, but they all are very simple: just moving and adding the subdirectory in `#include` statements.

Keeping such large PR without merge will cause a lot of conflicts. Su could you check and approve this PR?

larspalo (approved):
At least in this PR GitHub didn't mess up the file names so much!

oleg68:
> At least in this PR GitHub didn't mess up the file names so much!

I made two commits instead the one: the first - only moving files into subdirectories without modifying any files, and the second - midifying `#include` statements and `CMakeList.txt`.

Seems GitHub does not like renaming and modifying the same files in a single commit.

oleg68:

rousseldenis (approved):

Issue #847, opened 2021-11-14
Crash to desktop Win10 large registration


mutti-wilson:
@oleg68 asked for a crashdump to see what's going on. Crashes after playing large registration.

I'm using the latest 3.4.2-0.2 build by @oleg68

[GrandOrgue.exe.8792.zip](https://github.com/GrandOrgue/grandorgue/files/7533203/GrandOrgue.exe.8792.zip)

oleg68:
@mutti-wilson Please try to use https://drive.google.com/file/d/1vlak-ZYcVHXUYTADA54FTitCt6qRCMV5/view?usp=sharing
(without installation; just unpack it to some directory and run `bin\GrandOrgue.exe`) and post here the new crash dump from this version.

mutti-wilson:
Here are 2 crash dumps. I don't know why they have a duplicate file name since the time generated is the same but they are slightly different sized so I thought sending both would be wis
[CrashDumps.zip](https://github.com/GrandOrgue/grandorgue/files/7536139/CrashDumps.zip)
e.

mutti-wilson:
So if it matters I have a little info to add as well as another crash dump. The first crash was using WDM drivers, second was ASIO and this crash was using Direct Sound. Direct sound lasted the longest before crashing. If you need any other background let me know and I'll do what I can to answer. Thanks for digging in on this.

[CrashDumps.zip](https://github.com/GrandOrgue/grandorgue/files/7543688/CrashDumps.zip)

oleg68:
Thank you, @mutti-wilson I will discover your crash dumps.

oleg68:
Asio:
```
   ExceptionCode: c0000005 (Access violation)
Attempt to read from address 000000000000008c

00000000`008ff0a8 00000000`004afc6b : 00000000`0fa17860 00000000`0000000e 00000000`00000000 00000000`0000000f : GrandOrgue!GetVelocityVolume+0xb
00000000`008ff0b0 00000000`0fa17860 : 00000000`0000000e 00000000`00000000 00000000`0000000f 00000000`0fa17860 : GrandOrgue!UpdateVelocity+0x1b
```

oleg68:
DirectSound:

```
00000000`008ff0a8 00000000`004afc6b : 00000000`0fa17860 00000000`0000000e 00000000`00000000 00000000`0000000f : GrandOrgue!GetVelocityVolume+0xb
[...]

PR #860, opened 2021-11-17
Fixed crash when playing with shared pipes

oleg68:
Resolves: #847

The reason of crash was in GOSoundEngine::UpdateVelocity: handle->pipe could become NULL in a parallel thread

Also it could be switched to another pipe, so I added passing the original pipe and checking. This code is not still safe: sometimes we can update velocity of a wrong pipe.  But I think the functionality of velocity is used very seldom, so I fixed only the possibility of a crash but not the possiblity of a wrong behavior.

The same possibility of a wrong behavior exists in lots of another methods of GOSoundEngine: they all check the passed pipe and do some action but do not protect against altering the pipe between the checking and the action. Fixing this behavior would require introducing a mutex in GO_SAMPLER and locking it. If another GO_SAMPLER-concurency related bugs appears we can make a more accurate locking with a mutex.

I also noted that the class name GO_SAMPLER doesn't conform our naming convention. I will rename it in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Fixed crash when playing with a shared pipe https://github.com/GrandOrgue/grandorgue/issues/847
+- Fixed a possible crash when playing with shared pipes https://github.com/GrandOrgue/grandorgue/issues/847
 # 3.4.2 (2021-11-14)
 - Added ASIO logo to About splash screen https://github.com/GrandOrgue/grandorgue/issues/823
 - Upgraded PortAudio Library to the latest stable release, v19.7.0. The upgraded library corrects PortAudio errors in GrandOrgue on macOS 11.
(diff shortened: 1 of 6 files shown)


larspalo:
Normally, velocity values from MIDI keyboards would rarely be used with organs. I've used attack velocity on occation to select what attack to play, and velocity volume on noise effects like mechanical action noises where a harder hit would result in louder noise. But there has been an Orgchestre sampleset done earlier with orchestral "stops" that used velocity volume a lot. So, I do think this would be worth fixing if it's indeed an issue.

oleg68:
> So, I do think this would be worth fixing if it's indeed an issue.

@larspalo Probability of wrong behavior existed earlier. Having introduced checking of the pipe I reduced this probability and fully excluded possibility of crash. Do you instist on a more strong fixing now or not?

The strong fixing would not be complex technichally, but may impact on the performance like all things with treading synchronisation.

larspalo:
> The strong fixing would not be complex, but may impact on the performance like all things with treading synchronisation.

No, it's not necessary to do now. When it becomes an issue we'll deal with it.

larspalo (approved):

oleg68:

oleg68:
> But there has been an Orgchestre sampleset done earlier with orchestral "stops" that used velocity volume a lot.

I've seen a piano sampleset that also used velocity. But this code is called only when the same pipe is used in more than one rank or from a copulation that is unusually for pianos. May be exist only in som very exotic instrument,

rousseldenis (approved):

PR #890, opened 2021-11-29
Feature/midi enchancement p1

oleg68:
The first part of changes for $703. It is just renaming and moving.

1. The class `GOSoundPortsConfig` has been renamed to `GOPortsConfig` and has been moved from `sound/ports` to `settings`. Because the same class will be used for Midi config.
2. The classes `GOMidiFactory` and `GOMidiRtFactory` have been renamed to `GOMidiPortFactory` and `GOMidiRtPortFactory`. Because they will inherit from a new class GOPortFactory that will be used for both sound and midi

No behaviour should be changed. I've made a separate PR because it would be difficult to review this lot of very simple changes among essential changes.


larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #796, opened 2021-10-24
Choice for checking midi device


paolo2504:
This is absolutely a poor nice to have request...
In some past release, it was correctly applied the possibility for GO to show the reason for showing error on opening the Settings window.
That's fine.
I ask now if it is possible to set some option in Settings to enable or not the MIDI device checking; something like "On startup always check if MIDI device is attached Yes/Not". This to prevent the opening of the Settings window plus the error reason dialog.
This way, GO will also be faster on restart i guess.

Note: why not to check a MIDI device attached or not ? Perhaps i'm the only one person doing that but... when i check new version, or check actual settings, initial settings, and so on... i do these preliminary checking not connected to my organ because i simply don't need to do that ! Of course, i'm not playing like this many times a day but.... u know, it is a little bit annoying to close two dialogs without real reason...

As i wrote, not really important just nice-to-have...

Issue #705, opened 2021-08-20
Jack Error messages on Raspberry PI


ajvdw:
I Switched the audio device to RT->Rt:Alsa (both enabled) and disabled Pa, Jack and Rt:Jack are also disabled in audio settings. It gives me the following messages when GO is running (audio works fine) on the Raspberry PI 8Gb with HifiBerry DACplus.

```
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiInJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel

```

AJ

Issue #703, opened 2021-08-18
Enchance Midi Settings


oleg68:
Now GrandOrgue automatically enables all available midi devices and then allows to disable some of them in Midi Settings.

GO keeps the names of disabled midi devices.

Sometimes midi devices change their names across system restart. For example, the name of alsa device names contain the midi port number that depends on the device initialisation order and can vary. In this case the old device with the new name will be automatically enabled that may be not desirable.

I'm going to make the following changes:
- [x] Introduce a subsystem filter for devices (alsa, jack, core etc)
- [x] Introduce a settings ``Enable new devices automatically``, by default enabled.
matched by regex
- [x] Introduce the setting not to check midi on startup #796 

Some other midi enchancements will be done in #885

ajvdw:
I'd like to enable/disable midi devices per RtMidi API (Jack, ALSA, etc?)

oleg68:
@ajvdw I've added it to the to-do list

oleg68:
@rousseldenis @larspalo 

Because this issue is not obvious, we need to discuss it. I'd ask you to post your thoughs and suggestions here.

larspalo:
The issue of (USB) MIDI devices changing names is not new. In rev 2174 Martin actually patched RtMidi because it's really a big part of the issue. With the updates to RtMidi those patches may long be gone now (I'm not certain if they were merged upstream or not). If the same effect can be achieved with regex comparison of devices it's all fine and dandy. But if possible, the list of devices should not name them so that any confusion will be created even if the port number changes but match the stored device to what the system presents, ideally in such a way that the user will never notice it. Otherwise it seems to be a good idea to enable a subsystem filter so that a lot of the "noise" caused by probing at startup can be eliminated once a working setup is achieved.

Issue #885, opened 2021-11-23
Midi keyboard identifications are not persistent 


pmwisse:
I've recently switched from the old Sourceforge version to this one and most is a tremendous improvement, for which I am very grateful. The names of the midi devices changed, but another problem went along with it, which is that the names are no longer persistent after restarts of Ubuntu. I use Ubuntu 21.10 and 3.4.2-1 with three UMX-61 keyboards. After I added a Bluetooth headphone to my Linux install, the names of the devices changed so I screwed up all the settings in my organs. I'm not sure this is a problem in GrandOrgue, but this was a problem that did not exist in the last versions of the Sourceforge Grandorgue.

oleg68:
Will be fixed in #703

pmwisse:
OK, thanks! I'll wait patiently! So happy to see active development on this precious piece of software!

oleg68:
I'm reopening this issues because the code modifications for #703 became very large.

Enchancement of midi device matching will be done in this issue

oleg68:
- [ ] Allow to match device names by regex
- [ ] Introduce the Matching button for changing regex for the device
- [ ] Midi events configured for the device should continue to work if the device changes it's name but is still existing

PR #897, opened 2021-12-05
Feature/midi enchancement

oleg68:
Resolves: #703
Resolves: #705
Resolves: #796 

This PR adds new capabilities in the MIDI settings:

1. The choice not to add new midi devices automatically.
2. The choice not to bring the Settings dialog if no active MIDI devices exist at startup.
3. The capability to disable some MIDI APIs.

The enchancements if matching MIDI device names have separated to another issue #885 and will be submited later.

The help page and the screenshot of the ``MIDI devices`` has also been updated.


oleg68:
midi_devices_tab

help/grandorgue.xml
```diff
@@ -4929,59 +4914,86 @@ messages to easily switch between samplesets.
...
+	    default. Therefore it is possible to use the same MIDI
+	    channel on multiple interfaces without configuring any
+	    channel offset.</simpara></note>
+	  </listitem>
+	</varlistentry>
+	<varlistentry>
+	  <term>MIDI output devices</term>
+	  <listitem>
+	    <para>
+Theis checklistshows all available MIDI output interfaces. Selecting
```
larspalo, line 4986: This checklist shows

src/grandorgue/settings/GOSettingsPorts.h
```diff
@@ -20,20 +20,53 @@ class GOSettingsPorts
...
+    wxString m_PortName;
+    wxString m_ApiName;
+
+    PortItemData(const wxString& portName, const wxString &apiName = wxEmptyString)
+    {
+      m_PortName = portName;
+      m_ApiName = apiName;
+    }
+
+    bool isFor(const wxString& portName, const wxString &apiName = wxEmptyString)
```
larspalo, line 35: isFor is a bit cryptic, I think. If it could it be specified more clearly what this "is for", it would be great.

larspalo (changes requested):
Great job, just some small things you might consider changing.

src/grandorgue/settings/GOSettingsPorts.h
oleg68: Renamed to isItemForPortApi

oleg68:
@larspalo next time you can use ``suggest change`` in the github interface.

larspalo (approved):

oleg68:
The [intermediate build](https://github.com/oleg68/GrandOrgue-official/actions/runs/1541786656) with these changes:

If I open the `File->Settings` dialog and then I press `Cancel` without having made any changes, GO stops listen any MIDI events.

After pressing to the `Panic` button MIDI starts working again.

Seems we need still to work more with this PR.

larspalo:
While the latest change fixes the immediate issue, we now have MIDI and sound working in the background when the settings dialog is open, while earlier it was stopped until the settings dialog was closed one way or another.

Now, if the user push the button to change audio device then sound is cut off and continues to be so until settings dialog is closed, even if cancel is selected.

Was there any specific reason why you choose to not stop audio and MIDI when the settings dialog was opened?

oleg68:
> While the latest change fixes the immediate issue, we now have MIDI and sound working in the background when the settings dialog is open, while earlier it was stopped until the settings dialog was closed one way or another.

The same behaviour was before this PR: MIDI and sound was reopened only after closing the setting dialog with `OK` but not with `Cancel`.

oleg68:
Closing MIDI it was not the single issue of this PR before the last change. The most important was the order of MIDI device may be changed that could cause wrong midi map even after `OK` and `Panic`.

Now everything seems ready. @larspalo do you agree with this PR? If it is difficult to revie the whole PR, I could extract changes in sound and moving the device name magic from `GOSoundPortFactory` to `GOPortFactory` to the separate PR.

larspalo (commented):
I don't mind that the sound and midi is not stopped until really necessary, so that's ok with me.

However, I see that the whitespace handling is somewhat different in some of the files. Like 2 blankspace indentations, brackets placements, and spaces around -> and such. This of course doesn't impact performance, but shouldn't we adhere to the existing project coding standards?

oleg68:
> However, I see that the whitespace handling is somewhat different in some of the files. Like 2 blankspace indentations, brackets placements, and spaces around -> and such. This of course doesn't impact performance, but shouldn't we adhere to the existing project coding standards?

Yes, I use 2-space indenting instead of 8-space (tab) in the totally new code, otherwise it does not fit the line length. But I don't mix different intendation in same functions.

Could you give me examples of wrong bracket placements or wrong spaces around `->`? I'm ready to fix them.

larspalo:
GOMidiPortFactory.cpp line 48 et al ( -> )
GOMidiRtPortFactory.cpp line 79 missing whitespace after for (same with 128). Indentations around line 92 are a bit strange (same with 141).
GOPortFactory.cpp line 18 has brackets in java style (like I personally use them...), but most of all it's inconsistent even within the same file.
In GOPortFactory.h you sometimes have brackets like on line 31 and sometimes like on line 34. It's similar in other files.

You'll find similar issues in other files too.

Regarding indentations we earlier did that only by tab in GO code. Personally I've used 2 blankspaces instead as it's more compact, but never in GO code. Now the style of formatting GO code is very diverse...

oleg68:
> GOMidiPortFactory.cpp line 48 et al ( -> ) 

Fixed

> GOMidiRtPortFactory.cpp line 79 missing whitespace after for (same with 128). 

The original code with `for(` was in GOMidiRtInPort.cpp, I've moved it here without adding a space. Now it is fixed.

> Indentations around line 92 are a bit strange (same with 141).

```
	    ports.push_back(
	      new GOMidiRtInPort(
		midi, api, deviceName,
		portFactory.ComposeDeviceName(PORT_NAME, apiName, deviceName)
	      )
	    );
```

The constructor call does not fit into one line. How do you suggest to reformat it?

> GOPortFactory.cpp line 18 has brackets in java style (like I personally use them...), but most of all it's inconsistent even within the same file. 

Fixed.

> In GOPortFactory.h you sometimes have brackets like on line 31 and sometimes like on line 34. It's similar in other files. You'll find similar issues in other files too.

When the statement is short, I put it at the same line. But when the statement is long and does not fit the line, I wrap it to the next line. I prefer not to put a more complex code in the `.h` file. Do you have a better idea of formatting?

> Regarding indentations we earlier did that only by tab in GO code. Personally I've used 2 blankspaces instead as it's more compact, but never in GO code. Now the style of formatting GO code is very diverse...

Not diverse. There are only two: 2 and 8 (tab).

Yes, it does not look right, but now the code is often hard to read because it does not fit the screen. I suggest to change the style to two-spaces. Now I do it only in the new code, but maybe we introduce an automatic clang formatting in the future?

larspalo:
As I said earlier, this is not anything essential, but a (reasonably) consistent style makes it easier to read the code.

Personally, if things tend to get too long for a single line I nowadays break it up with one parameter each line:
~~~
whatever someFunctionOrSimilar(
  int firstParameter,
  float secondParameter,
  and so_on,
  until all,
  have been,
  written out
)
{
  ...
}
~~~

But I tend to lean toward java style and would normally have placed the bracket on the same line as the end parenthesis.

I absolutely agree upon not putting anything in the header files that you won't expect to be possible to inline, thus only very short code should be there if at all any.

But, again, to me this is not a really essential topic even if I think it's much more enjoyable to read code that's clear, concise and consistent.

oleg68:
> Personally, if things tend to get too long for a single line I nowadays break it up with one parameter each line:

I do the same in function declarations but not always in function calls.

Should I change

```
	      new GOMidiRtInPort(
		midi, api, deviceName,
		portFactory.ComposeDeviceName(PORT_NAME, apiName, deviceName)
	      )
```

to

```
	      new GOMidiRtInPort(
		midi, 
                api, 
                deviceName,
		portFactory.ComposeDeviceName(PORT_NAME, apiName, deviceName)
	      )
```
?


> But, again, to me this is not a really essential topic even if I think it's much more enjoyable to read code that's clear, concise and consistent.

I'm trying to make the code like this.

Could you reapprove this PR if you do not have essential comments more?

larspalo:
> Should I change ... to ... ?

To me, that looks nicer.

oleg68:
> To me, that looks nicer.

OK. I've changed it.

What other changes should I make?

larspalo:
Maybe we should let this go for now. There are many other (perhaps much more important) things to focus on, but eventually we should agree upon, and work out (including documenting), a coding standard we intend to follow and over time adapt all project code to.

larspalo (approved):

rousseldenis (commented):
Seems great!

For code formatting, I use pre-commit and black (for python). But we could use something like https://github.com/pocc/pre-commit-hooks.

So, we don't care about format before committing (after repo configuration).

rousseldenis (approved):

PR #909, opened 2021-12-17
Midi matching Part 1

oleg68:
This is a first preparing PR for #885. GO behaviour should not be changed.

1. Because matching is some complex thing that must be implementing in both MidiIn and MidiOut, I made some unification: replaced the lists GOMidiInPort and GOMidiOutPort with the unified list of GOMidiPort. It allows to have a single code working with the lists. I use casting from GOPort to GOMidiInPort or GOMidiOutPort in a few places where it is necessary.
2. I added const iterators to `ptr_vector` and replaced loops on it with ``for (el: list)`` form in several places.

### DIFF
--- a/src/core/ptrvector.h
+++ b/src/core/ptrvector.h
@@ -10,7 +10,8 @@
 #include <vector>
 
 template<class T>
-class ptr_vector : protected std::vector<T*> {
+class ptr_vector : protected std::vector<T*>
+{
 private: 
 	/* Disable copying as it will break things */
 	ptr_vector(const ptr_vector&);
@@ -81,6 +82,12 @@ public:
 	{
 		std::vector<T*>::insert(std::vector<T*>::begin() + pos, ptr);
 	}
+
+	typename std::vector<T*>::const_iterator begin() const noexcept
+	{ return std::vector<T*>::begin(); }
+
+	typename std::vector<T*>::const_iterator end() const noexcept
+	{ return std::vector<T*>::end(); }
 };
 
 #endif
--- a/src/grandorgue/midi/GOMidi.cpp
+++ b/src/grandorgue/midi/GOMidi.cpp
@@ -6,10 +6,13 @@
 
 #include "GOMidi.h"
 
+#include "midi/GOMidiWXEvent.h"
+#include "ports/GOMidiInPort.h"
+#include "ports/GOMidiOutPort.h"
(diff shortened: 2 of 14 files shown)


larspalo (changes requested):
In the ranged for loops like
GOMidi.cpp line 49 (and all similar places), I suggest a blankspace before the : so that it would look like
`for (GOMidiPort* pPort : m_midi_in_devices)`

Then, if I want to split long if conditions into multiple lines, I prefer to have the logical operator last on the line and not starting it. Like for instance SettingsMidiDevices.cpp line 107 - 110 (and all similar places):
~~~
    if (
      portsConfig.IsEnabled(port->GetPortName(), port->GetApiName()) &&
      port->IsToUse()
    )
~~~
But that's of course only my preference...

oleg68:
> In the ranged for loops like GOMidi.cpp line 49 (and all similar places), I suggest a blankspace before the : so that it would look like `for (GOMidiPort* pPort : m_midi_in_devices)`

Fixed

> Then, if I want to split long if conditions into multiple lines, I prefer to have the logical operator last on the line and not starting it. Like for instance SettingsMidiDevices.cpp line 107 - 110 (and all similar places):
> 
> ```
>     if (
>       portsConfig.IsEnabled(port->GetPortName(), port->GetApiName()) &&
>       port->IsToUse()
>     )
> ```

It would be harder to understand the structure of the expression especially when the left side is not fully visible. In my case parentheses and operation signs are on the beginning of the lines, so the structure of a the complex expression is unrderstadable easier. So I woldn't change this.

larspalo:
> So I woldn't change this.

Sure, I just said what I prefer. To the compiler this doesn't matter at all.

Just out of curiosity, what kind of screen/resolution are you using that have issues with such short lines anyway?

larspalo (approved):

oleg68:
> Just out of curiosity, what kind of screen/resolution are you using that have issues with such short lines anyway?

My main desktop computer has a 2560x1440 screen, so I don't have problems on it even in the two-side mode (the header files are opened on the left side of screen and the code files are opened on the right side).

But sometimes I work with GrandOrgue source code on my small notebook with a 1366x768 screen. It is hard to work with long lines on it.

larspalo:
Ok, thanks.

oleg68:

rousseldenis (approved):

PR #912, opened 2021-12-18
Midi matching part 2

oleg68:
This is a second preparing PR for #885 The GrandOrgue behaviour should not change.

1. Each MIDI device started having three names
    1. Physical name - the same name as earlier. It may be not persistent.
    2. Logical name - the name is stored in the GrandOrgue settings. It may me some meaningful, ex. `Manual 1`
    3. Regex pattern - for matching with a (not persistent) physical name. It is also stored in the GrandOrgue settings
2. The source code has been rewritten to match devices with regex pattern. If the pattern is not specified, then GrandOrgue expects that the Logical Name is the same as the Physical Name
3. In the future GOMidiPort and it's descendents will set some default Logical Name and the Regex pattern (by removing a not persistent part of the name). But now GOMidiPort::GetDefaultLogicalName returns the same as the physical name and GOMidiPort::GetDefaultRegex returns an empty string (these methods have be introduced in the previous PR)
4. Earlier GOSettings kept midi device attributes in several scalar values, accessed by the names. But because the logical name may be changed, it is hard to maintain. So I rewrote GOSetttings in an object maner: to keep the config of each midi device as an object. The two classes `GOMidiDeviceConfig` and `GOMidiDeviceConfigList` have been introduced.
5. SettingsMidiDevice may edit GOMidiDeviceConfig lists and then may save or discard the changes. So the inner class `MidiDeviceListSettings` has been introduced for oprating with both input and output midi devices in the same manner.
6. In this PR it is stll not possible to change the Logical Name and the pattern. So now the logical names are always equal to the physical name until the next PR. So GrandOrgue should operate as earlier.

### DIFF
--- a/ide-projects/NetBeans12/nbproject/configurations.xml
+++ b/ide-projects/NetBeans12/nbproject/configurations.xml
@@ -339,6 +339,8 @@
             <in>MIDIEventSendDialog.cpp</in>
           </df>
           <df name="settings">
+            <in>GOMidiDeviceConfig.cpp</in>
+            <in>GOMidiDeviceConfigList.cpp</in>
             <in>GOPortFactory.cpp</in>
             <in>GOPortsConfig.cpp</in>
             <in>GOSettings.cpp</in>
@@ -737,9 +739,6 @@
           <in>GOrgueUtil.cpp</in>
           <in>GOrgueWave.cpp</in>
         </df>
-        <df name="grandorgue">
-          <in>GOSoundWorkItem.cpp</in>
-        </df>
       </df>
     </df>
     <logicalFolder name="ExternalFiles"
@@ -786,6 +785,15 @@
           <cleanCommand>${MAKE} -f Makefile clean</cleanCommand>
           <executablePath>../../build/current/bin/GrandOrgue</executablePath>
           <ccTool flags="5">
+            <incDir>
+              <pElem>../../build/current/src/core/go_defs.h</pElem>
+            </incDir>
+            <preprocessorList>
+              <Elem>GO_STD_MUTEX=1</Elem>
+              <Elem>WXUSINGDLL</Elem>
+              <Elem>_FILE_OFFSET_BITS=64</Elem>
+              <Elem>__WXGTK__</Elem>
+            </preprocessorList>
           </ccTool>
         </makeTool>
         <preBuild>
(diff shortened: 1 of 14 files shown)


larspalo (approved):
Everything seems to be working as far as I've tested it. However, I think the code formatting (again), especially the indentations, is all over the place, but it might just be the GitHub display that's messing up things... Please have a look at it in the files changed tab of this PR, especially places like in GOMidi.cpp line 63 to 69. It seems to be a mix of tab and spaces...

Somehow, the header files usually makes more sense on GitHub than the definition files, when it comes to formatting. But I don't think that should hold back this PR, it's just something I think we need to eventually address somehow. Maybe we should get back to tab only indentations and you set the display of them to whatever width you want in the IDE?

oleg68:
> Please have a look at it in the files changed tab of this PR, especially places like in GOMidi.cpp line 63 to 69. 

What is wrong in these lines? I put avery argument of `Append` in a separate line, as you like.

```
	    if (! pDevConf && isToAutoAdd)
	      pDevConf = midiIn.Append(
		pPort->GetDefaultLogicalName(),
		pPort->GetDefaultRegEx(),
		true,
		physicalName
	      );
```

> It seems to be a mix of tab and spaces...

Yes, 1 tab = 8 spaces. So instead of typing 10 spaces I type one tab and 2 spaces. 14 spaces = 1 tab + 6 spaces. The IDE calls this as ``optimal fill``. The text is displayed correctly in all standard text editors, including gedit and notepad.

> Maybe we should get back to tab only indentations and you set the display of them to whatever width you want in the IDE?

Lets talk about it in a special discussion topic, not here.

larspalo:
> What is wrong in these lines?

This is wrong:

WrongDisplayOfIndention

> The IDE calls this as optimal fill.

I call it senseless and makes it more difficult to read than it has to. And no, one shouldn't mix tab and spaces. One stick to either or. Tabs can be shown at any amount of blankspaces the user wants to in most editors.

oleg68:
> > What is wrong in these lines?
> 
> This is wrong:
> 
> WrongDisplayOfIndention
> 
Where have you got this view? Github shows it correctly:
github

oleg68:
@larspalo Yes, it is. [There is a user prefernce setting in github](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-user-account-settings/managing-your-tab-size-rendering-preference)

I have the default value `8` there. Seems you have `4`.

Although several text editors allow to change it, `8` is the only worldwide standard.

larspalo:
@oleg68 Something else must be causing it since I still have the 8 space default in my account settings appearance, and while browsing code in the repository things look ok. While 8 spaces default (that comes from the old typewriter times) might be an universal standard, yes I do indeed prefer 4 spaces and I use that in most editors!

If we stick to only tabs this would never be an issue. Nor would it be an issue if we opted to only use 2 blankspaces... Mixing them causes all kinds of possible issues. But since tabs can indeed be displayed at any width nowadays, it would be easier if we just selected that road.

oleg68:
@larspalo I replaced the second tab with spaces only in the lines 65-69 in `GOMidi.cpp`. If I replace the first tab also, the code will be indented diferently relatives the lines I haven't change.

My opinion we should go to the automatic formatting and intendation.

larspalo:
> My opinion we should go to the automatic formatting and intendation.

Absolutely, if we can get it to do the right thing! ;-)

oleg68:
Yes, but not in this PR

rousseldenis (approved):

oleg68:
> > My opinion we should go to the automatic formatting and intendation.
> 
> Absolutely, if we can get it to do the right thing! ;-)

I created the issue #918 and the discussion topic #920

PR #913, opened 2021-12-23
Add support for building against system-wide RtAudio and RtMidi libraries

svillemot:
Note that this patch is part of my ongoing work to create an official Debian package for GrandOrgue (official from Debian’s perspective, of course).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 - Added a setting option whether to check for existence of active midi devices on startup https://github.com/GrandOrgue/grandorgue/issues/796
 - Added capability of switching auto enabling of new midi devices on/off https://github.com/GrandOrgue/grandorgue/issues/703
 - Added capability of enabling and disabling particular MIDI APIs https://github.com/GrandOrgue/grandorgue/issues/703
+- Added support for building against system-wide RtAudio and RtMidi libraries
 # 3.4.4 (2021-11-27)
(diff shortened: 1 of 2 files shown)


oleg68:
@svillemot, please add an entry in CHANGELOG.md.

svillemot:
> @svillemot, please add an entry in CHANGELOG.md.

Done.

oleg68 (approved):

oleg68:
@larspalo could you look at this PR?

larspalo:
@oleg68 We don't have any current changes/patches against the RtAudio/RtMidi source right? If so, we could even change the order and check for a system wide version first and then if not found for a supplied version, right?

oleg68:
> @oleg68 We don't have any current changes/patches against the RtAudio/RtMidi source right? 

Yes, we don't.

> If so, we could even change the order and check for a system wide version first and then if not found for a supplied version, right?

I haven't catch your idea. What order are you talking about? No there is a variable `USE_INTERNAL_RTAUDIO` that switches between internal and external rt in the same way as `USE_INTERNAL_PORTAUDIO`.

And Merry Christmass!

larspalo:
Merry Christmas!

My thoughts were that in this PR the in source RtAudio and RtMidi is default (ON). Since we're not actively maintaining any changes in the libraries anymore we could as well have this as OFF by default and recommend using distribution supplied versions instead.

larspalo (approved):

PR #927, opened 2021-12-31
Midi matching part 3

oleg68:
This is the last preparing PR for #885

1. Fixed a bug in Midi Device list iterator (GOMidiDeviceConfigList.h)
2. Added the typedef GOMidiDeviceConfig::List (GOMidiDeviceConfig.h)
3. Fixed not saving midi device regex (GOMidiDeviceConfigList.cpp)
4. Added a capability of removing `MidiDeviceConfig` elements from MidiDeviceConfigList with duplicate logical names
5. The inner class SettingsMidiDevices::MidiDeviceListSettings converted to the sepate class GOSettingsMidiDeviceList

No new features haave been added to GO.

After merging this PR I'll submit the https://github.com/oleg68/GrandOrgue-official/tree/feature/midi-matching with added a capability of redefining midi names and regex (the work is still in progress now)

### DIFF
--- a/src/core/ptrvector.h
+++ b/src/core/ptrvector.h
@@ -83,6 +83,13 @@ public:
 		std::vector<T*>::insert(std::vector<T*>::begin() + pos, ptr);
 	}
 
+	void erase(unsigned pos)
+	{
+		if (at(pos))
+		  delete at(pos);
+		std::vector<T*>::erase(std::vector<T*>::begin() + pos);
+	}
+
 	typename std::vector<T*>::const_iterator begin() const noexcept
 	{ return std::vector<T*>::begin(); }
 
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -147,6 +147,7 @@ GOPanelView.cpp
 settings/GOMidiDeviceConfig.cpp
 settings/GOMidiDeviceConfigList.cpp
 settings/GOPortFactory.cpp
+settings/GOSettingsMidiDeviceList.cpp
 settings/GOSettingsPorts.cpp
 settings/SettingsArchives.cpp
 settings/SettingsAudioGroup.cpp
--- a/src/grandorgue/settings/GOMidiDeviceConfig.h
+++ b/src/grandorgue/settings/GOMidiDeviceConfig.h
@@ -18,7 +18,7 @@ private:
   wxRegEx* p_CompiledRegEx = NULL;
 
 public:
-  typedef std::vector<GOMidiDeviceConfig*> List;
+  typedef std::vector<GOMidiDeviceConfig*> RefVector;
 
   wxString m_LogicalName;
   wxString m_RegEx;
--- a/src/grandorgue/settings/GOMidiDeviceConfig.h
+++ b/src/grandorgue/settings/GOMidiDeviceConfig.h
@@ -7,6 +7,8 @@
 #ifndef GOMIDIDEVICECONFIG_H
 #define GOMIDIDEVICECONFIG_H
 
+#include <vector>
+
 #include <wx/regex.h>
 #include <wx/string.h>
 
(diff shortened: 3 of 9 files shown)


larspalo:
Just out of curiosity and since I've recently had loads of issues with using vectors of pointers... Are the elements pointed to also stored in a vector (or array), and is it possible that they too can be deleted in different positions, not just popped at the end?

oleg68:
> Just out of curiosity and since I've recently had loads of issues with using vectors of pointers... Are the elements pointed to also stored in a vector (or array), and is it possible that they too can be deleted in different positions, not just popped at the end?

As I understand, ptr_vector is a container of its pointed elements. So when an element is removed from the vector or the whale vector is destroyed, the elements also must be deleted. It is assymetric with creation of elements: the elements are created outside the ptr_vector and then their ownership is delegated to it.

The standard c++ alternative of ptr_vector is std::vector<std::unique_ptr>. But such replacing would require modifying all constructors.

oleg68:
@larspalo I sorry for adding more changes after you have started the review. No I've finished composing this PR, and I'll push new commits only in response to CR comments,

larspalo:
> As I understand, ptr_vector is a container of its pointed elements. So when an element is removed from the vector or the whale vector is destroyed, the elements also must be deleted. It is assymetric with creation of elements: the elements are created outside the ptr_vector and then their ownership is delegated to it.

Yes, that should be just fine in this case. If indeed the pointer vector also owns the elements (and they are created on the stack/free store), then there won't be a problem with this at all, I think.

When working on the odf creation tool I recently ran into an issue involving objects owned and stored in one class (on my first attempt in a vector) and then referenced to by pointers (first tried also storing them in vectors) by other classes... Since that program doesn't have any performance constraints, I solved that by using std::list instead, but for GO that wouldn't be appropriate and unique_ptr would be better.

larspalo (commented):
This is something that will need to be tested over many computer restarts to be able to evaluate it properly.

Otherwise my only consideration is the

`typedef std::vector<GOMidiDeviceConfig*> List;`

which is named in a very generic way, when it's really quite specific. I'd consider renaming it to something more descriptive.

oleg68:
> When working on the odf creation tool I recently ran into an issue involving objects owned and stored in one class (on my first attempt in a vector) and then referenced to by pointers (first tried also storing them in vectors) by other classes...

Pointers to elements stored in std::vector may be changed, for example, when the vecror is resizing. That is why I use ptr_vector in `GOMidiDeviceConfigList` instead of std::vector.

> but for GO that wouldn't be appropriate and unique_ptr would be better.

std::vector<std::unique_ptr> has the same semantic as ptr_vector in GO. Replacing ptr_vector with std::vector<std::unique_ptr> in GO would require modifying a huge lot of files. I usually prefer making such extremally large but simple changes in a sope of a separate isue and not combining them with semantic changes.

oleg68:
> This is something that will need to be tested over many computer restarts to be able to evaluate it properly.

Now logical names are always equal to physical names so if the device name is changed over restart than the device will not be matched. In https://github.com/oleg68/GrandOrgue-official/tree/feature/midi-matching will be able to change the logical names and set up regexes manually.

> `typedef std::vector<GOMidiDeviceConfig*> List;`
> 
> which is named in a very generic way, when it's really quite specific. I'd consider renaming it to something more descriptive.

This typedef is inner in GOMidiDeviceConfig, so it's name is GOMidiDeviceConfig::List, so `A list of GOMidiDeviceConfig`. I don't find useful to double the name of the main class in the inner type like GOMidiDeviceConfig::GOMidiDeviceConfigList.

But your are still right. There is another class GOMidiDeviceConfigList that's name may be confused with GOMidiDeviceConfig::List. So I've renamed GOMidiDeviceConfig::List to GOMidiDeviceConfig::RefVector.

larspalo (approved):

oleg68:

rousseldenis (approved):

PR #930, opened 2022-01-01
Midi matching p4

oleg68:
This PR adds the capability to redefine MIDI device logical names and the matching regex to users. The new buttons `Matching` now available in the `MIDI devices` page.


larspalo:
Earlier I had three identical keystations, each connected by USB directly to the computer, their names only differ by their port numbers (which would be random at reboots) - they all send on same channel (at each startup they would anyway reset to channel 1). With the earlier RtMidi fix that Martin made they would be persistant over reboots. How would this regexp be setup to be able to do this?

larspalo (changes requested):
help/grandorgue.xml

around line 5020 I suggest:
If the user alters the logical name, a regex pattern must be provided
so that GrandOrgue can find this device by its physical name.

around line 5024 seems to be incomplete

GOSettingsMidiMatchDialog.cpp
line 26, add blankspace before :

at line 37 and similar (also in SettingsMidiDevices.cpp), I've never seen sizers variable names specified as const pointers before, but I guess it really doesn't matter that much...

at line 93 (line 133, 174), I don't like variables with names such as rc that makes you guess their meaning and purpose. Already names like you have on line 125 is borderline in my opinion where errMsg can still be fairly easily decoded as errorMessage, but it's un-necessarily shortened, I think.

oleg68:
> Earlier I had three identical keystations, each connected by USB directly to the computer, their names only differ by their port numbers (which would be random at reboots) - they all send on same channel (at each startup they would anyway reset to channel 1). With the earlier RtMidi fix that Martin made they would be persistant over reboots. How would this regexp be setup to be able to do this?

You can make three logical names `Manual 1`, `Manual 2` and `Manual 3` with the same regex `alsa: USB Keystation 61es`. In this case the device with the lowest port will be matched with `Manual 1`, the next - with `Manual 2` and the next - with `Manual 3`

larspalo:
> You can make three logical names `Manual 1`, `Manual 2` and `Manual 3` with the same regex `alsa: USB Keystation 61es`. In this case the device with the lowest port will be matched with `Manual 1`, the next - with `Manual 2` and the next - with `Manual 3`

Yes, but this doesn't seem to mean that the one matched to manual 1 will necessarily be the lowest one in my keyboard stack... I'll need to test this!

oleg68:
> Yes, but this doesn't seem to mean that the one matched to manual 1 will necessarily be the lowest one in my keyboard stack... I'll need to test this!

Usually all internal usb controllers enumerate all devices in the order of usb port numbers. But external usb hubs may enumerate devices in parallel, so the order may vary in this case. So the regex solution will help when your keyboards are connected directly and may not help with usb hubs.

The right solution is to implement capability of specifying the MIDI client name for snd_usb_audio with udev rules.

Because my keyboards are connected with a usb hob, I use a shell script that connects keyboard based on their usb port numbers.

larspalo:
Ok, earlier, (before Martin fixed it in GO) I did it like this http://familjenpalo.se/vpo/soundandmidi/

But quite a few years back, I got fed up and invested in a https://www.esi-audio.com/products/m4uxt/ that takes care of all this kind of issues.

I just wonder about the actual effectiveness of this PR feature for any real usage, but I'll have to test it...

oleg68:
> help/grandorgue.xml
> 
> around line 5020 I suggest: If the user alters the logical name, a regex pattern must be provided so that GrandOrgue can find this device by its physical name.
Fixed
> around line 5024 seems to be incomplete
Fixed
> GOSettingsMidiMatchDialog.cpp line 26, add blankspace before :
Fixed
> at line 37 and similar (also in SettingsMidiDevices.cpp), I've never seen sizers variable names specified as const pointers before, but I guess it really doesn't matter that much...
I prefer to use `const` everywhere as possible. It helps find more errors at compile time.
> at line 93 (line 133, 174), I don't like variables with names such as rc that makes you guess their meaning and purpose.
Renamed `rc` to `isValid`.
> Already names like you have on line 125 is borderline in my opinion where errMsg can still be fairly easily decoded as errorMessage, but it's un-necessarily shortened, I think.
Renaming `errMsg` to `errorMessage` would increase line length. Sometimes it would irequire to split lines
(ex. lines 109, 115). So I'd continue to use this abbreviation, because it is clearly understandable and allows to compactify the code.

oleg68:
> I just wonder about the actual effectiveness of this PR feature for any real usage, but I'll have to test it...

You can take [the intermediate build](https://github.com/oleg68/GrandOrgue-official/runs/4684348565) for testing. I'm going to add some heuristics to the GO code in the next PR that GO would assign the logical name and the regex pattern automatically.

larspalo (approved):

rousseldenis (approved):

PR #933, opened 2022-01-03
Added default regex patterns for matching MIDI devices

oleg68:
Resolves: #885

This is the last PR on midi matching.

After this PR GrandOrgue tries to make a non-volatile logical name and a regex pattern for each midi device, so no user intervention is required.

The only thing rests to fix is the old devices that GrandOrgue remembered when it matched only with the physical name. Earlier I wanted to add a button and a special dialog box that would allow to see the absent devices and to remove them. But after creating GO config from scratch it is not more an issue, so now I'm not going to create such UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Added default regexe patterns for matching MIDI devices with alsa, jack and winmm https://github.com/GrandOrgue/grandorgue/issues/885
+- Added default regex patterns for matching MIDI devices with alsa, jack and winmm https://github.com/GrandOrgue/grandorgue/issues/885
 - Added capability of matching MIDI devices with regex expressions https://github.com/GrandOrgue/grandorgue/issues/885
 - Added a setting option whether to check for existence of active midi devices on startup https://github.com/GrandOrgue/grandorgue/issues/796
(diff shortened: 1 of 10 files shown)


larspalo (approved):
I only found a few non important things.

In GOMidiDeviceConfigList.cpp on line 105, in the comment, there's a typo for Replace.

The code formatting is sometimes inconsistent. You have sometimes done like in GOMidiRtOutPort.cpp line 36 to 38, and sometimes like for instance GOMidiRtPortFactory.cpp line 282-283. Again this is something we should have fixed with the automatic formatting, when we get that working.

For now, issues like this can be considered relatively un-important, so I'm approving without requesting any changes.

oleg68:
> In GOMidiDeviceConfigList.cpp on line 105, in the comment, there's a typo for Replace.

Fixed
> 
> The code formatting is sometimes inconsistent. You have sometimes done like in GOMidiRtOutPort.cpp line 36 to 38, and sometimes like for instance GOMidiRtPortFactory.cpp line 282-283.

I made GOMidiRtPortFactory.cpp line 282-283 like GOMidiRtOutPort.cpp line 36 to 38

> For now, issues like this can be considered relatively un-important, so I'm approving without requesting any changes.

OK, thank you. I'm waiting for @rousseldenis approval.

oleg68:

rousseldenis (approved):

Issue #918, opened 2021-12-27
Automatic formatting the code


oleg68:
Provide a way for automatic formatting the GrandOrgue source code.

There are several variants:

- Some tool (ex. `git clang-format`) to run manually before commit
- A test in the Github workflow that checks that all code is formatted correctly
- Some Github push hook
- anything else

The formatting rules are discussed at #920

PR #946, opened 2022-01-06
[IMP] pre-commit: Use formatting

rousseldenis:


rousseldenis:
@oleg68 @larspalo This is a first implementation of pre-commit runs and in the logs you can see the proposal diffs with the actual configuration.

Related to https://github.com/GrandOrgue/grandorgue/discussions/920
And #918

oleg68:
@rousseldenis how does it work? Where can I see its result?

rousseldenis:
> @rousseldenis how does it work? Where can I see its result?

You can go in pre-commit build details and unfold the run that failed.

It's a first quick test I've done.

But you can reproduce locally

oleg68:
@rousseldenis Let's merge at first #947 before this PR, otherwise we'll receive a lot of commits.

rousseldenis:
> @rousseldenis Let's merge at first #947 before this PR, otherwise we'll receive a lot of commits.

Yes, of course. I'm currently trying to fix errors for each hook. But maybe, at the end I'll need your help as it linked to specific things in cpp programming (I'm fixing the obvious things).

rousseldenis:
So, this is ready with only clang formatting with pre-commit.

If we merge this, you need to do 'pre-commit install' before further git committing.
And the magic will occurs !

src/core/GOBitmap.h
```diff
@@ -11,30 +12,31 @@
...
-private:
-	wxImage* m_img;
-	wxBitmap m_bmp;
-	double m_Scale;
-	int m_ResultWidth;
-	int m_ResultHeight;
-	unsigned m_ResultXOffset;
-	unsigned m_ResultYOffset;
+class GOBitmap {
+ private:
```
oleg68, line 16: Something went wrong.

When I run clang-format, I receive

```
class GOBitmap {
private:
  wxImage *m_img;
```

Could you paste the result of ``clang-format --dump-config | grep AccessModifierOffset`` here?

rousseldenis: @oleg68 You are on my branch ?

rousseldenis: You have the wrong format locally ?
Because the pre-commit build here on github passes (so, the clang-format output is the same as mine).

rousseldenis: image

src/core/GOBitmap.h
```diff
@@ -11,30 +12,31 @@
...
+  wxBitmap m_bmp;
+  double m_Scale;
+  int m_ResultWidth;
+  int m_ResultHeight;
+  unsigned m_ResultXOffset;
+  unsigned m_ResultYOffset;
 
-	void ScaleBMP(wxImage& img, double scale, const wxRect& rect, GOBitmap* background);
+  void ScaleBMP(wxImage& img, double scale, const wxRect& rect,
+                GOBitmap* background);
```
oleg68, line 26: Here I receive
```
  void ScaleBMP(
    wxImage &img, double scale, const wxRect &rect, GOBitmap *background);
```

Seems the options 

```
AlignAfterOpenBracket: AlwaysBreak
AlignOperands: 'false'
BinPackArguments: 'false'
BinPackParameters: 'false'
```

aren't not work with your clang. Or your clang-format ignores the `.clang-format` file

rousseldenis: Oh yes I see.

Ok, I will try to force the config file

rousseldenis: @oleg68 Ok, I got it. The style attribute should point file to take that one into account.

rousseldenis: Is this ok for you ?

But reading the clang-documentation, it recommends to do a clang-format dump-config to generate a complete setting file (A lot of config entries instead of this for the moment).

I can try it.

rousseldenis: > I can try it.

Done in last commit. It seems good. 
I've changed the repo for clang as the last version of clang-format is on pre-commit one.

src/core/GOMemoryPool.cpp
```diff
@@ -28,39 +28,33 @@
...
-      m_PoolLimit(0),
-      m_PageSize(4096),
-      m_CacheSize(0),
-      m_MallocSize(0),
-      m_MemoryLimit(0),
-      m_AllocError(0),
-      m_TouchPos(0),
+    : m_PoolStart(0), m_PoolPtr(0), m_PoolEnd(0), m_CacheStart(0),
+      m_PoolSize(0), m_PoolLimit(0), m_PageSize(4096), m_CacheSize(0),
+      m_MallocSize(0), m_MemoryLimit(0), m_AllocError(0), m_TouchPos(0),
```
oleg68: It is not desired.

Please add

```
ConstructorInitializerIndentWidth: 2
ConstructorInitializerAllOnOneLineOrOnePerLine: true
```

to the `.clang-format`

rousseldenis: > ConstructorInitializerAllOnOneLineOrOnePerLine: true

That one is already in place

rousseldenis: > ConstructorInitializerIndentWidth: 2

Done

oleg68: > > ConstructorInitializerAllOnOneLineOrOnePerLine: true
> 
> That one is already in place

No, it isn't. It should place every `m_PoolSize(0),` in a separate line

rousseldenis: It did it:

image


You need to go to files in Github.

rousseldenis: image

.clang-format
```diff
@@ -0,0 +1,212 @@
+---
+Language:        Cpp
```
oleg68, line 2: It is hard to read. Lets keep here only the keys wit not default values.

rousseldenis: Yes we can.

rousseldenis: Done

oleg68: Thank you, but you forgot to add 

```
ConstructorInitializerIndentWidth: 2
ConstructorInitializerAllOnOneLineOrOnePerLine: true
```

rousseldenis: Yes, just saw :sweat: Done

src/grandorgue/config/GOMidiDeviceConfigList.cpp
```diff
@@ -1,113 +1,92 @@
...
-
 #include "go_limits.h"
 
-GOMidiDeviceConfig* GOMidiDeviceConfigList::FindByLogicalName(
-  const wxString& logicalName
-) const
-{
-  GOMidiDeviceConfig* res = NULL;
+GOMidiDeviceConfig *
+GOMidiDeviceConfigList::FindByLogicalName(const wxString &logicalName) const {
```
oleg68: Please add ``PenaltyReturnTypeOnItsOwnLine: 1000`` for avoiding this kind of break

oleg68 (changes requested):

src/grandorgue/config/GOMidiDeviceConfigList.cpp
rousseldenis: Done

oleg68 (approved):

oleg68:
@larspalo please, take a look at the new formatting of the code.

There are two things may be to change
1. Lack of line breaks before `)`. It waits for the clang 14 becomes available
2. Earlier GrandOrgue used the style ``Type* var`` and ``Type& var``. The new style is ``Type *var`` and ``Type &var``. I'd chose the new style because it is more standard for C/C++. What do you think?

larspalo:
@oleg68 It's looking great, I think. I've got no problem with needing to wait for clang 14 to get things perfect, it will be better than before anyway, in my opinion. I completely agree on using Type *var as standard.

larspalo (approved):

oleg68:
@rousseldenis How to rerun formatting of the whole project?

rousseldenis:
> @rousseldenis How to rerun formatting of the whole project?

Locally, you can always do 'pre-commit run -a' to force checking all the current directory.

PR #947, opened 2022-01-06
Feature/reorganise source code of settings

oleg68:
According to to discussed in https://github.com/GrandOrgue/grandorgue/discussions/793#discussioncomment-1900125

this PR does the following:

1. Rename GOSettings to GOConfig
2. Split `src/grandorgue/settings` directory to `src/grandorgue/config` with the classes storing the program configuration and to `src/grandorgue/settings-gui` with the classes implementing GUI of the Settings dialog
3. Add the `GO` preffix to the `Settings*` classes and files


larspalo:
I basically agree overall. Though, the usage of settings versus config is not "crystal clear" in the GO code. It never was... But now with this PR we'll have for instance in GOApp.cpp on line 109:

`m_Settings = new GOConfig(m_InstanceName);`

What we'll need in the long run is to clearly differentiate in the code (variable names) **what** settings/configs we're dealing with right there and then. And yes, it **will** mean longer variable names, that's inevitable.

oleg68:
> I basically agree overall. Though, the usage of settings versus config is not "crystal clear" in the GO code. It never was... But now with this PR we'll have for instance in GOApp.cpp on line 109:
> 
> `m_Settings = new GOConfig(m_InstanceName);`
> 
> What we'll need in the long run is to clearly differentiate in the code (variable names) **what** settings/configs we're dealing with right there and then. And yes, it **will** mean longer variable names, that's inevitable.

`m_Settings` is the only variable name used in the GrandOrgue code. So I'd rename `m_Settings` to `m_config` in all files.

If you agree, I'll post that commit.

larspalo:
@oleg68 Sure, I don't mind that change.

oleg68:
@larspalo I've submitted renaming m_Settings and m_settings to m_config.

There are still variables `m_SettingsPath` and `m_SettingsPageNum` that remain untouched,

rousseldenis (approved):

larspalo (approved):

Issue #966, opened 2022-01-16
GrandOrgue crashes under OSx when the midi event dialog is open


oleg68:
It's a know issue, I've seen it before but I can't find it. It has something to do with an automatically released wxDialog. I don't know where to find the crash report but I tried to debug and this is what I get:

```
`GrandOrgue(785,0x10043bdc0)` malloc: *** error for object 0x7ffeefbfbbb8: pointer being freed was not allocated
GrandOrgue(785,0x10043bdc0) malloc: *** set a breakpoint in malloc_error_break to debug
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff72e4633a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00000001015719bc libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff72dcd808 libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff72ec750b libsystem_malloc.dylib`malloc_vreport + 548
    frame #4: 0x00007fff72eca40f libsystem_malloc.dylib`malloc_report + 151
    frame #5: 0x000000010014397c GrandOrgue`MIDIEventDialog::~MIDIEventDialog() + 44
    frame #6: 0x0000000100f9d052 libwx_baseu-3.1.dylib`wxAppConsoleBase::DeletePendingObjects() + 74
    frame #7: 0x0000000100f9d50c libwx_baseu-3.1.dylib`wxAppConsoleBase::ProcessIdle() + 82
    frame #8: 0x0000000100a8649e libwx_osx_cocoau_core-3.1.dylib`wxAppBase::ProcessIdle() + 22
    frame #9: 0x00000001009f18e8 libwx_osx_cocoau_core-3.1.dylib`wxApp::ProcessIdle() + 26
    frame #10: 0x00000001010624ff libwx_baseu-3.1.dylib`wxCFEventLoop::CommonModeObserverCallBack(__CFRunLoopObserver*, int) + 75
    frame #11: 0x00007fff38eab7bc CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 23
    frame #12: 0x00007fff38eab6ec CoreFoundation`__CFRunLoopDoObservers + 457
[...]

PR #971, opened 2022-01-17
Fixed crash under OSx when closing a MIDI event dialog

oleg68:
Resolves: #966


larspalo (changes requested):
In MIDIEventDialog.cpp there's two different stylings of similar if / else statements. Compare lines 80-83 and 87-90. I know it has no practical importance, but maybe we should have a single styling for the single line if/else statements.

oleg68:
> In MIDIEventDialog.cpp there's two different stylings of similar if / else statements. Compare lines 80-83 and 87-90. I know it has no practical importance, but maybe we should have a single styling for the single line if/else statements.

Fixed

larspalo (approved):

rousseldenis (approved):

PR #984, opened 2022-01-22
Fixed processing of debounce time for Bx Controller events

### LINKED ISSUE #967: GrandOrgue ignores Debounce Time settings for the "Next memory" midi event
If to assign pressing on an expression pedal to the `Next memory` sequencer action, sometimes it switches more than one step.

The reason that the pedal may send more than one ``Bx Controller 64`` MIDI event in a short time.

The ``Debounce time`` setting in the event dialog should prevent multiple switching combinations, but actually does not.
NextMemoryEvent

oleg68:
Resolves #967

Seems Mardin did not support debounce for ``Bx Controller`` when he made 7d4de54cc57a709b7cf7135ceb9e4c1710ac9665


oleg68:
@larspalo  @rousseldenis could you allrove this?

rousseldenis (approved):

larspalo (approved):

PR #987, opened 2022-01-25
Update de.po

Mikelectric:
The de.po file is much too short. It only has 2133 lines. So I can't translate the tabs that have changed. I've only changed the editor's name and hope that after the pull request the problem will be solved.


po/de.po
```diff
@@ -7,7 +7,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: ourorgan-developers@lists.sourceforge.net\n"
 "POT-Creation-Date: 2013-02-18 09:03+0100\n"
 "PO-Revision-Date: 2013-02-16 13:30+0100\n"
-"Last-Translator: Martin Koegler\n"
+"Last-Translator: Michael P. Schmidt\n"
```
oleg68, line 10: It is the gratest change!

What's about adding new pairs?

oleg68:
> The de.po file is much too short. It only has 2133 lines. So I can't translate the tabs that have changed. I've only changed the editor's name and hope that after the pull request the problem will be solved.

You've got an old `po.de`. Please take one from https://github.com/oleg68/GrandOrgue-official/blob/bugfix/translation-german/po/de.po. All messages to translate are marked with `***`. You can easy find all of them.

Mikelectric:
Ok, thanks, the translation is almost ready now. I just have to test. Then
I make a new pull request.

Am Mi., 26. Jan. 2022 um 08:10 Uhr schrieb Oleg Samarin <
***@***.***>:

oleg68:
The new PR #988

Issue #985, opened 2022-01-24
"Copy current receive event" in the Midi-Settings dialog doesn't work.


willeke1234:
"Copy current receive event" in the Midi-Settings dialog doesn't work. The send event is reset to (none). I can't find this feature in the Help.

Steps to reproduce:
1. Open de Midi-Settings dialog
2. Select an event
3. Switch to Send
4. Click "Copy current receive event"

The event is not copied. I've been debugging and the event isn't copied because the input devide can't be found in the output devices. The input device is "Any device".

5. Switch to Receive and select a Device
6. Switch to Send and click "Copy current receive event"

The event is not copied because the device is not in the output devices list. If the device is in the output devices list then the event is not copied because the output device isn't connected to the input device. Apparently I have to connect the input and output device with the button "MIDI output device..." in Settings. But now GrandOrgue crashes because Cocoa can't set the Channel popup button to 4294967294 (0xFFFFFFFE). The channel of the receive event is "Any channel".

7. Switch to Receive and select a Channel
8. Switch to Send and click "Copy current receive event"

After setting the channel to a number, "Copy current receive event" does work. But only if the event is sendable, otherwise the event is not copied and GrandOrgue doesn't tell me.

Summary:
Bug: "Any device" isn't copied.
Bug: "Any channel" crashes.
Bug: The "Copy current receive event" and "MIDI output device..." buttons are missing in the Help.
Feature request: If the receive event can't be copied then tell the user why.

PR #997, opened 2022-01-29
Prepared for fixing MidiSendDialog

oleg68:
It is a preparing commit for #985. It contains mostly renaming according to the GO conventions.

GOMidiRecvDialog and GOMidiSendDialog now have m_MidiIn, m_MidiOut and m_MidiMap references instead of m_config

PR #1008, opened 2022-02-03
Fixed "Copy current receive event"

oleg68:
Resolves: #985

This PR changes the behaviour of the "Copy current receive event" button in the Send page of the Midi Event Editor.

The changes:
- Now the event is copied even the input device is either not selected or does not have an output device specified. But in these cases the Device fiels is set to ``Any Device``
- Does not crash if the source input event has ``Any Channel``. In this case the channel becomes `1`
- OK is possible only all sending events have correct `Device` values. ``Any device`` is not valid and the user should select a device manually
- Added information about this button to the help

Also there are small changes in the layout of the Midi Event Editor.

This PR is based on #997, so most of changed files will go after merging #997


oleg68:
@larspalo @rousseldenis could you look at this PR?

src/grandorgue/midi/MIDIEventDialog.cpp
```diff
@@ -73,14 +74,44 @@ void MIDIEventDialog::RegisterMIDIListener(GOMidi *midi) {
...
+  wxWindow *notValidTab = NULL;
+  wxString errMsg;
+
+  if (!notValidTab && m_sendPage && !m_sendPage->Validate(errMsg))
+    notValidTab = m_sendPage;
+  if (
+    !notValidTab && m_sendDivisionPage && !m_sendDivisionPage->Validate(errMsg))
+    notValidTab = m_sendPage;
+
+  // Distplay the tab that has problems
```
rousseldenis, line 87: Display

rousseldenis (approved):
Code review. Nitpicking

src/grandorgue/midi/MIDIEventDialog.cpp
oleg68: Fixed

oleg68:
@larspalo could you approve this?

larspalo (approved):

Issue #1002, opened 2022-01-30
Wrong navigation in the Settings Dialog


oleg68:
If there is no midi devices available then GrandOrgue opens the Settings Dialog and displays the message ``No active MIDI devices``

But when the user clicks OK, the `Reverb` page becomes active instead of the ``Midi Devices``

Issue #1003, opened 2022-01-30
Help from the Settings Dialog doesn' show the correct topic


oleg68:
When to press Help from the Settings dialog then the opening help page depends on the current language.

If the currnt language is English then the help topic corresponds to the current active tab in the dialog
If the current language is not English then the help is opened always at the root.

The reason is the GrandOrgue tries to use the tab name as the index in the help. But there is no localised index in the help. So if the tab names are translated, the help navigation becomes wrong

PR #1012, opened 2022-02-05
Fixed automatic opening the Reverb settings page when there were no active midi devices

oleg68:
Resolves: #1002

This is a temporary bugfix.

Now the tab order of the settings dialog must be in two places: in the GOSettingsDialog::PageSelector and in GOSettingsDialog::GOSettingsDialog. Making a change in one of this place without a related change in another place introduces a bug like this.

I'm going to rewrite the dialog code for #1003 that there would be only one place with the tab order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
+- Fixed automatic opening the Reverb settings page when there were no active midi devices https://github.com/GrandOrgue/grandorgue/issues/1002
 - Fixed matching midi events with midi devices https://github.com/GrandOrgue/grandorgue/issues/1000
 # 3.6.1 (2022-01-29)
-- Updating French, German, Polish, Spanish and Swedish translations of the Settings Dialog https://github.com/GrandOrgue/grandorgue/discussions/934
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
+- Fixed automatic opening the Reverb settings page when there were no active midi devices https://github.com/GrandOrgue/grandorgue/issues/1002
 # 3.6.1 (2022-01-29)
-- Updating French, German, Polish, Spanish and Swedish translations of the Settings Dialog https://github.com/GrandOrgue/grandorgue/discussions/934
+- Updated French, German, Polish, Spanish and Swedish translations of the Settings Dialog https://github.com/GrandOrgue/grandorgue/discussions/934
```
rousseldenis, line 3: Usually, changing a passed version is strange even if there is a mistake.

rousseldenis (approved):

CHANGELOG.md
oleg68: > Usually, changing a passed version is strange even if there is a mistake.

Yes, agree, but later is better than never.

oleg68:
@larspalo could you approve this?

larspalo (approved):

Issue #1001, opened 2022-01-30
Add the Help button to the Midi Event Dialog


oleg68:
There is a help topic related to the Midi Event Dialog, but now it is impossible to navigate from the dialog to the corresponding help page

PR #1031, opened 2022-02-12
Added GODialogCloser and rewrote GOSPlash

oleg68:
This is a first part of changings for #1003 and #1001 

- Added the new class `GODialogCloser` that incapsulate the logic of closing modal and modeless dialogs.
- Rewrote `GOSplash`: the closing logic was removed from this class because it inherits `GODialogCloser` now

In the future the new class will be also used in Settings and MidiEvents dialogs.


rousseldenis (approved):

larspalo (approved):
The license dates should be changed to 2022.

Issue #1004, opened 2022-01-30
Help from the Settings Dialog is covered by the dialog itself.


oleg68:
When to press Help from the settings dialog, the opened help window is covered with the dialog window

paolo2504:
Correct. Where is the problem ?

oleg68:
The working window should be on top of all non working windows.

The dialog window is not clickable when the help is open. So it shouldn't cover the help window. Instead the help window should cover the dialog.

paolo2504:
> 

thanks Oleg. To say the truth, i dont experience what you say. I open the Settings dialog, click on Help, its dialog opens over but the below dialog is still clickable , selectable and can be navigated. It can also be closed (help dialog in this case stays open.
go r.3611 Windows 10 64b. This is why i asked which problem.

paolo2504:
https://user-images.githubusercontent.com/50132661/153083340-bcfc1f9f-9a03-4f81-aaa5-b3c51e5179b6.mp4

oleg68:
https://youtu.be/ipLdxJ93LBU

oleg68:
This is the root cause: https://stackoverflow.com/questions/7545804/modeless-parentless-wxdialog-still-always-above-wxframe-window-in-z-order

PR #1039, opened 2022-02-18
Bugfix/help overlapped

oleg68:
Resolves: #1004

The root cause of the bug: dialogs are always above other kind of windows under linux: https://stackoverflow.com/questions/7545804/modeless-parentless-wxdialog-still-always-above-wxframe-window-in-z-order

For fixing it the help frame has to have the extra style wxTOPLEVEL_EX_DIALOG. The new derived class GOHelpController creates the help frame with the extra style required.


larspalo (approved):

oleg68:
@rousseldenis could you approve this CR?

rousseldenis (approved):

Issue #949, opened 2022-01-06
GrandorgueTool in Window


paolo2504:
I just found that the GrandOrgueTool that comes with GO installation apparently doesn't work. It does no action. Tried also "grandorguetool -h" to get help.
At first i tried with winrar and 7z to compress an organ folder setting compression ratio to 0 (store) and renaming the file with .orgue. With File/install organ package GO does try to open the file but after a while it returns the error "no end record found".
Samplesets found on the net are usually compressed with normal level; GO does recognize them (after renaming to .orgue) but in this case the returned error is "only stored compression supported".
This to explain why i finally came to GrandOrgueTool, unfortunately discovering that it does not work...
thanks for help.

<from Discussion>
I know that organ packages are just compressed archives that can be opened with zip/rar tools. They are loaded in GO using file/install package. This is fine but I don’t see anywhere the uncompressed organs so I wonder how they are managed. I mean: I don’t believe they are uncompressed on the fly (in temp dir ?) and then deleted on close…
I post this question to understand if it would be better having only compressed samplesets just changing the file extension. 
note: after some search i found interesting posts by Lars Palo on the GO Tool included in the installation. I'm looking at it; however, at first sight it does not answer a basic Q: most samplesets com in .zip or .rar format. They are recognized by GO as packages but they fall in error because the compression ratio/method is usually not set to 0 (GO returns the error "only stored compression supported").
The GOTool i suppose is able to compress and generate organ packages from folder and subfolders: ok, but this means one has to uncompress samplesets downloaded from internet before submitting them to GO tool...
Question is at this point: what is the reald advantage of using Install Organ Package ? Size is the same... organ are loaded into ram no matter they are in folder structures or in one compressed file...
just to understand better. thanks.

![errore no end record found](https://user-images.githubusercontent.com/50132661/148430902-54f18e02-adc1-4e9a-91c9-d4108800297e.jpg)
![errore only stored compression supported](https://user-images.githubusercontent.com/50132661/148430906-fb1cb7cf-60b8-48b5-ac8a-97c0504db3b9.jpg)

_Originally posted by @paolo2504 in https://github.com/GrandOrgue/grandorgue/discussions/944_

paolo2504:
i've just looked at post #915 and now from Lars's words i understand much more. I see this is something that can be improved but not for the moment..

<However, i would appreciate to know if there is some way to make grandorguetool work under Windows. thanks.>

paolo2504:
[...]

Issue #868, opened 2021-11-18
Perftest, where?


oleg68:
### Discussed in https://github.com/GrandOrgue/grandorgue/discussions/867

<div type='discussions-op-text'>

<sup>Originally posted by **jason-barba** November 18, 2021</sup>
With regard to https://github.com/GrandOrgue/grandorgue/discussions/862 and other threads, where can I find the perftest application? My OS i Ubuntu 20.04 LTS.</div>

ahall41:
_make perftest_ gives an error:
```
.../grandorgue/src/tools/perftest.cpp:14:10: fatal error: GODefinitionFile.h: No such file or directory
   14 | #include "GODefinitionFile.h"
```

oleg68:
Fixed with #1076

PR #1076, opened 2022-03-07
Fixed Building PerfTests

oleg68:
This is the first commit for fixing performance-test related problems #868

- Renamed `perftest` to `GrandOrguePerfTest`
- Removed dependencies on the GrandOrgue application. The new classes `GOLogicalRect` and `GOResizable` were introduced
- Added unconditional building of GrandOrguePerfTest and packaging to the client distribution
- Renamed `perftest.cpp` to `GOPerfTest.cpp`
- Fixed sume failures when running the perftest.

No GrandOrgue behaviour should be affected.

There is still a remaining task: GrandOrguePerfTest can not be easy used on Windows because now it depends on graphical components and windows doesn't consider it as a console application.
For resolving this I need to split the current grandorgue code to a graphical and a graphicless parts and to move the graphicless to from `grandorgue` to `core` like it has been done with #949. It may require a lot of works.

@larspalo what do you think, should it be done in the same issue #868 or it is better to create a separate issue?


larspalo:
@oleg68 I think it's excellent to separate the graphical parts, but it can certainly be done as a separate issue if you think this is complete enough for now. Just, for actually running the PerfTest one must point to the directory where the test files exist (test directory of GO source). Thus, the usage error message on line 198 should perhaps be changed to:

`wxLogError(wxT("Usage: GrandOrguePerfTest <GO source tree>/test"));`

Or, should we include the test files in the binary distibutions and automatically point to them and remove the need for argument to the program?

oleg68:
> Or, should we include the test files in the binary distibutions and automatically point to them and remove the need for argument to the program?

Yes. I is a good idea. But the program parameter should rest for compatibility.

oleg68:
@larspalo I added packaging the test wav files. `GrandOrguePerfTest` uses them unless another test dir is specified in the command line

larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1085, opened 2022-03-17
Updated copyright

oleg68:


larspalo (approved):

rousseldenis (approved):
I'm just wondering if, to avoid this each year, we could adapt to '2009-TODAY' as I saw in many project ?

Issue #1095, opened 2022-03-23
Please provide an option to link against system-wide zita-convolver


svillemot:
It would be useful to have a CMake option similar to `USE_INTERNAL_RTAUDIO` and `USE_INTERNAL_PORTAUDIO`, when zita-convolver is available system-wide.

In the meantime, the following patch has been applied to the Debian package (for using the version of the library provided by the `libzita-convolver-dev` package): https://salsa.debian.org/debian/grandorgue/-/blob/debian/latest/debian/patches/use-system-zita-convolver.patch

PR #1099, opened 2022-03-24
Allow to use system zita convolver

vimpostor:
Resolves: #1095

I am right now creating an [Arch Linux package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=grandorgue-git&id=48868a5b5853c79b9729be64fa2ebaa2c91a349e#n30) for grandorgue, but it is frowned upon compiling and bundling libraries with applications instead of dynamically linking to system shared libraries.
All other submodules already do have an `USE_INTERNAL` switch to disable the bundled library and use a system library instead, except for the zita convolver.

This patch adds a switch for zitaconvolver, so that all libraries can be linked dynamically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ option(RTMIDI_USE_MM          "Enable RtMidi support for MM (Windows only)" ON)
 option(USE_INTERNAL_RTAUDIO   "Use builtin RtAudio/RtMidi sources" ON)
 option(INSTALL_DEMO           "Install demo sampleset" ON)
 option(USE_INTERNAL_PORTAUDIO "Use builtin PortAudio sources" ON)
+option(USE_INTERNAL_ZITACONVOLVER "Use builtin Zita Convolver sources" ON)
 option(GO_USE_JACK	      "Use native Jack output" ON)
 if (WIN32 OR APPLE)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -184,6 +185,13 @@ else()
   pkg_check_modules(PORTAUDIO REQUIRED portaudio-2.0)
 endif()
 
+# include zitaconvolver
+if (USE_INTERNAL_ZITACONVOLVER)
+  set(ZITACONVOLVER_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/submodules/ZitaConvolver/source)
+else()
```
oleg68, line 191: It would be more clear to set `ZITACONVOLVER_INCLUDE_DIRS` and `ZITACONVOLVER_LIBRARIES` to empty values here.

oleg68 (changes requested):
Pleas add the clearing the variables

oleg68:
@vimpostor please add an entry to CHANGELOG.md describing this change

vimpostor:
> Pleas add the clearing the variables



> @vimpostor please add an entry to CHANGELOG.md describing this change

Done

oleg68 (approved):

larspalo (approved):

Issue #983, opened 2022-01-22
GrandOrgue crashes when Convolution Reverb is activated while the organ is still running


matheus-j-ressurreicao:
I would very much like to report what could be a possible GrandOrgue error. Let's see. I went to perform a test with the convulation reverb feature enabled. It seems to work normally, provided that a valid impulse response has been selected. What happens, however, is that if the user changes any impulse response file or even turns the feature on or off when the organ is still running, it seems to me that GrandOrgue starts crashing and showing a lot of instability, with strong bursts in the audio that are only resolved either by closing the program or sometimes by pressing the "panic" button. I ask, in this spirit, if this could be some bug in the program or if the convolution reverb should not be changed when the organ is still running?

oleg68:
@matheus-j-ressurreicao I cann't reproduce this issue.

Could you provide more information: what operating system and what version of GO are you using?

Could you explain ``turns the feature on or off when the organ is still running`` step-by step?

matheus-j-ressurreicao:
[...]

PR #1101, opened 2022-03-26
Bugfix/reverb hang

oleg68:
This PR resolves one of two problems mentioned in https://github.com/GrandOrgue/grandorgue/issues/983 - the hang when `Panic` or exiting from Settings.

The reason of the hang was the same as in #702 and was fixed in #837: the standard ZitaConvolver code used posix-semaphores for synchronising threads.

These semaphores worked unstable: sometimes a thread calling `sem_wait` did not return when another thread called `sem_post`.

I changed the implementation with std::mutex and std::condition_variable and now it works stable.

### DIFF
--- a/src/grandorgue/sound/GOSoundReverb.cpp
+++ b/src/grandorgue/sound/GOSoundReverb.cpp
@@ -81,6 +81,7 @@ void GOSoundReverb::Setup(GOConfig &settings) {
       data[i] *= gain;
     if (len >= offset + settings.ReverbLen() && settings.ReverbLen())
       len = offset + settings.ReverbLen();
+    /*
     wxLogMessage(
       "GOSoundReverb::Setup before resample: offset=%u, len=%u, "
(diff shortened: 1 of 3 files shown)


larspalo (approved):

src/grandorgue/sound/GOSoundReverb.cpp
```diff
@@ -81,13 +81,15 @@ void GOSoundReverb::Setup(GOConfig &settings) {
       data[i] *= gain;
     if (len >= offset + settings.ReverbLen() && settings.ReverbLen())
       len = offset + settings.ReverbLen();
+    /*
```
rousseldenis, line 84: Are those logs comment wanted ?

oleg68: I added this log for debugging #983 and I forgot to remove it before some previous PR.

Because I'm still working on this issue, I still need this log, so I'd not remove it.

So now I'm commenting out this logging.

rousseldenis (approved):

PR #1105, opened 2022-04-07
Fixed sound distortion while a reveberation is active

oleg68:
Resolves: #983

The reason of the sound boost described in #983 is reading an uninitialised buffer when preparing data for reveberation (resample_block called from GOSoundReverb::Setup).

### DIFF
--- a/src/grandorgue/sound/GOSoundResample.cpp
+++ b/src/grandorgue/sound/GOSoundResample.cpp
@@ -146,14 +146,16 @@ float *resample_block(
     unsigned max_i = len - position_index;
 
     for (unsigned j = 0; j < SUBFILTER_TAPS; j += 4) {
-      if (j < max_i)
-        out1 += in_set[j] * coef_set[j];
-      if (j + 1 < max_i)
-        out2 += in_set[j + 1] * coef_set[j + 1];
-      if (j + 2 < max_i)
-        out3 += in_set[j + 2] * coef_set[j + 2];
-      if (j + 3 < max_i)
-        out4 += in_set[j + 3] * coef_set[j + 3];
+      unsigned k = j;
+
+      if (k < max_i)
+        out1 += in_set[k] * coef_set[k];
+      if (++k < max_i)
+        out2 += in_set[k] * coef_set[k];
+      if (++k < max_i)
+        out3 += in_set[k] * coef_set[k];
+      if (++k < max_i)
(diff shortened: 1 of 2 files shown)


src/grandorgue/sound/GOSoundResample.cpp
```diff
@@ -143,11 +143,17 @@ float *resample_block(
...
     float *in_set = &data[position_index];
+    unsigned max_i = len - position_index;
+
     for (unsigned j = 0; j < SUBFILTER_TAPS; j += 4) {
-      out1 += in_set[j] * coef_set[j];
-      out2 += in_set[j + 1] * coef_set[j + 1];
-      out3 += in_set[j + 2] * coef_set[j + 2];
-      out4 += in_set[j + 3] * coef_set[j + 3];
+      if (j < max_i)
+        out1 += in_set[j] * coef_set[j];
```
rousseldenis, line 150: I'm asking myself if there is no cleverer thing to avoid those variables (out1/2/3/4) to get the result ?

larspalo: @rousseldenis Well, that's how the resampling works. The expert on such things is @nickappleton who could better assess the quality of the code and possibly suggest improvements.

My initial thoughts are that if the cause of the original issue indeed is reading from an uninitialized buffer, perhaps that's where the real fix should be. However, what this PR seems to do is to make sure that one doesn't read out of bounds.

rousseldenis: Yes for sure. Just a thought about code

rousseldenis (approved):

src/grandorgue/sound/GOSoundResample.cpp
oleg68: I also think that we may eliminate out1/2/3/4 because they all are added and we may write an inner loop from 0 to 3. Making this change would not be too difficuld but it would require testing the new code for correctness: that the new code gives the same result as the old one. Unfortunally we don't have any unit tests now.

oleg68: Another thouth is that the  polyphase interpolation resampling is implemented in several places: here and in `GOAudioSection::MonoUncompressedPolyphase`/`GOAudioSection::StereoUncompressedPolyphase`.

It would be better to make a single inplementation that would be used in all places

larspalo (approved):

Issue #1097, opened 2022-03-24
Send MIDI event to Any device


willeke1234:
GrandOrgue 3.4.2-0 sends a MIDI event to all connected devices if the device is set to Any. The Help says "If a device is selected in the dropdown, GrandOrgue outputs that event only to that device. If no device is selected, GrandOrgue outputs that event to all selected output devices ( in the Midi Devices Tab) connected to the computer.". GrandOrgue 3.6.3-1 still sends the event to all devices but refuses to accept device Any in the Midi Event Editor. I get "Invalid MIDI event. Output device is not selected. Select one, set the type to None or delete this event.". Is this a bug or a feature? If it's a feature then the Help is outdated.

oleg68:
Iyt seems a bug. The capability of sending to all device was broken.

PR #1110, opened 2022-04-16
Fixed setting up output midi events for all devices

oleg68:
Resolves: #1097

This PR partially reverts the changes of #1008 that prohibited a capability of setting midi events to `Any device` up.


larspalo:
Do you intend to re-work the commented section in some way? If not, it could very well be removed completely so that we don't clutter the code with too much non functional garbage.

oleg68:
> Do you intend to re-work the commented section in some way? If not, it could very well be removed completely so that we don't clutter the code with too much non functional garbage.

You are right. I've just totally removed the extra code.

larspalo (approved):

rousseldenis (approved):

Issue #1066, opened 2022-03-03
Help and settings dialog problems


paolo2504:
i tested latest r3631 under linux (raspbian) and WIndows 10 and found several issues.
linux: 
help dialog opens but Settings dialog looses focus and cannot be anymore selectable.
In addition, if the minimize button in the Help modal dialog is selected, then it disappears and cannot be reopened anymore (even if it can be seen in the top app bar). As a result, as both Settings, Help and GO windows/dialogs are no more focused they all cannot be closed and it is necessary to kill GO process from terminal.
A less important issue is that Help dialog on first open is 0-size and appears as a small vertical bar (look at last screen recording).

Windows:
more or less the same issues apart there is no need to kill process. When opening Help, its dialog stays always over the Settings dialog: this is not good in Windows as it should get fore and back as the two dialogs take focus (see the screencam on Treesize that is a small tool i ran just for example). And again, if the minimize button is choosen then it doesnt appear anymore, even pushing again the Help button. No Help is later on available even from Menu. The only way to reset situation is to close and restart GO.

Pls look at the screen captures below.

https://user-images.githubusercontent.com/50132661/156633497-2e2ccb01-96e0-4ceb-a8ea-73b89fb0ca99.mp4

https://user-images.githubusercontent.com/50132661/156633676-944fb7f9-a297-49c2-8b54-765e684b197f.mp4

https://user-images.githubusercontent.com/50132661/156633719-d291543e-5151-4ea8-a8a1-78ec391cca71.mp4

https://user-images.githubusercontent.com/50132661/156643002-5c9ce9b0-dbe2-47c0-b952-6881d5d5db1b.mp4

https://user-images.githubusercontent.com/50132661/156644170-071babea-0b0b-469b-99b0-3b5909d21e6e.mp4

ahall41:
I found when building from the latest Git repository that minimise/maximise and close are missing from Help, but this is OK with the 3.6.2-1 branch. Help also always stays on top, but I can close with Alt+F4. I'm running Ubuntu 20.04.3 LTS and 18.04.6 LTS.

oleg68:
@paolo2504 I've just tested GrandOrgue 3.6.4-1 on Raspbian 32-bit. I cann't reproduce the problem:

1. When I click on `Help`, the help window appears with a reasonable size.
2. The minimize and maximize buttons are greyed (disabled), so it is impossible to get it out.
3. The settings window is not selectable while the help window is open, but it is a desirable behavior.

[...]

PR #1115, opened 2022-04-28
Fixed help issues

oleg68:
Resolves: #1066 

A help window has been opened from a modal window is not compatible from one has been opened from modeless window.

This PR differentiates them and checks and closes the old help window if it is not good.

Also some help-related code has been extracted from the `GOFrame` class.


larspalo (approved):

rousseldenis (approved):

paolo2504:
> Resolves: #1066
> 
> A help window has been opened from a modal window is not compatible from one has been opened from modeless window.
> 
> This PR differentiates them and checks and closes the old help window if it is not good.
> 
> Also some help-related code has been extracted from the `GOFrame` class.

Oleg

I see new version 3661 does fix bug 1 and bug 3 as referenced in my original post #1066 but not the bug 2.

Here follows how to reproduce it.
1. do File/Settings and click the Help button in whatever tab: the underlaying taskbar is still not selectable and you have to move the Help dialog outside it to be able to choose menu items. 

Try also this (ex-bug 1):
1. select Help/Help from the taskbar menu
2. now select File/Settings: the Settings dialog opens but the underlaying Help dialog is no more selectable.
3. at this time, if you click on the Help button of whatever tab, you'll see the underlying Help dialog close while a new one is created; however, you cannot choose anything in the Help and the only thing you can do is Cancel the Settings dialog.


Definitely: bugs 1 & 3 "partially" fixed, bug 2 not fixed imho.

regards.

paolo2504:
https://user-images.githubusercontent.com/50132661/166115495-5b8cf782-958b-41f5-a1a3-da11871c6bcf.mp4

oleg68:
@paolo2504 I can not understand what do you find wrong in your video. The Setting dialog is `Modal` so you can not select anything in another windows until you close it. 

But you can open a help window from the settings dialog, and this window also becomes modal. After this the settings dialog is only partially working. You have to close the help window before continuing do something in the Settings. This is an intended behavior.

paolo2504:
Oleg, leave it that way if you think it is correct. No problem my side. 
But I am sure this is wrong; I never saw a windows application/program where the help    dialogs are depending on the calling point/window. I can admit that since time Windows application link to internet Help pages… but when the Help manuals is installed locally , when you call it it’s always indipendent and you can work use the App and the Help applications at the same time. 
As I noted befeore, this is not crucial and I can get rid of it. 
Appreciate your effort.

Issue #853, opened 2021-11-15
Publish debug symbols on Github build for windows.


oleg68:
When GrandOrgue crashes on Windows (for example, #847), the OS creates a crash dump file in the `AppData\Local\CrashDumps` directory.

WinDbg is the only tool that can analyze the dump, but requires a debug information about the executable and libraries.

But there are two problems

1. Now the debug information is striped out from the executable and all libraries for decreasing the size of files
2. Even if the debug information was not striped out, mingw cross-compiler supports only DWARF format of it, but `WinDbg` requires the information in PDB format.

So we need to save the debug information in PDB format in separate artifacts during the github build.

The basic idea of implementation is to use [cv2](https://github.com/rainers/cv2pdb) for converting DWARF to PDB debug symbols. Unfortunally
1. It requires that MSBuild with C++ support would be installed. Shipping a few dlls from MSBuild does not help
2. This tool may run only on windows and cann't run under wine (maybe because MSBuild is not installed under wine)

PR #1127, opened 2022-05-10
Publishing windows debug information

oleg68:
Resolves: #853

- For converting debug symbols to the windows debugger compatible format [cv2pdb](https://github.com/rainers/cv2pdb) is used.
- It is a win32 executable and it requires MSVC to be installed
- I use a [special build of MSVC](https://github.com/GrandOrgue/DockerMsvcCpp) based on https://github.com/madduci/docker-msvc-cpp project
- for running cv2pdb under ubuntu I have to install wine32
- unfortunally there is a dependency problem while installing wine32 on a github runner machine: there is an odd version of `libpcre2-8-0` installed here buth there is no `libpcre2-8-0:i386` installed, but it is required by wine32.. This version is absent in the standard ubuntu repositories for installing the same version of `libpcre2-8-0`. So I need to replace the installed version `libpcre2-8-0` with a standard one.
- The resulting debug symbols are packaged into `grandorgue-debug`-version.`windows.x86_64.zip`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Added publishing debug symbols on windows builds https://github.com/GrandOrgue/grandorgue/issues/853
 # 3.6.6 (2022-04-29)
 - Fixed help issues https://github.com/GrandOrgue/grandorgue/issues/1066
 - Fixed saving organ settings on closing an organ or on finishing GrandOrgue https://github.com/GrandOrgue/grandorgue/issues/1069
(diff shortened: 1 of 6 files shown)


larspalo (approved):

rousseldenis (approved):

oleg68:
@larspalo @rousseldenis   I moved generation of the debug symbols from the shell script to cmake steps for better parallelism. Please approve this PR again.

larspalo (approved):

Issue #1104, opened 2022-04-04
Names of temperaments do not get translated


kerkovits:
Names of temperaments do not get translated regardless of the chosen language. Is this the intentional behaviour?

oleg68:
Yes, it is. Most names are the same in all languages so there is no reason to translate them.

willeke1234:
It looks like a bug to me. The names of the temperaments were translated, are in the .po files and are still translated but before the language setting is read. In `GOrgueApp::OnInit()` the language was set before and is now set after `m_Settings->Load()` where the list of temperaments is created.

kerkovits:
I see the point, but present situation is quite unclear for translators. All names of temperaments are marked by macro `WX_TRANSLATE`, suggesting that they should be translated. If the developers want that this must not get translated, the translation macro should be removed from the source code.

rousseldenis:
> I see the point, but present situation is quite unclear for translators. All names of temperaments are marked by macro `WX_TRANSLATE`, suggesting that they should be translated. If the developers want that this must not get translated, the translation macro should be removed from the source code.

@oleg68 Indeed, temperaments are translatable. Shouldn't this be addressed through #1139. Even if temperament names are equal, some locals could have some specific characters. I see also some that should be translated as 'Kellner's Bach tuning' - 'tuning' is a normal word.

rousseldenis:
@kerkovits @oleg68 Temperaments are well translated in last version. So, nothing to change

See in French:

![image](https://user-images.githubusercontent.com/19529533/182395608-65ecbc94-cdd5-4def-8c67-5eb0febcc5d1.png)

kerkovits:
> @kerkovits @oleg68 Temperaments are well translated in last version. So, nothing to change
> 
> See in French:
> 
> ![image](https://user-images.githubusercontent.com/19529533/182395608-65ecbc94-cdd5-4def-8c67-5eb0febcc5d1.png)

@rousseldenis @oleg68 This makes it even more spurious. This shows that the translated names of temperaments are known by the program, but they are not applied in the main window. Check the submenu Sound/Midi->Temperaments. It is in English. https://github.com/GrandOrgue/grandorgue/pull/1139 would correct this bug by loading the system locale before initializing the temperaments, but it introduces a regression that the language set in the Settings panel is not respected. In my opinion, the locale object should be initialized before loading the predefined temperaments but after reading the configuration files.

PR #1139, opened 2022-06-20
Init locale before loading configuration

kerkovits:
Resolves https://github.com/GrandOrgue/grandorgue/issues/1104

### DIFF
--- a/src/grandorgue/GOApp.cpp
+++ b/src/grandorgue/GOApp.cpp
@@ -33,7 +33,6 @@ IMPLEMENT_APP(GOApp)
 GOApp::GOApp()
   : m_Restart(false),
     m_Frame(NULL),
-    m_locale(),
     m_config(NULL),
     m_soundSystem(NULL),
     m_Log(NULL),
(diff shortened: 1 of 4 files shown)


oleg68 (changes requested):
This may not work if the language is redefined in the settings.

rousseldenis:
@kerkovits I think you should rebase on this master as the libwine build error has been solved

kerkovits:
@rousseldenis I see, thank you. Nevertheless, I succeeded to build it, but it still does not work as intended. The initialization of the locale could be moved into GOConfig.cpp just before the initialization of temperaments. However, if the locale is not the default one, the method `GetLanguageId()` still does not enforce the value set in the configuration. It seems, that the locale is customized by `m_config->Load();` in GOApp.cpp, However, I cannot find this explicitly in `void GOConfig::Load()` found in GOConfig.cpp. It seems that initalization of the locale is still too early in the middle of GOConfig::Load().

kerkovits:
https://github.com/GrandOrgue/grandorgue/pull/1164/ made this obsolete.

oleg68:
@kerkovits Thank you

PR #1155, opened 2022-08-10
Use CMAKE_INSTALL_LIBDIR to specify LIBINSTDIR on Linux

xhub:
The directory for system libraries is a nice conundrum on Linux.
It could be lib64 or lib or lib/<multiarch-tuple>. To sort this out,
CMake provides via GNUInstallDirs the CMAKE_INSTALL_LIBDIR variable,
see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
The distributions are then free to set this according to their tastes.

This should be tested in MacOS.


rousseldenis:
@oleg68 what do you think of?

oleg68:
It requires more testing for OSx

oleg68:
@xhub Could you enable GitHub actions in your fork in order to get the built version with your changes?

xhub:
> @xhub Could you enable GitHub actions in your fork in order to get the built version with your changes?

I've just done that but I had to push a new commit, it wouldn't let me manually run it

oleg68:
@xhub I've testet the built GrandOrgue in OsX. It works.

But the main problem of this PR is that `CMAKE_INSTALL_LIBDIR` is calculated for the building system, not for the target linux system.

So if we are building on a debian-based system (ex. Ubuntu), both .deb and .rpm packages get the `lib` directory. It is not problem for a redhat-based system, because it searches libraries in both directories.

But if we are building on redhat-based system, both . and .rpm packages get the `lib64`. It won't work on a debian-based.

So I think hardcoding `'lib` is better than using `CMAKE_INSTALL_LIBDIR`.

xhub:
Hm I see; it's sad that there is no agreement across linux distribution.

The issue I face is that on my distro (Gentoo) either `lib` or `lib64` is searched, depending on the architecture (32 vs 64 bits). If one build from source and do a system-wide install, it fails. I've create a package for Grandorgue to leverage the package manager and I had to create that patch. 

Would you consider a cmake option `USE_SYSTEM_LIBDIR` that if active, would set `LIBINSTDIR` to `CMAKE_INSTALL_LIBDIR`; if off, would set ` LIBINSTDIR` to `lib`? This would be very useful for those who don't use the prebuilt binaries.

oleg68:
> Would you consider a cmake option `USE_SYSTEM_LIBDIR` that if active, would set `LIBINSTDIR` to `CMAKE_INSTALL_LIBDIR`; if off, would set `LIBINSTDIR` to `lib`? This would be very useful for those who don't use the prebuilt binaries.

This option is acceptable, but it is not the best.

The current building model of GrandOrgue is at the first to make `x86_64` linux executables and at the second to package the same executables for different linux distributions. Now only .rpm and .deb are supported, but we'd like to add support of another distributions too.

So the right place of calculating the libdir is the second (packaging) phase, not the first (build) phase.

The desired solution is
1. to use CMAKE_INSTALL_LIBDIR at the build phase
2. to move libraries to the appropriate libdir at the packaging phase

@xhub Could you try to implement this approach?

xhub:
I understand the intention. However, I don't believe that would work for my case.
Gentoo is a source-based distro, not a binary one. The "recipe" to install GrandOrgue is [grandorgue-3.7.0.1.ebuild](https://github.com/xhub/xhub-repo/blob/master/media-sound/grandorgue/grandorgue-3.7.0.1.ebuild), which leverages the package manager. It's unclear what the benefit of using CPack would be. The only change needed is the one of this PR :)

Also, inasmuch I have some CMake knowledge, I don't have any CPack knowledge.

oleg68:
Yes, CPack does not give any benefits for source-based distros.

OK, please, introduce `USE_BUILD_SYSTEM_LIBDIR` cmake option with the default value `OFF`.

oleg68:
@xhub have you managed to add the `USE_BUILD_SYSTEM_LIBDIR` option?

xhub:
> @xhub have you managed to add the `USE_BUILD_SYSTEM_LIBDIR` option?

Sorry I got behind on this. I did implement it, but when testing it got bitten by #1184. Once I can locally test it, I will update this PR.

xhub:
@oleg68 I've now tested the cmake option and it works well here. The checks also passed successfully, with cpack it's hard to see what is happening as the lib files are not installed in the usual way (as in via `make install`). Sorry for the delay.

CMakeLists.txt
```diff
@@ -93,6 +94,11 @@ ELSE()
   set(BININSTDIR "bin")
   IF (WIN32)
     SET(LIBINSTDIR "bin")
+    ELSEIF (USE_BUILD_SYSTEM_LIBDIR)
```
oleg68, line 97:
```suggestion
  ELSEIF (USE_BUILD_SYSTEM_LIBDIR)
```

oleg68 (changes requested):

oleg68 (approved):

PR #1164, opened 2022-08-27
Fixed translation of temperaments' names

oleg68:
Resolves: #1104

This is an alternative fix of #1104 instead of #1139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed translation of temperaments' names https://github.com/GrandOrgue/grandorgue/issues/1104
 - Fixed saving ReleaseTail https://github.com/GrandOrgue/grandorgue/issues/1161
 - Moved build for OsX to OsX 11 https://github.com/GrandOrgue/grandorgue/discussions/1149
 # 3.7.0 (2022-08-11)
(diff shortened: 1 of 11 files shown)


oleg68:
@rousseldenis could you add @kerkovits as a reviewer of this PR?

larspalo (approved):

kerkovits:
In src/core/temperaments/GOTemperamentList.cpp, the switch from wxTRANSLATE to wxT results in that gettext will not extract the names of temperaments as translatable strings.

oleg68:
> In src/core/temperaments/GOTemperamentList.cpp, the switch from wxTRANSLATE to wxT results in that gettext will not extract the names of temperaments as translatable strings.

Yes, you are right. I reverted this change.

kerkovits:
I tested your build. It does not work as intended. In src/core/temperaments/GOTemperamentList.cpp, there are numerous lines looking like `group = _("Some string");`. These strings appear as submenus under the menu temperament, and they are always rendered in English regardless of the locale. (The reason is that the source now calls function `_();` before initializing the locale.)

oleg68:
@kerkovits I moved translation of temperaments groups too.

oleg68:
@kerkovits Have you managed to test the last version intermediate? You can download it from https://github.com/oleg68/GrandOrgue-official/actions/runs/2959025636

kerkovits:
I checked your sources and downloaded the last build. Thank you for your hard work, I think, it is perfect. I did not have time to check it yesterday. I closed the other pull request being obsolete.

oleg68:
@kerkovits Thank you

rousseldenis (approved):

Issue #1174, opened 2022-09-02
appimage always forgets cache and loads demo organ


jandeboever:
When starting grandorgue-3.7.1-1.x86_64.**AppImage** (or 3.6.7-1 and probably earlier versions as well) always:
- displays the log panel saying that the cache in no longer actual
- loads the demo organ

When closing de demo-organ and subsequently loading some other organ it:
-  again displays the log panel saying that the cache in no longer actual
- loads the requested organ.
- and the saved settings for that organ ARE remembered

When installing and running the grandorgue_3.7.1-1_amd64.**deb**-version these problems do NOT occur.

Background:
- running on Linux Mint V20.2 64-bit 
- Cinnamon desktop

Thanks in advance for looking into this

PR #1178, opened 2022-09-06
Bugfix/appimage in container

oleg68:
I started working on #1174 but I cann't build appimage with my environment, because the build script works only on a bare Ubuntu and I have Fedora and run Ubuntu in a podman container.

This PR fixes container-related issues.


oleg68:
Duplicated by #1179

PR #1185, opened 2022-09-10
Bugfix/appimage cache

oleg68:
Resolves: #1174

The reason why appimage did not respect the cache was taking the executable path into account when calculating the hash.

This PR excludes the path of resourses from calculation the hache.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed rebuilding the cache every time upon the appimage starts https://github.com/GrandOrgue/grandorgue/issues/1174
 - Fixed building appimage in an ubuntu container
 # 3.7.1 (2022-09-02)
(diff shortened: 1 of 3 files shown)


larspalo (approved):
Seems to work. However I did notice that I also get duplicate organ entries (of the included demo sampleset) with each run of the AppImage. Should this perhaps be a separate issue?

oleg68:
It is a separate issue: 
1. Appimage unpacks it content into a temporary dir (every time a new one) and runs GO from it
2. When GO runs from a new dir, it register the new instance of demo package
3. Earlier only packages were multiplied, but not organs. But after reworking the Organ tab, organs are multiplied as well.

I don't have any ideas how to prevent this behavior. Maybe not to allow to register several packages with the same id and deregister them automatically?

oleg68:

rousseldenis (approved):

Issue #1170, opened 2022-09-01
Crescendo pedal removes actual stops


krgrwrg:
**Problem**
* Crescendo pedal works different like on Hauptwerk. 
* Current behaviour : Instead of adding stops to current preset it changed my preset
* Expected behaviour :
** New stops will be added or removed when crescando value is changing.
** My stops activated before changing of crescendo value will stay unchanged

**Example**
* Crescendo pedal value : 0
* Stops activated : 32/32
* Crescendo pedal value : 1
* Stops activated : 3/32
* Crescendo pedal value : 32
* Stops activated : 32/32
* Crescendo pedal value : 1
* Stops activated : 3/32
* Crescendo pedal value : 0
* Stops activated : 3/32

oleg68:
@larspalo This behavior can be changed, but I'm not sure which behavior is right. What is your opinion? Should crescendo remove only the stops that were only added with it?

larspalo:
The main problem would be that in the real world you can find examples of both behaviors - Crescendos that overrides any current active manual selection, and crescendos respecting any existing manual selection. In my opinion the best solution would be to add another button on the Crescendo pedal that for each selected crescendo (A-D) individually would specify the "override mode" for that (and only that) crescendo.

krgrwrg:
I think correct behaviour is "override mode", because of crescendo pedal is for 'boost volume' by activating more stops than are current set.
In 3.X version of software is changing default behaviour of something what users may using for years wrong decision.
By my opinion, the best solution is to add button "override mode" - disabled by default to keep backward compatibility.

oleg68:
@larspalo @krgrwrg 

Where should I put this button on the Crescendo Panel?

![image](https://user-images.githubusercontent.com/11586506/189518311-f89d716b-be32-4bd7-bc98-ca4844a01bc1.png)

oleg68:
@krgrwrg Could you test the `Override` button on the crescendo panel in the intermediate build https://github.com/oleg68/GrandOrgue-official/actions/runs/3039571990

Does it satisfy your needs?

krgrwrg:
> @krgrwrg Could you test the `Override` button on the crescendo panel in the intermediate build https://github.com/oleg68/GrandOrgue-official/actions/runs/3039571990
> 
> Does it satisfy your needs?

Sorry, i was on vacation without laptop in mountains. I will test this new feature during weekend. Thank you for implementation.

krgrwrg:
[...]

PR #1187, opened 2022-09-13
Feature/crescendo smart

oleg68:
Resolves: #1170

This PR adds the Override button on the Crescendo panel. The state of this button can be saved with the organ settions or the preset.

The ODF section `CrescendoX` with the boolean key `OverrideMode is also supported.


src/grandorgue/GOSetter.cpp
```diff
@@ -723,12 +763,28 @@ void GOSetter::Crescendo(int newpos, bool force) {
   if (pos == m_crescendopos)
     return;
 
+  bool crescendoAddMode = !m_CrescendoOverrideMode[m_crescendobank];
+
   while (pos > m_crescendopos) {
-    m_crescendo[++m_crescendopos + m_crescendobank * CRESCENDO_STEPS]->Push();
+    const unsigned oldIdx = m_crescendopos + m_crescendobank * CRESCENDO_STEPS;
```
rousseldenis, line 769: @oleg68 Wouldn't it be more optimized to have one while() call per mode (AddMode or Override) to avoid testing crescendoAddMode at each iteration ?

oleg68: > @oleg68 Wouldn't it be more optimized to have one while() call per mode (AddMode or Override) to avoid testing crescendoAddMode at each iteration ?

Might be, but I prefer to reuse common pieces of code as more as possible. This code is more understandable.

rousseldenis: Ok, we'll see with usage.

Issue #1196, opened 2022-09-15
Add capability of loading and saving combinations to the `Combination Setter` panel. 


oleg68:
Now exporting and importing combinations are possible only with a menu.

It would be more convinient to do it with special controls in the `Combination Setter` panel. Hauprtwerk already has a similar capability.

oleg68:
@larspalo @rousseldenis 

I'm trying to add more controls to the sequencer panel.

What variant is better:
![classic](https://user-images.githubusercontent.com/11586506/231866667-92bf16a5-d10b-496f-a9ef-f3ea31b04101.png)

or

![newui](https://user-images.githubusercontent.com/11586506/231866714-48cbdf6d-91ef-458a-b162-8ccd79874d66.png)?

larspalo:
@oleg68 The first one. Not the green icons please.

larspalo:
Becacuse the icons doesn't actually say what the buttons are doing. An arrow right means what... Right? Whereas the text is immediately obvious to anyone that can read. Not to mention that the icons have a style that doesn't match the existing drawstop (buttons) at all.

larspalo:
@rousseldenis In my opinion, if we want to modernize the built-in graphics, it should be done as new additional bitmaps (and using them inside GO by setting new defaults). But they should all be really thought through so that they will harmonize together in a coherent style.

oleg68:
OK, let's choose the first variant.  I'm finishing working on the issue now. The evaluation build is already abvailable at https://github.com/oleg68/GrandOrgue-official/actions/runs/4693356345 .

As soon I adjust the help, I'll open a PR with it.

Issue #1195, opened 2022-09-15
Export combinations separate from the organ settings


oleg68:
Now combinations are saved in the same .cmb file as the organ settings. It rases several problems:

1. The menu entries `Import Settings` and `Import Combinations` are the neighbors, so they can be refused. I a user select `Import Settings` by mistake `instead of Import Combinations`, the organ settings will be overriten and it will take a lot of time.
2. The combination file is not human-readable so it cann't be used as a base for setting combinations on a different organ.

The idea is to export combinations only to a separate file in yaml format. Theese files would be stored in a separate folders GrandOrgue/combinations/, separated by organ names for preventing confusing.

oleg68:
@larspalo @rousseldenis 

This in an example of yaml-file with combinations saved:

```
info:
  organ-name: Oleg Home, op. 2
  grandorgue-version: v3.11.0-0.local
  saved-time: Sat 11 Mar 2023 21:10:29 MSK
sequencer:
  001:
    manuals:
      000:
        name: Pedal
        stops:
          001: PrincipalBass 16'
          002: Octavbass 8'
          003: Subbass 16'
          005: Posaune 16'
          007: Dry Gedacktbass 8'
          009: Superoctavbass 4'
          011: Pommer 4'
          017: Untersatz 32'
        couplers:
          002: III/P
          003: II/P
      001:
        name: Hauptwerk
        stops:
          001: Montre 8'
          002: Roerfluit 8'
          005: Mixtura
          006: Principal 8'
          007: Octava 4'
          008: Superoctava 2'
          014: Quinta 2 2/3'
          018: Bordun 16'
          019: Bourdon 8'
      002:
        name: Brustwerk
        stops:
          001: Geigenprincipal 4'
          004: Viola Pomposa 8'
          005: Rohrfloete 4'
          012: Gedackt 8'
          017: Geigenprincipal 8'
          018: Salicette 4'
      003:
        name: Oberwerk
        stops:
          005: Orchesterfloete 4'
          008: Doppel Gedackt 8'
          014: Piccolo 2'
        couplers:
          001: 4' III
  002:
    manuals:
      000:
        name: Pedal
        stops:
          001: PrincipalBass 16'
          002: Octavbass 8'
          003: Subbass 16'
          007: Dry Gedacktbass 8'
          011: Pommer 4'
          017: Untersatz 32'
        couplers:
          003: II/P
      001:
        name: Hauptwerk
        stops:
          001: Montre 8'
[...]

Issue #708, opened 2021-09-01
[Feature Request] Banked Divisional Pistons (Ergo: A Traditional US/UK Combination Action System)


pgcrowe:
The lack of banked **divisional** pistons is one of the greatest omissions on the GrandOrgue platform — one that past developers have refused to address. GrandOrgue's current combination action system is atypical to nearly every combination action in the US and UK. While one can compromise by making a banked **divisional** piston out of a banked **general** piston, the process is unintuitive, and leaves too few pistons left on a given memory level. Since code is already in place for banked **general** pistons, it couldn't be too difficult to add banked **divisional** pistons. It was on my GrandOrgue coding to-do list, but you can probably add it faster than I (while I commit the code for per-pipe release truncation).

larspalo:
Well, this issue is not as trivial as it might seem.

GO can presently provide a completely **traditional** US/UK general/divisional combination system built in the panels. You have 50 generals and 10 divisionals (per division).

On top of this it's possible to re-purpose the 1000 free combinations (sequencer) existing (per .cmb file) using them as a "banked" version of generals - but in truth it's still the same 1000 free combinations existing in the background - only they are accessible in steps or "banks" possible to configure for the user at will. Also it's possible to use them in other ways, like pointed out above...

By the way, in a **really traditional** US/UK combination system you don't have banks, you just get what you see as pistons on the organ. It's only in more modern versions you have banks of generals and divisionals to overcome the obvious limitation of the old traditional system.

Anyway, while the banked generals could simply be a re-purposing of existing functionality it's not really the same with the divisionals since we have nothing to re-purpose there.

What we can do is of course to provide more than the current 10 divisionals each division, and provide a way to access them. However, the actual numbers of either divisionals or generals is not the same for every organ. While it's easy for us to say that in GO there could be a total of 100 divisionals in 10 banks of 10 each built in, in truth, most US/UK organs won't have that number of banks or pistons for each division - usually they are in some other number like 8, or anything really...

This makes me think that inside GO, if we at all want to expand on the number of divisionals, we should perhaps not at all assume a certain setup of banks/frames but just provide a larger number of total divisionals available and make the user responsible to map them to his console in whatever way it's working.

oleg68:
@larspalo Would it be correct if switching the current bank of the genetal setter (±10, ±100) also switched the bank of all divisionals without changing the current combinations until some divisional would be pressed?

larspalo:
[...]

PR #1197, opened 2022-09-15
Combination dir 

oleg68:
I'm going to start working with combination-related issues: #708, #1195, #1196

This PR moves the combinations-related source files to the separate `src/grandorgue/combinations` dir.
I've tried to make separate directories for MVC, but unfortunally the current interclass calls do not fullfil the MVC approach.
It is just moving and renaming. No behavior should be changed.


larspalo (approved):

oleg68:
@rousseldenis could you approve this PR,

rousseldenis (approved):
@oleg68 Thanks for this.

As a recurrent criticism occurs on some forums about code documentation, when such refactoring occurs, shouldn't we include code comments as much as possible ?

This in order to enhance reviewers understanding and further contributors.

oleg68:
@rousseldenis you are right. Most of existing code is not documented now and I don't understand it. When I make some changes in a piece of code, I must to understand it and I'm trying to put some comments.

But the changes like this are more formal: some classes and variables are renamed and their files are moved, and it introduces a lot of formal changes to all references to the renamed/moved objects. Most of the changes are made automatically, So I cann't comment them now.

I doesn't consider commenting the existing code as a completable task, but it is a good practice to add comment to a code piece changed manually.

PR #1204, opened 2022-09-17
Fix various typos

svillemot:
Most of them were found using lintian.

Note that I did not update the translation files correspondingly in this commit, but this should most certainly be done if this commit is merged.


rousseldenis (approved):
Seems legit.

oleg68 (approved):

PR #1211, opened 2022-09-30
Added divisional combination banks

oleg68:
Resolves: #708


oleg68:

rousseldenis:
> @larspalo @rousseldenis Could you take a look a this PR? This PR changes the GrandOrgue behavior in a not obvious way and brings incompatibility with the old saved settings (the old saved divisional combinations of the main GUI pannel are mapped to the bank K, not to the default A), so I'd like to see your feedback. But if you don't answer, I'll merge this changes by myself.

I'll try to review it tonight. I want to take time to undersand.

oleg68:
@rousseldenis You can download [the intermediate build](https://github.com/oleg68/GrandOrgue-official/actions/runs/3157374568) for testing how does it work. Open `Panel`->`Divisionals` for having access to the banked divisionals.

larspalo (commented):
I think this is the right way to go with increasing the amount of divisionals. However, it currently does create loads of reported log warnings with older saved settings that are kind of not really desired... A smother way of integrating the old settings into the new ones would have been nicer.

oleg68:
@larspalo 

Earlier there were two independing sets of divisionals: the first was the divisionals on the main GUI panel (if they were present in ODF), and the second was the divisionals on the `Divisionals` panel.

There was a very complex code of mapping the first divisionals to the combination numbers (https://github.com/GrandOrgue/grandorgue/pull/1211/files#diff-a9cc93bad66d4df763ee8ac9c3ccf35400f3b89ce77fa4bf0968aba7ea87b566, lines 217-228) with capability of remapping, usualy they were 000, 001, ..., 007 at each manual.

The second divisionals were mapped to 100 - 109.

I made a single flat division combination space from 000 to 199 for each manual. Divisional buttons on the main GUI panel and on the Divisional panel all are hardly mapped (without capability of remapping) to the same combinations xx0 - xx9, where `xx` is the current bank number. Switching the current bank affects both sets of buttons.

It may be incompatible with the old presets and the combinations saved earlier, so if the combination has been lost, the warning message is desired.

It brings some incompatibility with ODFs with built-in divisional combinations. But I don't know any existing sample set for GrandOrgue with built-in divisional combinations and I doubt that they ever exist.

So I'd like to push the new divisional scheme without supporting the old one and I'd like to remove this too complex code https://github.com/GrandOrgue/grandorgue/pull/1211/files#diff-a9cc93bad66d4df763ee8ac9c3ccf35400f3b89ce77fa4bf0968aba7ea87b566, lines 217-228.

Do you have any other idea how to manage the old preses smoother?

larspalo:
Actually, in GO, there's both the old style divisionals and generals that's possible to configure via the ODF as well as the newer built in divisionals and generals (which really is somehting different from the setter - sequencer) which is available from within the panel menu in GO.

I can understand the desire to re-arrange the combination handling in GO by completely eliminating the ODF defined versions - but that would not only break backwards compatibility it would also eliminate the possibility to implement fixed console combinations on an organ. And believe it or not, many older samplesets do indeed define both divisionals and generals via ODF (as that was the only option in the early days). Not to mention that there is a huge number of real life organs that do have fixed combinations either instead of user definable or right beside the divisionals/generals/free combinations/setters etc.

As a summation: I'm not at all convinced about the approach you're taking here is the best path forward.

oleg68:
@larspalo The new approach provides a possibility to implement fixed console combinations in ODF but in a different way, not compatible with the old one.

I'l try to bring some level of compatibility.

rousseldenis:
Tested and works like it does.

As @larspalo said, it should be great to be compliant with former ODF structure.

oleg68:
@rousseldenis @larspalo I managed to make the new approach compatible with the old ODF settings: the first divisionals are load from DivisionalXXX and then from SetterXXXDivisionalXXX.

If the static combination is present in ODF it will be loaded until it is not overriden by SetterXXXDivisionalXXX in the preset.

larspalo:
@oleg68 I think you're very much on the right track with this now! As the generals defined in the ODF is not intermingled with the setter generals, I think that the same approach should be taken with the divisionals.

That is, if divisionals are set in the odf - they are defined there and kept there without affecting the divisionals defined inside of GO which is odf independent (in a way). In this way we won't have any issues with any older organs while still having a much improved divisional (banked) system available from within GO that's not dependent, or interfering, with what the organ might have defined.

The outcome is exactly what we've always tried for - allowing ODFs created by organ builders to stay as true to the original organ as possible while at the same time allow users to exploit all features they like to use from GO itself.

I'll test this latest build a little further and see if I have more suggestions...

larspalo:
Testing with PiteaMHS organ package yields some very strange behaviours that will need to be traced. When activating any divisionals on the main panel the Set button (on the main panel)) suddenly starts displaying current active combination. Also, pressing GC will crash GO.

oleg68:
> As the generals defined in the ODF is not intermingled with the setter generals, I think that the same approach should be taken with the divisionals.

My approach is the opposite: to unify all types of generals. There are tree different sets of generals: the main panel generals, the banked generals panel and the combination setter panel. I'd like
1. To make a single set of generals  from 0 to 999
2. To map banked generals (50 generals and 20 banks) to them
3. To make the main panel generals the same as banked generals; they would be affected by the current bank.

This enchancement of generals is a subject of a separate PR, but I've made the main panel generals the same as the banked generals so they are affected by the switching the bank.

> That is, if divisionals are set in the odf - they are defined there and kept there without affecting the divisionals defined inside of GO which is odf independent (in a way). 

It was not true in the prior versions: every ODF-defined divisional might be redefined with the `Set` button unless it was `protected`. But now they are mapped to the divisionals of the bank `A` and can be either protected or redefined with `Set`.

oleg68:
> When activating any divisionals on the main panel the Set button (on the main panel)) suddenly starts displaying current active combination

Yes, I managed to reproduce it. I'll debug this.

larspalo:
@oleg68 I don't agree with your suggested strategy immediately. Neither for generals nor divisionals. My main objection is that it's not clearly separating the original organ modeled in the odf with the user opted usage of builtin GO features.

If, and really - and only if, you're hell bent on wanting to simplify the combination structure at least all the mapping of older builtin combinations should be made silently without the tons of warnings, just a simple note that the odf contains older combinations mapped to the new system.

Also, if your approach really is a way forward to consider we need to in documentation suggest to organ sampleset producers to define the setter elements for the generals and divisionals from the odf so that they in fact are the same as the GO builtin (and of course provide the means to it).

Just as a side question in your quest to simplify combination handling, would you intend to keep the sequencer combinations apart from the generals as a separate combination system?

larspalo:
> I cann't reproduce this. Could you provide a step-by-step instruction?

1. I downloaded grandorgue-3.9.0-0.7.linux.x86_64.tar.gz from https://github.com/oleg68/GrandOrgue-official/actions/runs/3260513287
2. Extract and run GO with pasuspender
3. Fix sound output that's changed device name
4. Load Pitea school of Music sampleset from the organ package
5. Press a few of the built in divisionals on the main panel (see the Set button change)
6. Then press the G0 (GC) to the right below lowest manual - GO crash.

oleg68:
@larspalo the reason why I want to make the simplification is that some real organs have banked divisionals and there was no any way to define them in ODF. My changes add this capability.

So if the ODF and GO combination systems bothh are banked, are there any reasons to keep them separate? Would it better to add some GUI sign showing whether this combination came from ODF or not?

larspalo:
> So if the ODF and GO combination systems both are banked, are there any reasons to keep them separate? Would it better to add some GUI sign showing whether this combination came from ODF or not?

I'll copy from my post above:

This however will create even more of a mess and possible confusions. If the divisionals or generals were defined in the odf as protected and still remain so in the unified import into the GO builtin versions - the users possibility to do whatever he wants is limited. If on the other hand the odf defined versions are ditched - it's impossible to model for instance fixed combinations on an organ or an older combination system. In my opinion either way is a failure.

For full context read above...

oleg68:
@larspalo Modelling of the old fixed combination is still possible. They should be defined in ODF as protected and should be mapped to 1-9 of the bank A. Then a user wants to define his own combinations he has to store them to another banks. So A will represent built-ins, and B-T represent user-defined.

rousseldenis:
> @larspalo Modelling of the old fixed combination is still possible. They should be defined in ODF as protected and should be mapped to 1-9 of the bank A. Then a user wants to define his own combinations he has to store them to another banks. So A will represent built-ins, and B-T represent user-defined.

IMHO that behaviour seems acceptable. Maybe something in GUI to indicate user (char color - A in black, B-T in red ?)

oleg68:
Now I'm planning to revert some of my changes and to keep the main panel divisionals separate with the banked divisionals. So the banks will be used only on th Divisionals panel.

Merging the several independing combination systems to the single unified system is a subject of a discussion and should be done with a separate PR.

oleg68:
@larspalo @rousseldenis I've separated the ODF divisionals with GrandOrgue dibvisionals as Lars suggested.

Only combinations accessible from the `Divisionals` panel are banked.

Please revie the new behavior.

larspalo:
@oleg68 The banked divisionals and the older divisionals in the odf now work exactly as they are expected as far as I've been able to test it. The previous bugs with the Set and GC buttons also are fixed. Great work!

Now, the only thing I'd like to expand upon is that the banked divisionals should be available as Setter Elements in the odf just like the banked generals are. This would of course also make them available just as any other GUI Element. The help would need to be updated to reflect this (The Grand Orgue file format). When this is fixed I'll gladly approve of this PR as another major step forward for GO!

oleg68:
> the banked divisionals should be available as Setter Elements in the odf just like the banked generals are.

They are available already.

> The help would need to be updated to reflect this

I updated the help in the last commit.

> When this is fixed I'll gladly approve of this PR as another major step forward for GO!

Now everything is fixed and I'm waiting for your approval.

rousseldenis:
@oleg68 As already discussed, for new code, shouldn't it be great to have comments to help review/debug ?

oleg68:
> @oleg68 As already discussed, for new code, shouldn't it be great to have comments to help review/debug ?

Yes, I've added some comments to the new files `src/grandorgue/combinations/GODivisionalSetter.h`, `src/grandorgue/combinations/GODivisionalSetter.cpp` and to the new code in `src/grandorgue/combinations/model/GODivisionalCombination.h`

larspalo:
@oleg68 In file help/grandorgue.xml on line 7077 to 7100 the XXX and YYY is not clearly expressed (or is simply wrong). Also, you've forgotten to add Setter before the first number in all places.

I have to specify DisplayAsPiston=Y to get [SetterXXXDivisionalYYY] showing up under the manuals (and respect the DispButtonRow and DispButtonCol values). Expected behavior is that they should display as piston by default (as a divisional). The same would go for [SetterXXXDivisionalPrevBank] and [SetterXXXDivisionalNextBank] (now they all display as drawstops by default).

oleg68:
> In file help/grandorgue.xml on line 7077 to 7100 the XXX and YYY is not clearly expressed (or is simply wrong). Also, you've forgotten to add Setter before the first number in all places.

Fixed

oleg68:
@larspalo 
> I have to specify DisplayAsPiston=Y to get [SetterXXXDivisionalYYY] showing up under the manuals (and respect the DispButtonRow and DispButtonCol values). Expected behavior is that they should display as piston by default (as a divisional). The same would go for [SetterXXXDivisionalPrevBank] and [SetterXXXDivisionalNextBank] (now they all display as drawstops by default).

I added the compatibility of setting DisplayAsPiston=Y for setter buttons. Please try the last version

oleg68:

oleg68:
@larspalo @rousseldenis Because my last changes were submitted about one week ago, tomorrow I'm going to merge them by myself.

rousseldenis:
As it is a big change, it requires more time to review :-)

oleg68:
> As it is a big change, it requires more time to review :-)

I haven't seen any review activity untill now since 6 days ago. But If your review is in progress, I'm ready to wait more.

rousseldenis:
@oleg68 By the way, is that normal your github account is not the normal one for the commits ?

rousseldenis:
image


There is not link to your github normal account (oleg68)

oleg68:
> There is not link to your github normal account (oleg68)

The link to my account is just after the `from` in the first line.

What have I do for converting my name to the link?

rousseldenis:
With a git commit amend, you can do it for the last commit but I found this :

I found this: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-a-single-commit

oleg68:
> With a git commit amend, you can do it for the last commit but I found this :
> 
> I found this: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-a-single-commit

What commit autor have I use for working the hyperlink in github?

rousseldenis:
> > With a git commit amend, you can do it for the last commit but I found this :
> > 
> > I found this: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-a-single-commit
> 
> What commit autor have I use for working the hyperlinks in github?

If you have changed your email address in GitHub, that could be the cause.

rousseldenis:
One note: if you have changed email address, don't drop former ones as you could loose your commit history. Keep them forever.

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -135,111 +135,111 @@ enum {
...
-    {wxT("CrescendoA"), ID_SETTER_CRESCENDO_A, true, true},
-    {wxT("CrescendoB"), ID_SETTER_CRESCENDO_B, true, true},
-    {wxT("CrescendoC"), ID_SETTER_CRESCENDO_C, true, true},
-    {wxT("CrescendoD"), ID_SETTER_CRESCENDO_D, true, true},
-    {wxT("CrescendoPrev"), ID_SETTER_CRESCENDO_PREV, true, true},
-    {wxT("CrescendoCurrent"), ID_SETTER_CRESCENDO_CURRENT, true, true},
-    {wxT("CrescendoNext"), ID_SETTER_CRESCENDO_NEXT, true, true},
-    {wxT("CrescendoOverride"), ID_SETTER_CRESCENDO_OVERRIDE, true, true},
-    {wxT(""), -1, false, false},
+    {wxT("Prev"), ID_SETTER_PREV, true, true, false},
```
rousseldenis, line 138: As I don't remember my c++ basis, I'm wondering if the dictionary can't be initialized with default values in order to simplify writings here ?

rousseldenis (approved):
Tested basically without the ODF stuffs + code review

src/grandorgue/combinations/GOSetter.cpp
oleg68: This kind of initializers doesn't respect the default values. A code without the values cann't be compiled without errors.

Issue #1212, opened 2022-09-30
Divisional without stops stays pressed


willeke1234:
Expected behaviour:
The piston of a divisional or general without stops doesn't stay pressed.

Actual behaviour:
The piston of a divisional without stops stays pressed. The piston of a general without stops doesn't stay pressed.

The [Walcker Wildervank sample set](http://www.pbrd.nl/index.php/vo/downloads/summary/3-hw-1-grandorgue-versie/8-walcker-wildervank-volledige-versie) has C divisionals, a divisional GC. They should work like GC, like they did, and not stay pressed.

oleg68:
The intended behaviour is to stay a piston pressed whether the combination has stops or not. So the current behavior of divisionals is correct (it was changed in #1068), but the behavior of generals is not correct.

willeke1234:
Why did the intended behaviour change? How do real pistons work?
Maybe we need a discussion about when a piston stays pressed and when it is released. At least it should be consistent and predictable.
See also #1209

oleg68:
Now the behavior is predictable: when you press the button and it switches the stops, the button became pressed.

#1068 explains why the behavior of divisionals is changed.

willeke1234:
How about GC?

oleg68:
Now GC is not a combination, but only a `resetter`. If you have any ideas about GC, post them at #1220

PR #1214, opened 2022-09-30
Fixed an empty stop set to a general combination

oleg68:
Resolves: #1212

Now general buttons are displayed as pushed even there aren't stops in the combination.


oleg68:

rousseldenis (approved):

oleg68:
@larspalo Could you take a look at this PR?

larspalo (commented):
Actually, technically this might be just correct. A general is always a physical button that can be pressed weather any stops are set to that combination or not.

oleg68:
> Actually, technically this might be just correct. A general is always a physical button that can be pressed weather any stops are set to that combination or not.

Yes, but the new behavior is exactle the same as you have written. Could you approve this PR?

Issue #1209, opened 2022-09-26
Crescedo pedal problems with presets ( general pistons )


krgrwrg:
current preset ( general ) is not lighted as active after using crescendo pedal
- is little bit confusing - i don't know which piston is activated
- especially when i assigned led buttons on midi controller to general pistons, after using crescendo pedal all leds are off.

when in switch between presets ( generals ) crescendo is ignored, but when i move it up or down, sound is changed.
- when i forgot to move crescendo pedal down before i swap between presets and when i play some note, it produced unwanted sounds ( especially when i move it down to zero value ).

PR #1219, opened 2022-10-04
Fixed generals buttons behaviour with the crescendo in add mode

oleg68:
Resolves: #1209

When current crescendo is operating in the Add mode

1. The current pressed general piston remains pressed when crescendo position is changed,
2. When a general piston is switched all stops added with the current crescendo position remain switched on.


larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1215, opened 2022-10-01
[bug]Initial size of window "Select organ to load"


willeke1234:
The window "Select organ to load" is very wide and the width can't be changed. The initial size of this window used to be

![image](https://user-images.githubusercontent.com/97431382/193400157-4e5145e6-1eb5-44ee-b13b-c4d7e7c67934.png)
 
and now is

![image](https://user-images.githubusercontent.com/97431382/193400175-372dba9f-2acf-4555-9071-3c505de53600.png)

The initial size should be a usable size.

PR #1221, opened 2022-10-06
Fixed size of the Organ Selection Dialog

oleg68:
Resolves: #1215

1. Removed columns `Builder info` and `Recorded date`
2. Decreased size of the Path column
3. Made the dialog size is not depending on the data size


larspalo (changes requested):
Shouldn't help and help images be changed to reflect this change also?

oleg68:
> Shouldn't help and help images be changed to reflect this change also?

Indeed. I've changed the image.

oleg68:

larspalo (approved):

rousseldenis (approved):

PR #1232, opened 2022-10-29
Fixed playing multitrack midi files with changes of tempo https://github.com/GrandOrgue/grandorgue/discussions/1225

oleg68:
As discussed at #1225 GO played midi files with tempo change incorrectly: only the first track followed the tempo markers and all other tracks were played at the last tempo.

Now all tracks use the tempo map from the first track


rousseldenis (approved):
Functionally tested + review

larspalo (approved):

Issue #1208, opened 2022-09-23
Control organ with a touchscreen


nebuloss:
When you control organ with a touchscreen, you cannot interact properly with the stop pannel. You have to tap several time on the stop to activate or deactivate it. However other panel such as combination panel works properly.

oleg68:
@nebuloss I cann't reproduce it with the Barton organ that has a separate ``Stop View`` panel: the stops are switched with the first tap.

Other organs have stop buttons on the main GUI panel and they are also switched with the first tap.

Could you provide more information: what organ, what panel?

nebuloss:
I am currently using an organ I built  myself from hauptwerk sample sets.
But I have also this issue with the default organ.
However, because the default organ produces sounds when you activate or deactivate stops, I tested if my tap is detected from grandorgue.
It seems that grandorgue detect each tap because it produces  a short sound. The problem might come from the delay allowed between each tap. I think that grandorgue just quickly activate and deactivate the stop.

oleg68:
@nebuloss What happens when you tap a stop with the demo organ? The sound is produced, but the stop remains deactivated, does'n it?

nebuloss:
When I tap a stop with the demo organ, it does not work most of the time. It remains activated or deactivated but it always produces a sound.
However with a mouse click, it works all the time.
My touch screen works perfectly with other apps in my system. 
(I am running the arm64 debian version of grandorgue)

oleg68:
1. What version of GrandOrgue do you use?
2. What desktop environment do you use (gnome, kde, lxde etc)?
3. Are you running on Wayland or on XOrg?
4. Do you have any midi-attached pistons for displaying and controling the stop state?

nebuloss:
To be clear, I am running grandorgue on raspberry pi 4b with 8GB of RAM and an external soundcard.
I have 2 midi keyboard and a midi pedalboard.
I only use keyboard additional buttons to control registrations.
I have plugged one touchscreen to control the computer.
I flashed raspberry pi os 64 bits on the board. It is based on debian 11 (bullseye) and uses lxde as desktop environment and X11 as window server. 
Because the CPU is an arm cortex A72, I have installed the arm 64 debian package of grandorgue.

oleg68:
@nebuloss Thank you for information.

I managed to reproduce this problem with Raspbery Pi. There are problems somewhere between the touchscreen, lxde desktop and GrandOrgue. I will research more.

[...]

PR #1240, opened 2022-11-04
Fixed controlling stops with a touchscreen on Raspberry Pi https://github.com/GrandOrgue/grandorgue/issues/1208

oleg68:
Resolves: #1208 

When a user taps a touchscreen, most window managers emulate two events:
- MouseLeftDown
- MouseMove

But the Raspbian desktop emulates three events:
- MouseMove
- MouseLeftDown
- MouseMove
It caused a stop was enabled and just was disabled

I added a deduplication code for processing mouse events. It broke processing of double clicks, so I added separate methods for MouseLeftUp and MouseLeftDblClick


rousseldenis (approved):
Functional + code review

larspalo (approved):

PR #1247, opened 2022-11-06
Bugfix/GitHub actions

oleg68:
Got rid a lot of deprecation warnings on github build


rousseldenis (approved):

larspalo (approved):

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3.1.0
```
VladimirSlavik, line 18: Consider bumping to the next major version, like `uses: actions/checkout@v3` That way you don't have to care about 3.2, 3.3... you get the idea :-)

PR #1248, opened 2022-11-07
Removed extra class GOGUIMessageStataTracker

oleg68:
There was a helper class `GOGUIMessageStateTracker` for storing several `GOGUIMouseState` instances at once, but this capability is not used. So I've removed this class and I've replaced it usage with direct usage of GOGUIMouseState.


rousseldenis (approved):
LGTM

larspalo (approved):

PR #1256, opened 2022-11-15
Refactored work distribution code on loading organ in parallel threads

oleg68:
1. Moved GOLoadThtread to the new `loader` subdirectory.
2. Encapsulated the code of distribution of work across threads from multiple places of ``GOOrganController`` to the new class `GOCacheObjectDistributor`.
3. The methods `GOEventDistributor::GetCacheObjectCount` and `GOEventDistributor::GetCacheObject` were replaced with the new one `GOEventDistributor::GetCacheObjects`.

I'm planning to extract the code of ``GOOrganController::Load`` and ``GOOrganController::UpdateCache``  to a separate class in the future because they are very large and complex.


larspalo (approved):

rousseldenis (approved):
Nice

rousseldenis:
My just little question, could the vector use be more bulky in terms of memory usage ?

rousseldenis:
My just little question, could the vector use be more bulky in terms of memory usage ?

oleg68:
> My just little question, could the vector use be more bulky in terms of memory usage ?

No, it could not, because the vector is passed by reference. So `GOObjectCacheDistributor` uses the same memory as the vector keeps at 'GOEventDistributor::m_CacheObjects'

PR #1257, opened 2022-11-15
Extracted a model-related part from GOEventDistributor to GOEventHandlerList

oleg68:
Towards to removing dependencies from the GOModel to GOOrganController I splitted to two classes control/GOEventDistributor and model/GOEventHandlerList.

GOEventHandlerList contains only a collection of various event handler; GOEventDistributor distributes events across the handlers stored in the model.

It is just refactoring. No GrandOrgue behavior should be changed.

### DIFF
--- a/src/grandorgue/dialogs/GOMidiListDialog.cpp
+++ b/src/grandorgue/dialogs/GOMidiListDialog.cpp
@@ -11,7 +11,6 @@
 #include <wx/listctrl.h>
 #include <wx/sizer.h>
 
-#include "control/GOEventDistributor.h"
 #include "midi/GOMidiConfigurator.h"
 
 #include "GOEvent.h"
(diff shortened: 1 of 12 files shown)


src/grandorgue/dialogs/GOMidiListDialog.cpp
```diff
@@ -69,8 +70,9 @@ GOMidiListDialog::GOMidiListDialog(
   buttons->Add(close);
   topSizer->Add(buttons, 0, wxALIGN_RIGHT | wxALL, 1);
 
-  for (unsigned i = 0; i < midi_elements->GetMidiConfiguratorCount(); i++) {
-    GOMidiConfigurator *obj = midi_elements->GetMidiConfigurator(i);
+  for (unsigned i = 0; i < midi_elements.size(); i++) {
```
rousseldenis, line 73: @oleg68 Couldn't we use the iterator like in other code parts as midi_elements is a vector?

rousseldenis: oooh, yes, didn't see

rousseldenis: But don't you have access to index on the vector one object ?

rousseldenis: I found you can use std::distance(vector.begin(), element).

It's nitpicking, just that I like iterator writing instead of indexes one. :-)

rousseldenis: But I saw distance() will do a loop, so no.

larspalo (approved):

rousseldenis (approved):

PR #1262, opened 2022-11-17
Added config to GOModel

oleg68:
Towards to future removing dependencies on GOOrganController from the model


rousseldenis (approved):
Trivial

larspalo (approved):

PR #1264, opened 2022-11-18
Got rid dependency on GOOrganController from GOPipeConfig

oleg68:
1. Earlier the organ modification flag was in GODocument. Now this flag is split between GOOrganController and GOModel
2. This change allowed to replace dependency on GOOrganController with GOModel for yet another class: GOPipeConfig


larspalo (approved):
Works functionally, just a few comments that you can change or not.

oleg68:

rousseldenis (approved):

PR #1268, opened 2022-11-21
[ADD][TESTS] Add a testing framework to GrandOrgue / Github actions (testing, coverage)

rousseldenis:
This will add :

- A testing framework that allows to test the feature we develop, the bug we are fixing along with the development phase.
- Check with Codecov if the test we wrote cover the code we are eventually testing:

Codecov example (within this current PR):

image


rousseldenis:
@oleg68 As I'm quite noob in testing with c++ and CMake, could you give me clues on how to get a first test working ?

I have problems with imports. I think I have missing ones in my tests directory, but don't know how to do.

Many thanks

oleg68:
@rousseldenis Unfortunally I've never made tests in c++ with cmake, so I don't know how to help you.

rousseldenis:
@oleg68 First, it's just about CMake imports:

image

I think something is missing here:

image

larspalo:
From the errors it seems to me that you have an issue with including wxWidgets. Normally when I use wxWidgets in a cmake project I use both lines:
```
# change to whatever modules you need after REQUIRED
find_package(wxWidgets REQUIRED html net adv core base)
include(${wxWidgets_USE_FILE})

```
and remember to link with it. See https://cmake.org/cmake/help/latest/module/FindwxWidgets.html

CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ option(GO_USE_JACK	      "Use native Jack output" ON)
 if (WIN32 OR APPLE)
    option(INSTALL_DEPEND      "Copy dependencies (wxWidgets libraries and Translations) on installation" ON)
 else ()
-   option(INSTALL_DEPEND      "Copy dependencies (wxWidgets libraries and Translations) on installation" OFF)
+   option(INSTALL_DEPEND      "Copy dependencies (wxWidgets libraries and Translations) on installation" ON)
```
larspalo, line 69: Also, you shouldn't need to switch on copy dependencies when building on Linux...

rousseldenis: Yes, it was just a test.

But I finally succeeded in using :

`include(UsewxWidgets)`

rousseldenis: Now, I have a problem including core libraries to get usable classes in main test file

rousseldenis: Ok, I got it. I needed to use target_include_directories() to ... include header files.
So, now ready to dig into code :smiley:

rousseldenis:
I've improved tests structure with a successful run!

- Github full builds are now conditioned to succeeded tests (in a separate job).

The test framework still needs to be refined in order to built tests results and display them correctly.

I want to split tests in several directories but I still face CMakeList problems (due to my current limited understanding). If someone can guide me through this...

rousseldenis:
@oleg68 Logs show several warnings.

I've done a test with a simple configuration as done in perftests:

image

I don't know if they should be solved in code or refined tests biulding.

larspalo:
> I don't know if they should be solved in code or refined tests biulding.

Did you bother checking the actual value of the returned size of the ptr_vector? At least for the metronome, a separate windchest is added by GO, so the size could most likely not be expected to be only 1 after explicitly adding a new windchest (since there might be others already existing).

Also, the terminology should be changed - what you're actually testing is if it's possible to add another windchest and thus should compare the number of windchests before and after adding a new windchest if it's one more afterwards. The unsigned returned by the AddWindchest function is the number of all existing windchests.

rousseldenis:
@oleg68 @larspalo 

This one is working quite flawlessly.

It allows to drive development changes along with testing quite easily (I tried).

The idea was to provide a method 'GOAssert()' to a test class in order to:

- Evaluate the test expression, providing error message
- Throw a custom Exception GrandOrgue oriented
- Display an understandable stack trace (with cpptrace library) with pointer to the faulty code line.

The base test class was developed in order to create the necessary GOOrganController before writing the first test. This in order to simplify the needed test code lines... to test only what we want to test.

I've provided already some unit tests for simple methods 'class-oriented'. @oleg68 doing this, I did a fix (the only one in GrandOrgue base code) as it was not possible to create a GOOrganController without configuration file. The commit is here : https://github.com/GrandOrgue/grandorgue/pull/1268/commits/c5f9daa97012c1d72726e0e03fb224d9f5cb85db

I need some time now to improve documentation on how to doing new tests.

The only change in Github flow is that test step should succeed before the first build is launched. It should be safe as this is working on my branch that is up-to-date with grandorgue/master. You can look it here: https://github.com/rousseldenis/grandorgue/tree/add-tests-dro

I've activated Codecov on my side as code coverage is also included in this PR (but not mandatory as a first report is done in test steps).
 
IMHO, this could be merged in master after review.

rousseldenis:
Tests results are shown as this:

image

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -1213,7 +1213,8 @@ void GOSetter::SetTranspose(int value) {
 }
 
 void GOSetter::UpdateModified(bool modified) {
-  m_buttons[ID_SETTER_SAVE_SETTINGS]->Display(modified);
+  if (m_buttons.size() != 0)
```
oleg68, line 1216: Please move the refactoring in GOSetter and GOOrganController to a separate PR.

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+add_option(-mstackrealign)
+
+include(UsewxWidgets)
+include_directories(${CMAKE_BINARY_DIR}/src/core/go_defs.h ${CMAKE_SOURCE_DIR}/src/core)
+
+# Add here your tests files
+set(go_test_lib
+    # Test framework - Don't modify
+    GOTestResult.cpp
```
oleg68, line 14: Please distribute test classes for certain GO classes (ex GOTestSwitch) across subdirectories

oleg68 (changes requested):

src/tests/CMakeLists.txt
rousseldenis: Yes I know but don't know how to make it correctly through cmakelists

oleg68: > Yes I know but don't know how to make it correctly through cmakelists

``
set(go_test_lib
    dir1/GOTest1.cpp
    dir2/dir3/GOTest2.cpp
)
``

.github/workflows/build.yml
```diff
@@ -70,7 +96,7 @@ jobs:
         body_path: "${{ steps.relnotes.outputs.file }}"
         
   build:
-    needs: [calc_ver]
+    needs: [calc_ver, tests]
```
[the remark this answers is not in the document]
oleg68: > In certain cases yes, but here the aim is not to build if tests are failing.

I don't like this approach because compiling for testing is long.  Lets do build and tests in parallel. The dependency of Release on tests will be enough.

build-scripts/for-linux/build-for-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+# Define BUILD_TESTING=ON
+GO_PRMS="-DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON \
+  $CMAKE_VERSION_PRMS \
+  $($DIR/cmake-prm-yaml-cpp.bash $TARGET_ARCH)"
+
+# Build simply GrandOrgue without packaging
+echo "cmake -G \"Unix Makefiles\" $GO_PRMS . $SRC_DIR"
+cmake -G "Unix Makefiles" $GO_PRMS . $SRC_DIR
+echo "cmake --build . --parallel $PARALLEL_PRMS"
+cmake --build . --parallel $PARALLEL_PRMS
```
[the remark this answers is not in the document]
rousseldenis: @oleg68 In fact, there is no test run here. It's just building here.

The run is launched directly in github action there : https://github.com/GrandOrgue/grandorgue/pull/1268/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R24

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Install requirements for tests infrastructure
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y git gcovr
+
+# Install cpptrace from sources
+mkdir cpptrace
+cd cpptrace
+git clone https://github.com/jeremy-rifkin/cpptrace.git .
```
oleg68, line 9: May be unpack from release https://github.com/jeremy-rifkin/cpptrace/archive/refs/tags/v0.4.1.tar.gz instead of cloning from git?

.github/workflows/build.yml
```diff
@@ -2,6 +2,34 @@ name: Build
...
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      
+      - name: Install Base Dependencies
+        run: ${{github.workspace}}/build-scripts/for-linux/prepare-debian-based.sh
```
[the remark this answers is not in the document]
rousseldenis: You want to say add a 'test' parameter to the prepare-debian-based.sh to keep one script only ?

.github/workflows/build.yml
```diff
@@ -2,6 +2,34 @@ name: Build
...
+      - name: Install Test Dependencies
+        run: ${{github.workspace}}/build-scripts/for-linux/prepare-debian-for-tests.sh
+
+      - name: Build
+        working-directory: ${{github.workspace}}
+        run: ${{github.workspace}}/build-scripts/for-linux/build-for-tests.sh amd64
+  
+      - name: Test
```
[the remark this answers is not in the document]
rousseldenis: Locally, we maybe don't want to launch the coverage part but only the test one. The ctest command is quite clear to not include it in a shell script. No ?

build-scripts/for-linux/prepare-debian-for-tests.sh
rousseldenis: Yes, maybe to avoid installing git

rousseldenis: Done

.github/workflows/build.yml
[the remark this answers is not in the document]
oleg68: No, prepare-debian-for-tests.sh is good.

I want the warkflow would call prepare-debian-for-tests.sh, and prepare-debian-for-tests.sh would call prepare-debian-based.sh itself.

.github/workflows/build.yml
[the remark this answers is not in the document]
oleg68: I think the `do-tests.sh` script should do all these commands: 

```
ctest -T test --verbose
ctest -T coverage
gcovr -e "submodules/*|usr/*"
```

and the workflow should look like

```
      - name: Run tests
        working-directory: ${{github.workspace}}
        run: ${{github.workspace}}/build-scripts/for-linux/do-tests.sh
      - name: Display the code coverage result
        uses: codecov/codecov-action@v3
```

src/tests/testing/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(go_tests
+    # Add here your tests files
+    GOTestOrganModel.cpp
+    GOTestSwitch.cpp
+    GOTestWindchest.cpp
```
oleg68, line 5:
```suggestion
    model/GOTestOrganModel.cpp
    model/GOTestSwitch.cpp
    model/GOTestWindchest.cpp
```

The testing files should be distributed across subdirectories as like the source files are.

oleg68 (changes requested):

src/tests/testing/CMakeLists.txt
rousseldenis: > I think the do-tests.sh script should do all these commands:

Usually, it's better to not run the coverage if tests failed (that's the case in the github workflow), that's why I splitted the commands. That's also the case when developing (usually I don't care about coverage until push on git).

rousseldenis:
@larspalo @oleg68 What do you think of this version ?

I really think that further developments should take advantages of testing to avoid regressions.

oleg68:
@rousseldenis 

> What do you think of this version ?

It is better than the old one. You moved the testing class files to the appropriate directory, - it is good. 

The next thing I wait you to do is reworking the shell scripts. The objective is to make the workflow file as simple as possible and move all the logic to the shell scripts.

.github/workflows/build.yml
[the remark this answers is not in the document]
rousseldenis: @oleg68 I have gathered the commands in a shell script.

But I'd prefer keeping the different steps in github workflow as if tests are failing, it's nonsense to run coverage.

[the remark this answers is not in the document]
rousseldenis: This has been changed

rousseldenis:
> @rousseldenis
> 
> > What do you think of this version ?
> 
> It is better than the old one. You moved the testing class files to the appropriate directory, - it is good.
> 
> The next thing I wait you to do is reworking the shell scripts. The objective is to make the workflow file as simple as possible and move all the logic to the shell scripts.

@oleg68 Could you test the "test" flow on your side ? All issues have been solved. I think this should be merged in order to add tests during further developments in order to challenge it and improve it.

larspalo (changes requested):
You have comments in french, please translate them to english or remove them. Also you have code that's commented out. Please remove or rework them. Then, as food for thought, what do you actually hope to acomplish with the tests? My experience, with among other softwares - libsndfile, is that the test suite only brings much more extra maintenance work and headaches to the development... Prove me wrong.

rousseldenis:
> You have comments in french, please translate them to english or remove them. Also you have code that's commented out. Please remove or rework them. 

I will do it.

> Then, as food for thought, what do you actually hope to accomplish with the tests? My experience, with among other softwares - libsndfile, is that the test suite only brings much more extra maintenance work and headaches to the development... Prove me wrong.

My own experience (10 years programming with testing frameworks) is showing that including tests during developments helps:

- Better understanding of what you are trying to achieve as programmer (before digging into code as looking at tests will show you functionally and technically what a part of software is doing, during programming as I - as much as I can - write the test before writing one line of code).
- Avoiding regressions in other parts of software you are currently changing.

To be clear:

- Tests are also intended to help people (like me) to understand the code more easily
- Tests are in separate build (is not loaded for normal ones). We can disable them at any time (even if it is not wanted).

We can refine the test framework, but for now, implementing new ones is quite easy to do. Call just `this->GOAssert(<the assertion you want>)` in a class that inherits from `GOCommonControllerTest` and that's it:

image


I know from experience too that a lot of people do not like testing as they consider this is a waste of time because of the doubled time to write them. BUT, in fact, they will help you to detect warnings or errors you didn't anticipated, to write better quality code and to help debugging without going through the GUI (in fact, the most important advantage for me).

rousseldenis:
> You have comments in french, please translate them to english or remove them. Also you have code that's commented out. Please remove or rework them.

Done

larspalo:
> and to help debugging without going through the GUI (in fact, the most important advantage for me)

Right, I still sometimes fall back to debug using printing to stdout just to check functionality and variable contents in the actual flow that's not "pre-set" like the tests that you've written obviously are. Normally, the logging functionality would also be possible to use as well as running the program in a debugger to examine what's happening.

But, yes, I can absolutely agree with the fact that you'll learn a lot about the code when you're writing tests for it. However, with the way the tests currently are done, they should definitely be disabled completely by default in the builds and development and only performed if explicitly asked for (locally).

I definitely do not want to have a situation where there's a requirement that in order to implement a new feature (or change an existing) you also must write a test for it!

rousseldenis:
> However, with the way the tests currently are done, they should definitely be disabled completely by default in the builds and development and only performed if explicitly asked for (locally).

That's the current behavior 😁 

That has to be explicitly set. Which is done in the test github workflow job.

larspalo:
> Which is done in the test github workflow job

Which then contradicts what I just said.

rousseldenis:
> and only performed if explicitly asked for (locally).
In fact, this is always good to run it in workflow to be warned about failed test.

However, as we don't run tests in PRs, this does not really matters

rousseldenis:
> > and only performed if explicitly asked for (locally).
> > In fact, this is always good to run it in workflow to be warned about failed test.
> 
> However, as we don't run tests in PRs, this does not really matters

Even if blocking merges without green tests is a good practice

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
```
oleg68, line 4:
```suggestion
DIR=`dirname $0`
$DIR/../for-linux/prepare-debian-based.sh wx32 amd64
```

.github/workflows/build.yml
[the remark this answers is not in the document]
oleg68: Please remove these two lines.

.gitignore
```diff
@@ -19,9 +19,9 @@
 /ide-projects/NetBeans12/nbproject/private/
 
 *.a
+*.cmake
```
oleg68, line 22: It is not correct, because we have lots of *.cmake files in the cmake directory

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+# Install requirements for tests infrastructure
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gcovr
+
+# Install cpptrace from sources
+CPP_TRACE="0.4.1"
+CPP_TRACE_VERSION="v$CPP_TRACE"
+mkdir cpptrace
```
oleg68, line 11:
```suggestion
mkdir -p cpptrace
```

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+# Install requirements for tests infrastructure
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gcovr
+
+# Install cpptrace from sources
+CPP_TRACE="0.4.1"
```
oleg68, line 9: Please add the check if cpptrace has already been installed?

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+CPP_TRACE_VERSION="v$CPP_TRACE"
+mkdir cpptrace
+cd cpptrace
```
oleg68, line 12:
```suggestion
pushd cpptrace
```

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+wget https://github.com/jeremy-rifkin/cpptrace/archive/refs/tags/$CPP_TRACE_VERSION.tar.gz
+tar -xf $CPP_TRACE_VERSION.tar.gz
+cd cpptrace-$CPP_TRACE
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release
+make -j$PARALELL_PRMS
+sudo make install
+cd ../..
```
oleg68, line 21: popd

oleg68 (changes requested):

oleg68:
> Then, as food for thought, what do you actually hope to acomplish with the tests?

It is useful to have capability to run some testcases. Especially when I was making the new resampling I lacked this capability.

And the second reason for writing tests is regression tests: whether a change breaks some functionality?

larspalo:
> It is useful to have capability to run some testcases.

Might be occasionally, but it heavily depends on if it shows/tests the correct things. But as a general part of the regular workflow? Not in my opinion. I should only be done when specifically invoked locally == not on Github. I think that any tests, if applicable, should have been done before the code arrived there.

oleg68:
Yes, your are right. Running tests on github is more useful for regression tests.

rousseldenis:
> Yes, your are right. Running tests on github is more useful for regression tests.

I would say, by experience and programmatically, every test should be run independently of where and when. That's the principle behind testing (unit testing).

Every test should produce the same result.

rousseldenis:
> I think that any tests, if applicable, should have been done before the code arrived there.

In a perfect world. But, running also on github ensure nothing has been forgotten. That's the difference between manual testing and automated test flows.

larspalo (changes requested):
I'll only approve of the tests if they are only optionally run locally. They should in no way be run automatically on Github as a judgement of code in any PR submitted to the repository. They can be built as a separate app and included in the releases at will, but never (as they currently are done) on action builds.

rousseldenis:
To summarize: what are you afraid of ? :sweat_smile:

oleg68:
@larspalo  

> I'll only approve of the tests if they are only optionally run locally. They should in no way be run automatically on Github as a judgement of code in any PR submitted to the repository. 

It contradicts the most common practice. The must important reason is to find regression errors **if the autor forgot to run run tests locally**.

As a compromice I'd suggest to run the tests on github at every push in a separate task that we could see their results but they would not affect the build outcome.

.github/workflows/build.yml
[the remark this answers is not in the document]
oleg68: You still call `prepare-debian-based.sh` from the workflow directly. It is not necessary because it is also called from  `prepare-debian-for-tests.sh`

larspalo:
> To summarize: what are you afraid of ? 😅

@rousseldenis Maintenance burden.

The internal API of GO is not in any way finalized. Nor are the naming convention of any methods that you're using for the tests absolutely "fixed in stone". A simple change in any get or set method of any "unit" would break your tests and make them fail!

I don't want the agile development model of GO to be limited by the tests you're doing here. If they would be enforced at code checking for merges into master from any PR, they will encumber every other developer with maintenance burden of your test suite (== needing to change also in the tests any naming of simple methods used for "testing", or any other behaviours for that matter). This is not in any way acceptable to me as it won't benefit further development. It only serves to "freeze" exactly what is already existing.

If any further feature development is to be encouraged (which I hope it will be!) there shouldn't be any extra burden on the developers doing it. As a side note, I have personal experience with this exact situation, and I sincerely hope to spare anyone else the same experiences!

Tests are fine for local "testing", but as the tests you've implemented only would burden other developers with updating also the test suite whenever changing anything (even a simple Get or Set naming), the tests are not that useful in my opinion if applied as a code quality judgement and condition for anything included in the code.

The old perftest on the other hand is perfectly fine and I think that any other tests added to GO should follow that model instead.

build-scripts/for-linux/prepare-debian-for-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+set -e
+
+DIR=`dirname $0`
+$DIR/../for-linux/prepare-debian-based.sh wx32 amd64
+
+# Install requirements for tests infrastructure
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y gcovr
+
+# Install cpptrace from sources
+if ! ls /usr/local/lib | grep libcpptrace &> /dev/null
```
oleg68, line 12:
```suggestion
if ! ldconfig -p | grep libcpptrace &> /dev/null
```

because the library might be installed in another dir

oleg68 (changes requested):

rousseldenis:
> @rousseldenis Maintenance burden.

OK. I think this is because you don't practice programming that begins by test writing (and then coding). I admit I was used to never write tests because of the reasons you mentioned before I really take the problem from the opposite side and the benefits gained.

> The internal API of GO is not in any way finalized. Nor are the naming convention of any methods that you're using for the tests absolutely "fixed in stone". A simple change in any get or set method of any "unit" would break your tests and make them fail!

FYI, current PR's tests were written one year ago, didn't changed and still green. But of course, any change that touches the test frontier would change the test result, but that's their purpose!

> I don't want the agile development model of GO to be limited by the tests you're doing here.

This is completely in phase with agile development. Early testing (unit and functional) is highly encouraged. E.g.: https://www.opentext.com/what-is/agile-testing

And, to be clear, at the end, if this is accepted, this will become 'our test flow' 😅.

> As a side note, I have personal experience with this exact situation, and I sincerely hope to spare anyone else the same experiences!

I understand you had a bad experience with tests, but let me maybe change your mind. Of course it will surcharge the dev but for a better code. Again, IMHO, you should along with code writing, write the tests (or change existing ones). That should goes naturally one with the other.

> if applied as a code quality judgement and condition for anything included in the code.

It's not about judgment. It's about better writing. On a more 'philosophical' point of view.

> The old perftest on the other hand is perfectly fine and I think that any other tests added to GO should follow that model instead

Those are more integration tests and don't have the same aim. Moreover, they don't cover the whole code and are more a little clue of the whole software is.

Nevertheless, be sure this is a contribution I've made in order to get a better software and evolving towards standards. No gain with this. I'm tough a GrandOrgue newbie. So, I will close this if it does not encounters majority.

@oleg68

build-scripts/for-linux/prepare-debian-for-tests.sh
rousseldenis: @oleg68 I've tested ldconfig, but it does not produce the result expected. I will try once more time

oleg68: > @oleg68 I've tested ldconfig, but it does not produce the result expected. I will try once more time

Could you give more information. What case ldconfig does provide incorrect result in?

rousseldenis: I'm not in front of my computer, but the output does not seem to find the lib.

oleg68:
@larspalo 

What is you opinion on my suggestion:
- To run tests on github on every push
- Not to require the tests would be passed for merge?

larspalo:
> To run tests on github on every push

No, please.

> Not to require the tests would be passed for merge?

Yes, if done as a separate app like GrandOrguePerftest and GrandOrgueTool, supplied together with the builds is perfectly fine for me and something I can approve of.

rousseldenis:
> > To run tests on github on every push
> 
> No, please.

I'm still questioning on this.

Even tests are red, PR can be merged. Then, a good will person can fix the tests in a further PR. This still does not block developments.

oleg68:
> Yes, if done as a separate app like GrandOrguePerftest and GrandOrgueTool, supplied together with the builds is perfectly fine for me and something I can approve of.

But contributors will forget to run tests manually. Automatic testing on push would compensate it.

What are your contras?

rousseldenis:
> But contributors will forget to run tests manually. Automatic testing on push would compensate it.

I think I understand what @larspalo wants.

Only a separate tool to test locally at will. And not to show red tests on github side (even on fork side).

@oleg68 I think you understand the interest of this running also on github side and be part of development flow.

larspalo:
> What are your contras?

Again, mostly it's burden of maintenance, secondly a question of test validity.

Failed tests might currently just indicate a change (whatever that change might be, to accessors or just any other re-naming really) and not at all that something in the actual code is wrong in the real performance - which is the most important thing.

I fear that enforced tests (without really good reasons) with big red flags just will deter progress (slow it down as it will require more maintenance work to fix "issues" that in reality doesn't exist, it's just faulty tests as of now) and that ultimately it could reduce the will of providing any contributions even more.

I also think that the whole focus of the testing is currently somewhat misplaced. Testing the internal workings of "units", _in the way it's suggested/done right now_ - which is actually just testing individual getters/setters, is fairly un-interesting compared to even a basic overall test of how a loading of a "canonical" ODF would behave against expectations - the tests could of course still be broken down into "units" - in order to know if something important would break by a certain code change. (Integration stuff like the testing of more complex subjects like inheritance of values between nodes and calculations of total/effective values among other things would also be highly beneficial in such contexts)

So, yeah, I'd definitely like integration tests as well as, or even more than, unit tests - and the unit tests should be much better designed than just testing individual getters/setters, in the best of worlds both types should most likely be combined.

In the end I don't think that these particular tests will "improve" the quality of any code at all (on any level, "philosophical" or not), it just means writing more code... If you want to do it, it's very fine with me as it most certainly will develop you personally, but please don't force it on everyone else.

rousseldenis:
@larspalo my first thoughts yesterday was to give up...

But, as I want to contribute more and as my development flow is test-driven one, I want this to be merged.

If I remove the github workflow part, will you approve it?

> In the end I don't think that these particular tests will "improve" the quality of any code at all (on any level, "philosophical" or not), it just means writing more code..

The first tests I've done were there for improving the test framework itself and have in fact no real value added. The final aim is to write tests of 'flows' of course.

I would say many libraries and projects do those automated tests and developers are testing those kind of units. As a function has a 'contract' (input and output) and that contract should always be respected (like an API) even if the inner implementation changes.

But nevertheless, let's time pass and maybe one day 😅

rousseldenis:
@larspalo @oleg68 

I've updated this in order to allow manual triggering the test workflow.

After merge, in 'Actions' menu of Github (in local repos but also here) will allow to run manually the tests:

image


I think this is the best compromise between a full automated test workflow and a local one.

That allows devs to to run the tests here to in addition of their local machine and do not bother use with failed jobs per default

build-scripts/for-linux/prepare-debian-for-tests.sh
rousseldenis: @oleg68 I don't know why but lpptrace is not in the list with ldconfig

rousseldenis: I don't know if that helps but the file is named `libcpptrace.a`

oleg68: It helps. ldconfig searches only .so files, not .a ones.

larspalo:
> after making some breaking changes (ex. renaming) the developer should also update tests as like he/she updates another code that relies on the name.

And that's exactly what I hope to avoid at (more or less) any cost, as it without any doubt just increase the maintenance burden...

> But in the weak workflow it is not mandatary.

That is the only way that I'll approve of. The tests must be allowed to fail as much as they possibly can when someone actually wants to trigger them (manually), but if a build of the GO code work - the build of the code should always be allowed to continue. The tests might very well be built together with other code - but if the build of the tests themselves fail for any reason it must not prevent the build of the actual GO code to proceed and succeed (if it can). If this is guaranteed, it's good enough for my approval.

oleg68:
> That is the only way that I'll approve of. The tests must be allowed to fail as much as they possibly can when someone actually wants to trigger them (manually), but if a build of the GO code work - the build of the code should always be allowed to continue.

Do you agree with the following approach: the tests are triggered automatically on github as well as manually, but the build continues even if the tests fail? So failed tests give more information (that is always good), but do not prevent building (what you want to avoid).

And, of cause, they may be run locally with the shell script.

larspalo:
> Do you agree with the following approach: the tests are triggered automatically on github as well as manually, but the build continues even if the tests fail?

@oleg68 Yes, but not only that. As long as the tests, failing or not, are not preventing _anything_ else (building, merging, releases etc.) and are considered **optional information** only. Failing tests should not be required to be "fixed" by any developer if broken by functional code - and **in that manner we explicitly remove any enforced extra maintenance of them**. On such terms, yes I'll approve, as this seems to be so terribly important to you.

oleg68:
@larspalo thank you

@rousseldenis could you restore running the testing job on push but remove the dependency of release on test?

rousseldenis:
> @rousseldenis could you restore running the testing job on push but remove the dependency of release on test?

Yes I'll do it

rousseldenis:
@oleg68 @larspalo I think this is ready

.github/workflows/build.yml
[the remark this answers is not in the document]
oleg68: You haven't still removed the lines 13 and 14

rousseldenis:
@oleg68 Done

oleg68 (approved):

src/tests/common/GOTestCollection.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "GOTestCollection.h"
+#include "GOTestException.h"
+#include "GOTestResultCollection.h"
+#include <iostream>
+#include <iterator>
+#include <vector>
+
+GOTestCollection::GOTestCollection() {}
+
+// GOTestCollection::~GOTestCollection() {
```
larspalo, line 16: It's usually bad practice to leave code thats commented out. Consider removing this if it's not going to be used in the very close future.

src/tests/common/GOTestResult.h
```diff
@@ -0,0 +1,36 @@
...
+  wxString m_result_message;
+  bool failed;
+
+public:
+  GOTestResult();
+  GOTestResult(wxString message);
+  GOTestResult(wxString message, bool failed);
+
+  const wxString GetMessage() {
+    wxString result = "";
```
larspalo, line 29: Why write things in this way to return the message?

src/tests/testing/model/GOTestOrganModel.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "config/GOConfig.h"
+#include "model/GOWindchest.h"
+
+GOTestOrganModel::~GOTestOrganModel() {}
+
+void GOTestOrganModel::run() {
+  wxString message;
+  // Set OrganModel Modified
+  this->controller->SetOrganModelModified(true);
+  message = "Is Organ Modified value is not True";
```
larspalo, line 23: Do note that all such kinds of wxString assignments will be at the mercy of the current locale encoding. Generally it of course should be safe with pure English characters, but using the appropriate macros is still recommended.

src/tests/testing/model/GOTestSwitch.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "model/GOSwitch.h"
+
+GOTestSwitch::~GOTestSwitch() {}
+
+void GOTestSwitch::run() {
+  wxString message;
+
+  // Check global Switch
+  GOSwitch *go_switch = new GOSwitch(*this->controller);
+  message = "The associated manual should be -1 as it is a global switch!";
```
larspalo, line 17: Ditto the comments above.

src/tests/testing/model/GOTestWindchest.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "model/GOWindchest.h"
+
+GOTestWindchest::~GOTestWindchest() {}
+
+void GOTestWindchest::run() {
+  wxString message;
+  // Check the size of Windchest is correct
+  unsigned w_size;
+  w_size = this->controller->AddWindchest(new GOWindchest(*this->controller));
+  message = "Windchest size is wrong";
```
larspalo, line 28: Ditto comments above.

larspalo (changes requested):

src/tests/common/GOTestCollection.cpp
rousseldenis: Of course, forgot it.

src/tests/common/GOTestResult.h
rousseldenis: Why not ?

rousseldenis: Oh yes. Was testing something appending strings to result.

I think I can simply return the variable.

rousseldenis:
@larspalo Done. Thanks reviewing the tests :smiley:

src/tests/testing/model/GOTestWindchest.cpp
oleg68: But you haven't change the code. Lars expected that you would write

``message = wxT("Windchest size is wrong")``

oleg68 (changes requested):

src/tests/testing/model/GOTestWindchest.cpp
oleg68: The same pattern in all other places. You should use the `wxT` macro instead of assignment char* to wxSrtring

rousseldenis: Mmmh, ok.

This code was written at the beginning of this. IMHO, we shouldn't depends on wx in tests too apart for things related to dialogs. I think changing strings representations to std::string is the good approach.

rousseldenis: @oleg68 @larspalo I've changed the manner result message is defined.

src/tests/common/GOTest.h
```diff
@@ -0,0 +1,44 @@
...
+ * License GPL-2.0 or later
+ * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
+ */
+
+#ifndef GOTEST_H
+#define GOTEST_H
+
+#include "GOOrganController.h"
+#include "GOTestUtils.h"
+#include <wx/string.h>
```
larspalo, line 12: Is it needed?

src/tests/GOTestExe.cpp
```diff
@@ -0,0 +1,52 @@
...
+  // Display tests results
+  int run_number_ = 0;
+  std::vector<GOTestResult *> test_results
+    = test_result_collection.get_results();
+  wxLogMessage("==================== TESTS RESULTS ====================");
+  for (auto current = test_results.begin(); current != test_results.end();
+       ++current, ++run_number_) {
+    auto test_result = *current;
+    wxLogMessage("-------------------------------------------------------");
+    wxLogMessage(wxString(test_result->GetMessage()));
```
larspalo, line 39: If all the wxLogMessages would be rewritten with std::cout you'll not need any wxString here either.

larspalo (changes requested):
If you move towards using std::string instead of wxString I think that you can also clean up lots of the includes as well as starting to use std::cout instead of wxLogMessage for the output. There are many similar places that I've not pointed out but it's just the same...

oleg68 (approved):

larspalo (approved):

PR #1292, opened 2022-11-30
Moved checking of HW1-compatibility filenames from GOLoaderFilename::Asign to the new method GOConfigReader::ReadFileName

oleg68:
Toward to removing dependency on GOOrganController I removed checking filenames for containing `/` from `GOLoaderFilename::Assign`.

The checking is now when this filename is read from the ODF instead of using this filename.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1293, opened 2022-11-30
Extracted GO_WAVE_LOOP from GOWave to a separate file and renamed to GOWaveLoop

oleg68:
There are a lot of doubled structure definitions of the same `Loop` structre in the source code. I'm going to use the single `GOWaveLoop` structure everywhere. This is PR extracts it from the `GOWave.h` file

No GO behaviour should be changed.


larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1249, opened 2022-11-10
Unhandled exception with r3.9.0.1


paolo2504:
hi.
Latest release 3.9.0.1 Win64 crashes on loading some , not all, organs returning a fatal error (screenshot below attached).
Previous r3.8.0.1 (as well other prior) does not show the error with those organs.

Note: organ packages that show error are installed properly. Error arise only on load. This happens with most of my organ packages but not all. By interrupting execution GO crashes and closes. Doing "Ignore", loading is canceled and GO remains open. I reverted to r3.8.0.1.

![unhandled exception](https://user-images.githubusercontent.com/50132661/201179079-64e4b1bd-156f-4e44-a03b-97e5f2b18879.jpg)

oleg68:
@paolo2504 could you attach here the crash dump file?

paolo2504:
[CrashDumps.zip](https://github.com/GrandOrgue/grandorgue/files/9991161/CrashDumps.zip)

3 dumps with different package organs...

oleg68:
@paolo2504 Does it always craches on the same packages? Or the packages sometimes are loaded, sometimes - are not?

If the problem is 100% reproducible with a certain package, could you place the smallest package somewhere in the file sharing and send me the link to it?

paolo2504:
Always with same packages. Even tested with previous 3.7.1 no problem. 
packages not showing issue (Friesach for example) are loaded correctly (even if apparently slower).
I’ll post the smallest package I have soon (Carillon beiaard).

paolo2504:
HEre below the smallest orgue package i have. It contains two organs, bith causing crash.
I noted that the exception dialog is shown twice i mean: i clock on abort first, then dialog is shown again, then i click second time on abort, dialog closes and after a second GO closes itself.

 
[CrashDumps.zip](https://github.com/GrandOrgue/grandorgue/files/9995102/CrashDumps.zip)
[beiaard.zip](https://github.com/GrandOrgue/grandorgue/files/9995105/beiaard.zip)

oleg68:
@paolo2504 Thank you, I managed to reproduce this issue.

oleg68:
This problem occures on any organ without a pedal.

paolo2504:
Interesting!

oleg68:
@paolo2504 Could you try [the intermediate build](https://github.com/oleg68/GrandOrgue-official/actions/runs/3452408113)? Does it solves the issue?

[...]

PR #1294, opened 2022-12-01
Fixed crashing on exit in some cases

oleg68:
In some situations, for example, if an unhandled exception occured (#1249), GrandOrgue exits and the exit causes a segfault.

The reason is `GOApp::OnExit` is called from wx when some windows, included `GOFrame` have not yet destroyed and continue to use objects `GOSound` and `GOConfig` that are deleted.

This PR forces destroying all windows before deleting the `GOSound` and `GOConfig`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed crashing on exit in some cases
 # 3.9.2 (2022-11-29)
 - Fixed sending incorrect midi values from an enclosure https://github.com/GrandOrgue/grandorgue/issues/1267
 - Fixed not sending midi events from pushbuttons, including GC https://github.com/GrandOrgue/grandorgue/issues/1220
--- a/src/grandorgue/GOApp.cpp
+++ b/src/grandorgue/GOApp.cpp
@@ -155,12 +155,6 @@ int GOApp::OnRun() { return wxApp::OnRun(); }
 int GOApp::OnExit() {
   wxLog::SetActiveTarget(NULL);
 
-  wxApp::CleanUp(); // delete all GUI elements in order not to use m_soundSystem
-
-  delete m_soundSystem;
-  delete m_config;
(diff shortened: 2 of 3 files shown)


larspalo:
I've not had GO crash on exit recently (on Linux of course...) but with this PR I'm getting a warning at exit similar to:

`(GrandOrgue:3500): GLib-CRITICAL **: 21:25:34.004: Source ID 6590 was not found when attempting to remove it`

Looking at the code I'm wondering if the wxApp::CleanUp() shouldn't be done after wxApp::OnExit()? I understand that the call is only made to get the return value, but perhaps the order matter anyway...

oleg68:
> I've not had GO crash on exit recently (on Linux of course...)

I reproduce this problem with adding ``throw wxT("Test exception");`` to the end of `GOOrganController::Load` before the `return`. Of cause, this exception is not a desired behavior, but even such exception occurs somewhere, GrandOrgue should not crash.



> but with this PR I'm getting a warning at exit similar to:
> 
> `(GrandOrgue:3500): GLib-CRITICAL **: 21:25:34.004: Source ID 6590 was not found when attempting to remove it`
> 
> Looking at the code I'm wondering if the wxApp::CleanUp() shouldn't be done after wxApp::OnExit()?

Yes, you are right. Normally `wxApp::CleanUp()`is called after `wxApp::OnExit()`, and with this PR `wxApp::CleanUp()` starts being called twice that causes the gtk warning.

oleg68:
@larspalo I've reworked this PR. Instead of calling wxApp::Cleanup() from GOApp::OnExit(), I overrided wxApp::Cleanup() with GOApp:Cleanup() and moved the deletion code here from GOApp::OnExit() after calling wxApp::Cleanup()

larspalo (approved):

rousseldenis (approved):

PR #1301, opened 2022-12-05
Extracted the file-finding logic from GOOrganController to the new class GOFileStore

oleg68:
Toward to simplifying GOOrganController I  extracted some code responsible for searching files and archives from the class to the new class `GOFileStore` in the `loader` subdirectory.

### DIFF
--- a/src/grandorgue/GOOrganController.h
+++ b/src/grandorgue/GOOrganController.h
@@ -19,6 +19,7 @@
 #include "control/GOEventDistributor.h"
 #include "control/GOLabelControl.h"
 #include "gui/GOGUIMouseState.h"
+#include "loader/GOFileStore.h"
 #include "model/GOModificationListener.h"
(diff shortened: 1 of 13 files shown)


rousseldenis (approved):
LGTM

src/grandorgue/GOOrganController.h
```diff
@@ -151,6 +147,16 @@ class GOOrganController : public GOEventDistributor,
   // Clears the organ modification flag
   void ResetOrganModified();
 
+  const GOFileStore &GetFileStore() const { return m_FileStore; }
+
+  /**
+   * Set organ directory without lroviding any odf. Called only from GOPerfTest
```
larspalo, line 153: * Set organ directory without providing any odf. Called only from GOPerfTest

larspalo (approved):

PR #1320, opened 2022-12-20
Refactor/rename gofile

oleg68:
1. Renamed `GOFile` to `GOOpenedFile` (in my opinion, `GOOpenFile` would be worse because `Opened` means the result of GOLoaderFilename::Open instead the `Open` state) and moved to the `src/core/files` subdirectory.
2. Moved another file-related classes: `GOInvalidFile` and `GOStandardFile` to the `src/core/files` subdirectory.

It is just moving and renaming. No GrandOrgue behavior should be changed.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1322, opened 2022-12-24
Help: fixed errors, typos and some inconsistencies

willeke1234:
- Fixed errors
- Ran spell checker
- Fixed some inconsistencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Help: fixed errors, did spell check and some cleanup
+- Fixed typos and spelling errors in the Help
 - Fixed continuing loading an organ after an exception in one loading thread
 - Fixed size of text fields in the Organ Settings dialog on OsX https://github.com/GrandOrgue/grandorgue/issues/1315
 - Fixed missing the object filename in an error message if some exception occured when loading this object
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Help errata, spell check and some cleanup
 - Fixed continuing loading an organ after an exception in one loading thread
 - Fixed size of text fields in the Organ Settings dialog on OsX https://github.com/GrandOrgue/grandorgue/issues/1315
 - Fixed missing the object filename in an error message if some exception occured when loading this object
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Help errata, spell check and some cleanup
+- Help: fixed errors, did spell check and some cleanup
 - Fixed continuing loading an organ after an exception in one loading thread
 - Fixed size of text fields in the Organ Settings dialog on OsX https://github.com/GrandOrgue/grandorgue/issues/1315
 - Fixed missing the object filename in an error message if some exception occured when loading this object
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -7406,7 +7406,7 @@ Has the following attributes:
           <term>DispShortcutKeyLabelFont</term>
           <listitem>
             <para>
-(string, required) Name of the font for shortcut labels
+(string, required) Name of the font for shortcut labels - provided just for HW1 compatibility. DEPRECATED.
      </para>
           </listitem>
         </varlistentry>
@@ -7414,7 +7414,7 @@ Has the following attributes:
           <term>DispShortcutKeyLabelColour</term>
           <listitem>
             <para>
-(color, required) Color for shortcut labels
+(color, required) Color for shortcut labels - provided just for HW1 compatibility. DEPRECATED.
      </para>
           </listitem>
         </varlistentry>
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -1,4 +1,4 @@
-<?xml version='1.0' encoding='iso-8859-1'?>
+<?xml version='1.0' encoding='ISO-8859-1'?>
 <!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.2//EN" "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd">
 <book lang="en">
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Help: fixed errors, did spell check and some cleanup
```
oleg68, line 1:
```suggestion
- Fixed typos and spelling errors in the Help
```

oleg68 (approved):

help/grandorgue.xml
```diff
@@ -6352,7 +6352,7 @@ only allows adjustments at pipe level.
       </para>
       <para>
 The samples in the format above may also be packed with wavpack
-http://www.wavpack.com/.  GO only supports the wavpack V4 format
+http://www.wavpack.com/. GO only supports the WavPack V4 format
```
larspalo, line 6355: Actually, I've tested with the stock WavPack of Ubuntu 20.04 - which is 5.2.0 - and it works just fine so this line should perhaps be changed...

help/grandorgue.xml
```diff
@@ -7406,7 +7406,7 @@ Has the following attributes:
           <term>DispShortcutKeyLabelFont</term>
           <listitem>
             <para>
-(string, required) Name of the font for keyboard labels
+(string, required) Name of the font for shortcut labels
```
larspalo, line 7409: I don't know if it should be mentioned in the help or not, but I don't think that GO use this feature at all (and as far as I recall never has) since the keyboard shortcuts aren't displayed next to any items that have them set. Perhaps it wasn't implemented in the displaymetrics, but I think that it dates all the way back to the MyOrgan time (which didn't display them either as I remember).

larspalo (approved):
Mostly excellent changes suggested. Some of the title corrections make me wonder if we'll eventually need to review all the titles in GO for consistency of word capitalization... The changes related to distinctive writing so typical of the English language will of course not necessarily translate to other languages though, as that way of writing often lead to direct language faults or possible misunderstandings in other languages.

help/grandorgue.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version='1.0' encoding='ISO-8859-1'?>
+<?xml version='1.0' encoding='iso-8859-1'?>
```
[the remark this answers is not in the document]
willeke1234: Or move to UTF-8? It took me several attempts to get Piteå instead of PiteÃ¥ (on the Panels page).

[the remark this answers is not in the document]
rousseldenis: Il tend to say utf-8 is the common standard, so I would privilege that approach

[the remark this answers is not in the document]
larspalo: Don't really know if it matters, but even though GO states ISO-8859-1 as canonical encoding it has long supported UTF-8 (according to the help: if it starts with an approriate byte order marker) as a valid alternative. Compared to the value of the rest of the corrections in this PR I'd say that changing the encoding can wait.

help/grandorgue.xml
larspalo: But this feature actually existed in HW1, which is why it still exist in the ODF as it's a valid entry in such sample sets.

help/grandorgue.xml
larspalo: It seems that according to https://www.wavpack.com/WavPack5PortingGuide.pdf the version 5 is compatible with version 4 so that line could be changed to: 
`http://www.wavpack.com/. GO supports both WavPack 4 and WavPack 5.`

help/grandorgue.xml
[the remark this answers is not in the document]
willeke1234: This is the encoding of the file grandorgue.xml and has nothing to do with other files. Is PiteÃ¥ a macOS only issue?

help/grandorgue.xml
willeke1234: I'm not familiar with WavPack. The text is "GO only supports the wavpack V4 format without hybrid compression and it must include all metadata. A good starting point for the wavpack compression options is -x6". Should I change it to "GO supports both the WavPack V4 and WavPack V5 formats, without hybrid compression and it must include all metadata. A good starting point for the wavpack compression options is -x6"? Or "GO only supports the wavpack V4 and V5 formats, without hybrid compression and it must include all metadata. A good starting point for the wavpack compression options is -x6"? Or something else?

help/grandorgue.xml
willeke1234: I'll add "provided just for HW1 compatibility. DEPRECATED." like `StopControlMIDIKeyNumber` and `MIDIProgramChangeNumber`.

help/grandorgue.xml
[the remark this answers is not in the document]
larspalo: @willeke1234 I cannot say anything about macOS, on Linux I think I had to add the ISO-8859-1 encoding explicitly in the text editor. Before that the characters å, ä and ö would indeed look wrong. If it helps others I'm perfectly fine with changing encoding to UTF-8 as long as all non-English characters are represented correctly in GO.

help/grandorgue.xml
larspalo: @willeke1234 WavPack doesn't use V4 or V5 to designate versions. Therefore I think that just the relevant part about the supported WavPack formats should be changed. The whole section in the help could then read:

`The samples in the format above may also be packed with WavPack http://www.wavpack.com/. GO supports both the WavPack 4 and WavPack 5 formats, without hybrid compression and it must include all metadata. A good starting point for the WavPack compression options is -x6.`

As noted in the document I linked to above, the 4 version is completely built into the 5 version and changes are only done on top of it (thus making it possible to use the WavPack 5 version exactly the same way as the 4 version was).

help/grandorgue.xml
larspalo: @willeke1234 That's fine with me.

PR #1329, opened 2022-12-31
Restored capability of logging all file-not-existing errors while loading samples

oleg68:
Before #1319 if sveral sample files had not existed, the full list of not existed files was logged in the error message, but after this change no filename appears in the logs. It makes difficult to work with ODFs with wrong filenames (for example, https://github.com/GrandOrgue/ODFEdit/discussions/7#discussioncomment-4559870).

Now I restored logging all absent filenames as it was before.

This change does not start logging absent 'bmp' files, because they also were not logged before.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1330, opened 2023-01-03
Moved IgnorePitch from GOOrganController to GOPipeConfig

oleg68:
Toward to eliminating dependency from GOOrganModel on GOOrganController, I moved `m_IgnorePitch` from `GOOrganController` to `GOPipeConfig`. I'm planning to make this setting overridable at the winchest/rank/pipe level in the future.

I introduced the `BooleanTripple` with three values: 0 - false, 1 - true and -1 - inherit from the parent. I added new methods  for reading and writing BooleanTripple values to the config file.

NO GrandOrgue behavior should be changed.

### DIFF
--- a/src/core/config/GOConfigReader.cpp
+++ b/src/core/config/GOConfigReader.cpp
@@ -139,15 +139,15 @@ bool GOConfigReader::ReadBoolean(
   return ReadBoolean(type, group, key, required, false);
 }
 
-bool GOConfigReader::ReadBoolean(
+int GOConfigReader::ReadBooleanTriple(
   GOSettingType type,
-  wxString group,
-  wxString key,
-  bool required,
-  bool defaultValue) {
+  const wxString &group,
+  const wxString &key,
+  bool required) {
   wxString value;
+
   if (!Read(type, group, key, required, value))
-    return defaultValue;
+    return -1;
 
   if (value.length() > 0 && value[value.length() - 1] == ' ') {
     if (m_Strict)
@@ -159,9 +159,9 @@ bool GOConfigReader::ReadBoolean(
     value.Trim();
   }
   if (value == wxT("Y") || value == wxT("y"))
-    return true;
+    return 1;
   if (value == wxT("N") || value == wxT("n"))
-    return false;
+    return 0;
   value.MakeUpper();
   wxLogWarning(
     _("Strange boolean value for section '%s' entry '%s': %s"),
@@ -169,9 +169,9 @@ bool GOConfigReader::ReadBoolean(
     key.c_str(),
     value.c_str());
   if (value.Length() && value[0] == wxT('Y'))
-    return true;
+    return 1;
   else if (value.Length() && value[0] == wxT('N'))
-    return false;
+    return 0;
 
   wxString error;
   error.Printf(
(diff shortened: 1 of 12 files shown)


rousseldenis (approved):

larspalo:
@oleg68 The ignore pitch info in sample files is a user setting only available through the GUI. The single purpose of this is to be able to use temperaments in older samplesets where no pitch info was embedded into the samples by the sample set creator. It's not supposed to be an ODF setting. One could simply consider this a legacy setting to increase the usability of older material - it should not be encouraged for anything new. So, better keep this out from the ODF!

In the ODF on the other hand there is another setting: "AcceptsRetuning" that have the purpose of telling GO that this sample is not intended to be retuned at all (it's mainly intended for usage on effects - noises, cymbelstern and similar stuff). This is the stuff that a sample set designer is supposed to use (and of course provide the necessary pitch info into the samples that indeed are supposed to be re-tuned).

I bet you're already aware of it, but the combination of -1, 0 and 1 as values from the ODF is one thing, -1 could also be indicating a default value or infinite duration.

oleg68:
> So, better keep this out from the ODF!

This PR keeps allowing this setting only at the preset-level instead of ODF. 
```
  m_IgnorePitch = cfg.ReadBooleanTriple(
    CMBSetting, m_Group, m_NamePrefix + wxT("IgnorePitch"), false);
```

Earlier the similar code was in GOOrganController::ReadOrganFile:

```
  m_IgnorePitch
    = cfg.ReadBoolean(CMBSetting, group, wxT("IgnorePitch"), false, false);
```

My primary intention is to make the code simpler by implementing the idea "Everething adjustable with the OrganSettingDialog is encapsulated in `GOPipeConfig`". GOSoundingPipe should take all properties that are necessary for producing the sound from it's GOPipeConfig instance.

This idea encourages me to move various ajustable stuff from GOOrganController to GOPipeConfig. It allows to eliminate unnecessary dependencies on GOOrganController everywhere.

This PR moves only `IgnorePitch`. The next one will make the same with `ReleaseTail`.

> but the combination of -1, 0 and 1 as values from the ODF is one thing, -1 could also be indicating a default value or infinite duration.

The format of the value in the setting file is kept the same: `Y` or `N`. But because GOPipeConfig has an inheritable nature, it requires distinction between the value set at this level and one inherited from the parent. So I introduced `BooleanTriple`. May be `int` is not the best choice for this abstraction, and enum would be better. If you insist on it, I agree to change the datatype.

For now, all PipeConfig::IgnorePitch may have only -1 except the root one that may be 0 or 1. But in the future we can make this property adjustable at Windchest/Rank/Pipe level (it may be useful for composite organs).

The PipeConfig abstract model gives us a lot of capabilities whether we use them or not now. But I always prefer to reuse a few existing abstract entities rather to use a lot of specialised ones according to the well known Occam's principle. Moving fields from `GOOrganController` to `GOPipeConfig` decreases the number of entities.

larspalo:
@oleg68 Sure, I just point out some of the more subtle things to keep in mind to ensure that we keep steering in the right direction.

`For now, all PipeConfig::IgnorePitch may have only -1 except the root one that may be 0 or 1. But in the future we can make this property adjustable at Windchest/Rank/Pipe level (it may be useful for composite organs).`

If this is indeed implemented, it should be done as how I insist on the reverb tail shortening. Not through ODF at all, but done like in the organsettings kind of voicing manner through user GUI. For the ODF and sampleset creators the AcceptsRetuning is already available both for rank and individual pipes.

larspalo (approved):

Issue #1333, opened 2023-01-05
Tuning/Temperament bug


larspalo:
While working on a sample set I discovered a major tuning and temperament bug that exist in GO. I don't know for how long it has existed but here's the problem: the pitch tuning (either from ODF with PitchTuning lines or from within GO in the organsettings dialog which end up in the .cmb files as Tuning lines) _and_ the selected temperament now work cumulatively. They are **not** supposed to do that! PitchTuning/organ settings dialog tuning should **only** affect the "original" temperament. As soon as another temperament is selected the embedded pitch in the samples (and/or pitch info supplied from ODF with MIDIKeyNumber _and_ PitchCorrection) should be the only things that matter to the actual tuning of a sample.

larspalo:
Steps to reproduce:
1. Load/Open a sample set with embedded pitch info in the samples (like PiteaMHS or similar)
2. Select any other temperament than original, for instance Equal
3. Open the organsettings dialog and adjust the tuning for a sample and it's pitch is changed even though it only should affect the original temperament

Expected behavior: tuning adjusted from the organsettings dialog (which equals to PitchTuning lines in ODF) should only affect pitch when original temperament is selected.

larspalo:
I can now confirm that this bug is old, it exist with the last build from SourceForge code. I think that the main reason that no one has detected it is that it's only apparent if there is a need to re-tune the original samples heavily - which then throws the temperaments out of whack. As long as no pitch tuning is done everything works just fine... But this is not how it's supposed to work.

larspalo:
@oleg68 I've tested a fix for this issue but since you're in the middle of refactoring I'm not sure if it fits in the scheme of things right now or if this issue should be solved in another way.

The only additional problem I've found is that with this fix the tuning in the master panel must be modified to also include PitchCorrection so that fine tuning adjustments can be done globally (the transposing already works as it should). But as I see it that's another issue and it's a fairly small thing compared to this bug.

oleg68:
@larspalo Thank you for reporting this bug and for providing a fix. But this problem is not so easy as it seems to be.

1. I remember [my old talking with Martin](https://sourceforge.net/p/ourorgan/feature-requests/26/) where Martin said:

> When implementing temperaments, I wanted to be able to change the pitch via GUI in non-original temperaments too.

[...]

PR #1341, opened 2023-01-13
Refactored temperament tuning

larspalo:
According to discussion in #1333 this is changed to just re-factoring tuning handling code in preparation for a better solution of the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,3 @@
-- Refactored tuning according to discussion in https://github.com/GrandOrgue/grandorgue/issues/1333
 - Fixed help: the wrong panal positioning on wayland was documented https://github.com/GrandOrgue/grandorgue/issues/1271
 - Fixed typos and spelling errors in the Help
 - Fixed continuing loading an organ after an exception in one loading thread
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Refactored tuning according to discussion in https://github.com/GrandOrgue/grandorgue/issues/1333
 - Fixed help: the wrong panal positioning on wayland was documented https://github.com/GrandOrgue/grandorgue/issues/1271
 - Fixed typos and spelling errors in the Help
 - Fixed continuing loading an organ after an exception in one loading thread
--- a/src/core/temperaments/GOTemperament.cpp
+++ b/src/core/temperaments/GOTemperament.cpp
@@ -13,22 +13,21 @@ GOTemperament::GOTemperament(wxString name, wxString group)
   : m_Group(group),
     m_GroupTitle(wxEmptyString),
     m_Name(name),
-    m_Title(wxEmptyString) {}
+    m_Title(wxEmptyString),
+	m_respectPitchTuning(true) {}
 
 GOTemperament::GOTemperament(
   wxString name, wxString title, wxString group, wxString groupTitle)
-  : m_Group(group), m_GroupTitle(groupTitle), m_Name(name), m_Title(title) {}
+  : m_Group(group),
+    m_GroupTitle(groupTitle),
+    m_Name(name),
+    m_Title(title),
+    m_respectPitchTuning(true) {}
 
 GOTemperament::~GOTemperament() {}
 
 float GOTemperament::GetOffset(
-  bool ignorepitch,
-  unsigned midi_number,
-  unsigned wav_midi_number,
-  float wav_pitch_fract,
-  float harmonic_number,
-  float pitch_correction,
-  float default_tuning) const {
+  unsigned index) const {
   return 0;
 }
 
@@ -42,4 +41,8 @@ wxString GOTemperament::GetGroup() const { return m_Group; }
 
 wxString GOTemperament::GetGroupTitle() const {
   return m_GroupTitle.IsEmpty() ? wxGetTranslation(m_Group) : m_GroupTitle;
+}
+
+bool GOTemperament::GetRespectPitchTuning() const {
+  return m_respectPitchTuning;
 }
--- a/src/core/temperaments/GOTemperament.cpp
+++ b/src/core/temperaments/GOTemperament.cpp
@@ -14,7 +14,7 @@ GOTemperament::GOTemperament(wxString name, wxString group)
     m_GroupTitle(wxEmptyString),
     m_Name(name),
     m_Title(wxEmptyString),
-	m_respectPitchTuning(true) {}
+    m_respectPitchTuning(true) {}
 
 GOTemperament::GOTemperament(
   wxString name, wxString title, wxString group, wxString groupTitle)
@@ -26,10 +26,7 @@ GOTemperament::GOTemperament(
 
 GOTemperament::~GOTemperament() {}
 
-float GOTemperament::GetOffset(
-  unsigned index) const {
-  return 0;
-}
+float GOTemperament::GetOffset(unsigned index) const { return 0; }
 
 wxString GOTemperament::GetName() const { return m_Name; }
 
--- a/src/core/temperaments/GOTemperament.cpp
+++ b/src/core/temperaments/GOTemperament.cpp
@@ -14,7 +14,7 @@ GOTemperament::GOTemperament(wxString name, wxString group)
     m_GroupTitle(wxEmptyString),
     m_Name(name),
     m_Title(wxEmptyString),
-    m_respectPitchTuning(true) {}
+    m_IsTemperamentOriginalBased(true) {}
 
 GOTemperament::GOTemperament(
   wxString name, wxString title, wxString group, wxString groupTitle)
@@ -22,7 +22,7 @@ GOTemperament::GOTemperament(
     m_GroupTitle(groupTitle),
     m_Name(name),
     m_Title(title),
-    m_respectPitchTuning(true) {}
+    m_IsTemperamentOriginalBased(true) {}
 
 GOTemperament::~GOTemperament() {}
 
@@ -40,6 +40,6 @@ wxString GOTemperament::GetGroupTitle() const {
   return m_GroupTitle.IsEmpty() ? wxGetTranslation(m_Group) : m_GroupTitle;
 }
 
-bool GOTemperament::GetRespectPitchTuning() const {
-  return m_respectPitchTuning;
+bool GOTemperament::GetIsTemperamentOriginalBased() const {
+  return m_IsTemperamentOriginalBased;
 }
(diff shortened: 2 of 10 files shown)


src/core/temperaments/GOTemperament.cpp
```diff
@@ -13,24 +13,20 @@ GOTemperament::GOTemperament(wxString name, wxString group)
   : m_Group(group),
     m_GroupTitle(wxEmptyString),
     m_Name(name),
-    m_Title(wxEmptyString) {}
+    m_Title(wxEmptyString),
+    m_IsTemperamentOriginalBased(true) {}
```
oleg68, line 17: It is better to introduce the third parameter `isOriginalBased = true` to the constructor.
So
```suggestion
    m_IsTemperamentOriginalBased(isOriginalBased) {}
```

src/core/temperaments/GOTemperament.cpp
```diff
@@ -13,24 +13,20 @@ GOTemperament::GOTemperament(wxString name, wxString group)
...
 
 GOTemperament::GOTemperament(
   wxString name, wxString title, wxString group, wxString groupTitle)
-  : m_Group(group), m_GroupTitle(groupTitle), m_Name(name), m_Title(title) {}
+  : m_Group(group),
+    m_GroupTitle(groupTitle),
+    m_Name(name),
+    m_Title(title),
+    m_IsTemperamentOriginalBased(true) {}
```
oleg68, line 25:
```suggestion
    m_IsTemperamentOriginalBased(isOriginalBased) {}
```

src/core/temperaments/GOTemperament.cpp
```diff
@@ -43,3 +39,7 @@ wxString GOTemperament::GetGroup() const { return m_Group; }
 wxString GOTemperament::GetGroupTitle() const {
   return m_GroupTitle.IsEmpty() ? wxGetTranslation(m_Group) : m_GroupTitle;
 }
+
+bool GOTemperament::GetIsTemperamentOriginalBased() const {
```
oleg68, line 43: This method is good for inlining in the header

src/core/temperaments/GOTemperament.h
```diff
@@ -16,25 +16,20 @@ class GOTemperament {
   wxString m_GroupTitle;
   wxString m_Name;
   wxString m_Title;
+  bool m_IsTemperamentOriginalBased; // To know how tuning should be handled
```
oleg68, line 19:
```suggestion
private:
  bool m_IsTemperamentOriginalBased; // To know how tuning should be handled
```

src/core/temperaments/GOTemperament.h
```diff
@@ -16,25 +16,20 @@ class GOTemperament {
...
 
-  virtual float GetOffset(
-    bool ignorepitch,
-    unsigned midi_number,
-    unsigned wav_midi_number,
-    float wav_pitch_fract,
-    float harmonic_number,
-    float pitch_correction,
-    float default_tuning) const;
+  virtual float GetOffset(unsigned index) const;
```
oleg68, line 27:
```suggestion
  // if m_IsTemperamentOriginalBased then it returns the deviation from the original temperament
  // else it returns the deviation from the equal temperament
  virtual float GetOffset(unsigned index) const;
```

src/core/temperaments/GOTemperament.h
```diff
@@ -16,25 +16,20 @@ class GOTemperament {
...
   wxString GetName() const;
   wxString GetTitle() const;
   wxString GetGroup() const;
   wxString GetGroupTitle() const;
+  bool GetIsTemperamentOriginalBased() const;
```
oleg68, line 32:
```suggestion
  bool IsTemperamentOriginalBased() const { return m_IsTemperamentOriginalBased; }
```

1. The `Is` prefix is sufficient for boolean getters;
2. The method is quite simple for inlining.

src/core/temperaments/GOTemperamentCent.cpp
```diff
@@ -24,6 +24,7 @@ GOTemperamentCent::GOTemperamentCent(
   m_Tuning[9] = 0;
   m_Tuning[10] = 0;
   m_Tuning[11] = 0;
+  m_IsTemperamentOriginalBased = false;
```
oleg68, line 27: It's better to pass `false` to the `GOTemperament` constructor.

src/core/temperaments/GOTemperamentCent.cpp
```diff
@@ -53,6 +54,7 @@ GOTemperamentCent::GOTemperamentCent(
   m_Tuning[9] = i10;
   m_Tuning[10] = i11;
   m_Tuning[11] = i12;
+  m_IsTemperamentOriginalBased = false;
```
oleg68, line 57: It's better to pass false to the GOTemperament constructor.

src/core/temperaments/GOTemperamentCent.cpp
```diff
@@ -83,23 +85,9 @@ GOTemperamentCent::GOTemperamentCent(
...
```
oleg68, line 88: It's better to pass false to the GOTemperament constructor.

src/core/temperaments/GOTemperamentCent.h
```diff
@@ -47,14 +47,7 @@ class GOTemperamentCent : public GOTemperament {
...
 
-  virtual float GetOffset(
-    bool ignorepitch,
-    unsigned midi_number,
-    unsigned wav_midi_number,
-    float wav_pitch_fract,
-    float harmonic_number,
-    float pitch_correction,
-    float default_tuning) const;
+  virtual float GetOffset(unsigned index) const;
```
oleg68, line 50:
```suggestion
  virtual float GetOffset(unsigned index) const override { return m_Tuning[index]; }
```

src/core/temperaments/GOTemperamentCent.cpp
```diff
@@ -83,23 +85,9 @@ GOTemperamentCent::GOTemperamentCent(
...
-  double concert_pitch_correction = 0;
-
-  if (!ignorePitch && wav_midi_number) {
-    concert_pitch_correction = (100.0 * wav_midi_number - 100.0 * midi_number
-                                + log(8.0 / harmonic_number) / log(2) * 1200)
-      + wav_pitch_fract;
-  }
-  return m_Tuning[midi_number % 12] - default_tuning - concert_pitch_correction
-    + pitch_correction;
+float GOTemperamentCent::GetOffset(unsigned index) const {
```
oleg68, line 91: It is too simple for placing in the implementation.

src/grandorgue/model/GOSoundingPipe.cpp
```diff
@@ -486,17 +509,11 @@ void GOSoundingPipe::UpdateAudioGroup() {
...
   else
-    m_TemperamentOffset = temperament.GetOffset(
-      m_PipeConfigNode.GetEffectiveIgnorePitch(),
-      m_MidiKeyNumber,
-      m_SoundProvider.GetMidiKeyNumber(),
-      m_SoundProvider.GetMidiPitchFract(),
-      m_HarmonicNumber,
-      m_PitchCorrection,
-      m_PipeConfigNode.GetDefaultTuning());
+    m_TemperamentOffset = temperament.GetOffset(m_MidiKeyNumber % 12);
```
oleg68, line 516:
```suggestion
    m_TemperamentOffset = m_RetunePipe ? temperament.GetOffset(m_MidiKeyNumber % 12) : 0;
```

Then we can get rid of `if`.

oleg68 (changes requested):

larspalo:
@oleg68 @rousseldenis The pre-commit hooks very often fails with the "corrections" that itself does on the code formatting. It's extremely frustrating. If it doesn't work and help... It simply doesn't have any place! It's only an hindrance when one need to get creative with the formatting just to bypass the so called "errors" the pre-commit has done itself and then "detects" in the formatting, which needless to say has absolutely nothing to do with the actual functionality of the code...

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Refactored tuning according to discussion in https://github.com/GrandOrgue/grandorgue/issues/1333
```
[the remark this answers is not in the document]
larspalo: Agree.

src/core/temperaments/GOTemperament.cpp
larspalo: Ok. But it's not better, it's just possible to override at will in the calls to the constructor and not hard coded to either specific case.

src/grandorgue/model/GOSoundingPipe.cpp
larspalo: It's possible. But I won't do it. One thing is the pre-commit hook failing and the other is that ternary operator is inherently more unclear to read than the more verbose if ... else blocks.

src/core/temperaments/GOTemperament.h
larspalo: It is not as clear as GetOffset()

oleg68 (approved):

oleg68:
> The pre-commit hooks very often fails with the "corrections" that itself does on the code formatting. It's extremely frustrating.

When the pre-commit fails, it corrects the formatting issues itself and changes the files. So you can just stage them and commit to the git repository.

If you do not like the formatting proposed by the hook, you may try to adopt https://github.com/GrandOrgue/grandorgue/blob/master/.clang-format to your wishes (of cause, in a separate PR because usually it requires reformatting the whole sourcecode), but it is not so easy.

rousseldenis:
@larspalo the idea behind pre-commit is to not take care about formatting for developer.

The thing to understand as @oleg68 said is that it can take two passes to get your changes committed.

So, first, do your changes.

Secondly, do a git commit. If your code is well formatted, it's directly committed. if not, pre-commit do the main changes on its own. So, you just need to run the git commit command one more time and that's it.

Don't take care about how pre-commit changed your code even it seems weird to you.

If you think it's wrong, maybe some options need to be changed.

larspalo:
@oleg68 @rousseldenis So if it fails the second time I run it (without me doing any changes) there's a bug, or conflicting instructions it tries to follow?

oleg68:
> @oleg68 @rousseldenis So if it fails the second time I run it (without me doing any changes) there's a bug, or conflicting instructions it tries to follow?

You have to stage changed files before the second attempt of commit.

larspalo:
> You have to stage changed files before the second attempt of commit.

Using git add separately? Because git commit -a -m 'Some comment' fail.

rousseldenis:
> > You have to stage changed files before the second attempt of commit.
> 
> Using git add separately? Because git commit -a -m 'Some comment' fail.

Yes, sorry you have to add them again

rousseldenis (approved):

Issue #1271, opened 2022-11-22
Not remembering Windows position or size


mattrixed:
Linux Ubuntu 22.04
GO 3.9.1

Loading samplet freisach for eg window is full screen.  Move it and position it - save - close down and start up again goes back to full screen.

oleg68:
@mattrixed Could you test the same with the demo organ?

oleg68:
@mattrixed I installed ubuntu22, grandorgue 3.9.2 and I can not reproduce this problem: after restarting GrandOrgue and loading Friesach it opens at the same position and size as it was saved.

I need more information and tests from you for reproducing and fixing this problem. Are you ready to do more tests and to provide me additional information?

mattrixed:
I too with the system running in the office/studio can't replicate either - the only difference is the monitor - when connected to the small touch screen 19" 4:3 monitor 1280x1024 it doesn't save - something to do with the display size if it detects a large screen (like in office 2560x1600) its fine if its under the size of the odf if goes full screen - 

Not a big deal.....
Have installed POP! distro will see if still does it with that also....

oleg68:
This issue is related to the `Wayland` windows server. As a workaround, when you log in the ubuntu session, you may click at the settings icon (at right bottom of the window) and choose `Ubuntu on Xorg`.

oleg68:
@larspalo I remember you use Ubuntu. Do you use `wayland` or `xorg`? Do you have any issues with positioning?

I've never used `wayland` before and now I see that the windows positioning code in GrandOrgue does not work at all with wayland. wxTopWindow::GetPosition() and wxTopWindow::GetRect() returns some inadequate result.

larspalo:
@oleg68 I'm still using Ubuntu 20.04 mostly at home so it is X11 (xorg) by default. I think they changed it in 21.04 so when I decide to upgrade to 22.04 it's likely that Wayland will be used. However, window positioning works fine here with 20.04. We really should be careful to not "fix" things that really are bugs in other software, I think. If we're using the "correct" wx methods and issues still arise it's still fine and the bugs should be fixed where they really are. If, however, there are multiple ways of doing things and one version indeed gives better compatibility then that's another case... In my experience, too many "special fixes" only makes for more issues somewhere down the line and it makes the code more difficult to maintain.

oleg68:
@larspalo After some research I found two issues:

[...]

PR #1342, opened 2023-01-14
Fixed saving position when some panel is outside the screen area https://github.com/GrandOrgue/grandorgue/issues/1271

oleg68:
Resolves: #1271

This PR fixes restoring panel position if the frame was not fully inside the screen area.

It works fine on windows, but some window managers (ex. gnome-shell/mutter) return the panel inside the screen area if it is only partially visible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
+- Fixed saving position when some panel is outside the screen area https://github.com/GrandOrgue/grandorgue/issues/1271
 - Fixed playing release samples for very short notes https://github.com/GrandOrgue/grandorgue/issues/1222
-- Fixed help: the wrong panal positioning on wayland was documented https://github.com/GrandOrgue/grandorgue/issues/1271
+- Fixed help: the wrong panel positioning on wayland was documented https://github.com/GrandOrgue/grandorgue/issues/1271
 - Fixed typos and spelling errors in the Help
 - Fixed continuing loading an organ after an exception in one loading thread
 - Fixed size of text fields in the Organ Settings dialog on OsX https://github.com/GrandOrgue/grandorgue/issues/1315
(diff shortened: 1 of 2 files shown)


larspalo:
With this adjustment the behavior is different on Ubuntu (20.04). Earlier if multiple panels were partially outside the screen, one of them (and only one!) would center and the other moved just to edge of screen. Now, with this, the panels are not centered at all but moved to be visible (inside screen) just aligned at the closest edge.

I don't really mind this behavior but it seems it's not quite what's intended.

oleg68:
> With this adjustment the behavior is different on Ubuntu (20.04). Earlier if multiple panels were partially outside the screen, one of them (and only one!) would center and the other moved just to edge of screen. Now, with this, the panels are not centered at all but moved to be visible (inside screen) just aligned at the closest edge.
 
I installed ubuntu20 and I made tests with the old GO version (3.4.1-1) and with the new version (3.4.1-1.13) with this PR.

Before `Save`:
before-save

After starting GrandOrgue 3.4.1-1:
restore-3 9 4

After starting GrandOrgue 3.4.1-13:
restore-3 9 13

So the new behavior is the same as you describe, but I **could not reproduce** the behavior you described with the old version. Earlier **all** panels

> I don't really mind this behavior but it seems it's not quite what's intended.

May be I intended restoring panels at the same positions where they were saved, but it is impossible with gnome shell: after GO opens all panels, gnome shell **moves** all panels to be fully visible. GrandOrgue can not distinguish moves are made by gnome shell from they are made by the user manually.

The intended behavior works only on Windows, where the window manager is "less cleaver" than gnome shell.

Whatever I find the new behavior is better than the old one, so I ask you to approve this PR.

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1222, opened 2022-10-11
The release section of samples will refuse to play if the note is short enough


ray890:
When a note is played and released, the sample is expected to play its release section before the voice is dropped. However, if the note is released quickly enough, the voice is instantly dropped instead, creating an audible click sound, 

If the note is right under the threshold of triggering this bug, then it will sometimes happen and sometimes not happen.

This will always happen, independently of the organ used, output device, output mode, latency settings, or computer.

This has had an impact with my MIDI Sequences that I have intended to play, and this can also have an impact on keyboard players who play staccato or fast musical scores.

https://user-images.githubusercontent.com/11812575/194974595-f3f88e32-830c-41c8-90ad-0a675d03bedb.mp4

Here's the interesting part about this. If I keep an organ instance open for a very long time (I don't know what triggers this), the bug will not occur with any length of note, but when I restart the organ the bug will return.

oleg68:
1. What is the GrandOrgue version?
2. What sampleset are you using?
3. Is the checkbox File->Settings-> Release Sampling set or not?
4. What are the values of File->Settings->Audio Sample Rates, Samples per buffer?

willeke1234:
I did some debugging, in [GOSoundEngine::ProcessSampler](https://github.com/GrandOrgue/grandorgue/blob/master/src/grandorgue/sound/GOSoundEngine.cpp#L250):

```
if (
  sampler->stop && sampler->stop <= m_CurrentTime
  && sampler->stop - sampler->time <= block_time)
  sampler->pipe = NULL;
```
`sampler->stop - sampler->time` is `block_time` and `sampler->pipe` is set to `NULL`. The release uses this pipe and doesn't play.

willeke1234:
Discussions:
Grand Orgue Sound Glitches #162
Cutoff on very short keypress? #306
Glitching problem #405
Sound breaking up badly with builds/versions above 1694 #494 (recommended reading)
Audio Problems with Large Sample Set #505
Samples will click/pop when organ is fed with short/rapid notes #1120

willeke1234:
Is only the click sound a problem or also the missing release?

oleg68:
Seems this bug has been introduced with 58c0bf31982d907829cfb7aa93d6f707a94a93ed and it is reproduced when the length of the keypress is less than one buffer. So setting a low buffer size is a possible workaround.

oleg68:
[...]

PR #1343, opened 2023-01-16
Fixed playing release samples for very short notes https://github.com/GrandOrgue/grandorgue/issues/1222

oleg68:
Resolves: #1222

This PR reverts [this Martin's change](https://github.com/GrandOrgue/grandorgue/commit/58c0bf31982d907829cfb7aa93d6f707a94a93ed).

I could not find any explanation, what kind of problems that change resolved. 

After this PR #1222 stopped reproducing. Today I played for two hours and could not hear any sound artifacts, mentioned by Martin in the commit message. Possible, there was another issue that has already been resolved.


larspalo (approved):
Fixes the issue at hand. If there's a problem with the "fix" we'll discover it eventually. The only thing I can think of that could cause a problem is if the buffer wasn't filled with data some strange things could possibly have happened in the fading earlier, but I have no clear memory of it.

rousseldenis (approved):

Issue #828, opened 2021-10-31
[Feature Request] Officially allow end of line comments


oleg68:
### Discussed in https://github.com/GrandOrgue/grandorgue/discussions/819

<div type='discussions-op-text'>

<sup>Originally posted by **vpoguru** October 31, 2021</sup>
It would greatly improve the clarity of the ODF code if comments could (officially) be placed at the end of the line. This already works in all 3 operating systems, but GrandOrgue gives a warning.

Example:

    Stop001=101 ; Principal 8'
    Stop002=102 ; Quintadena 8'

... instead of:

     ; Principal 8'
    Stop001=101
    ; Quintadena 8'
    Stop002=102</div>

PR #1359, opened 2023-01-28
Added support of ODF comments from ';' at any position to the end of line https://github.com/GrandOrgue/grandorgue/issues/828

oleg68:
Resolves: #828

I've checked, that hauptwerk also can open files with comments starting at not the first position, so there are no reasons why to forbid it in GrandOrgue.

### DIFF
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -6260,7 +6260,8 @@ UTF-8, if it starts with an appropriate byte order marker.
       </para>
       <para>
         Comments are any text from <emphasis>;</emphasis> to the end of line.
-        Empty lines or lines containing only comments are ignored.
+        Empty lines or lines containing only comments are ignored. 
+        Thus it's not possible to include <emphasis>;</emphasis> in any string
       </para>
       <para>
 Various settings are grouped in blocks. Each block is started with a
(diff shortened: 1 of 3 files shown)


larspalo:
The only drawback with this approach to comments is that you cannot have a ; in a text string anymore.

oleg68:
> The only drawback with this approach to comments is that you cannot have a ; in a text string anymore.

May be to introduce an escaping or a quote-separated syntax of text fields (not to regard `;` in the quotas as a comment delimiter)?

larspalo:
@oleg68 No, I don't think we should not complicate things too much. I think we can _either_:

1. Disallow any comments on lines that are read as strings, while just skipping them as comments on any other line
2. Document the behavior in help

_or_ just document in the help that semicolon cannot be used in strings in the odf (because they are parsed as comments)

oleg68:
> @oleg68 > just document in the help that semicolon cannot be used in strings in the odf (because they are parsed as comments)

I like this choice. Moreover HW1 reads string keys only until the semicolon.

help/grandorgue.xml
```diff
@@ -6259,7 +6259,8 @@ It is a text file in ISO-8859-1 encoding. The standard extension is
 UTF-8, if it starts with an appropriate byte order marker.
       </para>
       <para>
-Comment lines are started with <emphasis>;</emphasis> in the first column. Empty lines are ignored.
+        Comments are any text from <emphasis>;</emphasis> to the end of line.
```
larspalo, line 6262: Thus it's not possible to include ; in any string.

larspalo (approved):

larspalo (approved):

rousseldenis (approved):

Issue #827, opened 2021-10-31
[Feature Request] Officially support "/" as path separator


oleg68:
### Discussed in https://github.com/GrandOrgue/grandorgue/discussions/821

<div type='discussions-op-text'>

<sup>Originally posted by **vpoguru** October 31, 2021</sup>
The path separator "/" already works in all 3 operating systems, but GrandOrgue gives a warning (at least if "Perform strict ODF" is checked).

Linux and macOS terminals work with "/" anyway, and Windows supports this separator as well. If I remember correctly, the "/" separator could even be used (unofficially) in DOS times.</div>

larspalo:
In july 2014 (regarding rev 1713) we had a short discussion about this. I thought basically the same thing as you, that we could use whichever path separator / or \ because wx will handle it correctly anyway but Martin was of another opinion. The help since old specifies \ as canonical separator and basic reason was complying to the HW1 standard.

I won't complain about removing the warning about which path separator is used, though.

PS. In DOS one would use / to specify parameters or switches to a program, never the paths, but this is very much irrelevant now.

oleg68:
I propose to introduce a separate settings checkbox ``Check ODF for HW1 compatibility`` and print warnings about `/` delimiter only if is enabled.

PR #1361, opened 2023-01-28
Added full support of '/' as the directory sepearator in ODF unless "CheckODF for HW1-compatibility" is set https://github.com/GrandOrgue/grandorgue/issues/827

oleg68:
Resolves: #827

I added the new GrandOrgue setting option: `Check ODF for HW1-compatibility`. By default it has the same value as `Strict ODF check`.

The warning about `/` is logged only when `Check ODF for HW1-compatibility` is enabled.


oleg68:

rousseldenis (approved):

Issue #1351, opened 2023-01-22
Capability of preserving the original pitch and deviations when switching the temperaments


oleg68:
Now GrandOrgue allows a user to choose a temperament. One of them is the `Original` temperament , others are a lot of  predefined user-defined temperaments.

There are some issues releated to the current temperament implementations:

1. When a non-original temperament selected, the organ is automatically tuned and pitched to 440hz (mentioned in https://github.com/GrandOrgue/grandorgue/discussions/1162#discussion-4330176).
2. Any user-tuning settings made with GUI can be appropriate for either the original temperament or non-original temperaments, but not for both #1333.

I propose to do several enchancements.

1. To introduce the `Temperament` key in the ODF with the name of original temperament.
2. To implement a separate feature `AutoTuning`. It would allow to tune the organ using the embeded pitch information regardless the temperament used. To introduce a special menu entry `Auto Tuning`,
3. To allow the 'PitchCorrection' ODF key at the organ, windchest, and rank level. It would allow to specify the original pitch other than 440hz and use it with auto tuning.
4. If the original temperament is specified in the ODF, switching to a non-original temperament may be done with or without the automatic tuning. In the second case the difference between the temperament offsets of the original and the target temperament is used.
5. To remove the `Tuning` field from the Organ Settings and to introduce two another user-tunable fields: `ManualTuning` and `AutoTuningCorrection`. 

@larspalo @rousseldenis do you agree with this approach?

oleg68:
> 1. The recorded samples contain everything that's original! ;-)

I call ``Original Temperament`` what was intended by the organ builder. The real pipes always have some deviations from the original temperament.

Now when the user chooses any non-original temperament, GO tunes all the pipes and discards the recorded deviations. It makes the sound "atrificial".

I'd like to separate the auto-tuning feature (discarding all deviations) from switching the temperaments. If the original intended organ temperament was known, GrandOrgue would be capable to recalculate temperaments with preserving the original pitch and deviations.

> I think you should abandon the idea of specifying temperament in odf and thus this point is invalid.

But capability of preserving the original pitch and deviations when switching the temperaments looks as attractive.

larspalo:
The general idea of using equal temperament with a1=440 Hz as base is because that's the general accepted standard. The fact that it almoast isn't used anymore in real life is another thing completely (most orchestras and opera houses uses another pitch...). It's still a point of reference (actually _the_ point of reference).

larspalo:
[...]

PR #1366, opened 2023-02-01
Added PitchCorrection for organs and windchest. Pipe999PitchCorrection became additive to PitchCorrection of the rank, of the windchest and of the organ https://github.com/GrandOrgue/grandorgue/issues/1351

oleg68:
I removed PitchCorrection from GORank and GOSoundingPipe to GOPipeConfig.

As a result, the `PitchCorrection` ODF key became available not only on the rank level, but also at the windchest and at the organ level.

As discussed in #1351, earlier `PitchCorrection` was just a default for `Pipe999PitchCorrection`, but now `PitchCorrection` of all levels are added to `Pipe999PitchCorrection`. It may be incompatible with some old ODFs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Added PitchCorrection for organs and windchest. Pipe999PitchCorrection became additive to PitchCorrection of the rank, of the windchest and of the organ https://github.com/GrandOrgue/grandorgue/issues/1351
 - Added full support of '/' as the file sepearator in ODF unless 'Check ODF for HW1-compatibility' is set https://github.com/GrandOrgue/grandorgue/issues/827
 - Added support of ODF comments from ';' at any position to the end of line https://github.com/GrandOrgue/grandorgue/issues/828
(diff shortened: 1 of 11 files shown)


larspalo:
@oleg68 There's (at least) a small problem with how this PR changes the behavior. I suggest you test it with PiteaMHS and the Voix Celeste (and Gamba together). The goal of the pitch correction is to work on top of the provided pitch info from the file and thus allow the de-tuning to happen even in a selected temperament. Now with this PR - the voix celeste is tuned perfect and the pitch correction specified in the odf is kind of ignored.

oleg68:
@larspalo Yes, there was a typo in GOPipeConfigNode::GetEffectivePitchCorrection. I fixed it.

larspalo (approved):
Now it seems to behave as it should. I could apply an organ level PitchCorrection that could indeed keep the organ at same pitch as the original temperament even when another temperament was chosen.

rousseldenis (approved):

PR #1370, opened 2023-02-04
Splitted Tuning Organ Settings to ManualTuning and AutoTuningCorrection https://github.com/GrandOrgue/grandorgue/issues/1333

oleg68:
Resolves: #1333 

As discussed in #1333 and #1351, I introduced two separate (GUI) settings `ManualTuning` and `AutoTuningCorrection` instead of `Tuning` before. Now the first one is used with the original temperament, the second one is used with other temperaments.


larspalo:
It seems that the PitchTuning info from the odf is migrating into the AutoTuning values that should only be populated from the PitchCorrection. Also this is somehow throwing off the original temperaments tuning.

src/grandorgue/model/pipe-config/GOPipeConfig.cpp
```diff
@@ -36,6 +37,32 @@ GOPipeConfig::GOPipeConfig(
...
+    false,
+    legacyTuning - m_PitchTuning);
+  m_AutoTuningCorrection = cfg.ReadFloat(
+    CMBSetting,
+    group,
+    prefix + WX_AUTO_TUNING_CORRECTION,
+    -1800,
+    1800,
+    false,
+    legacyTuning - m_PitchTuning);
```
larspalo, line 63: This is likely what's causing issues. For the auto tuning you should not do that calculation.

oleg68: Changed to 0

oleg68:

larspalo:
@oleg68 I've tested a build from your repository of this PR with the latest changes and now the temperament works just fine and the auto tuning is at zero like it should to begin with! But still with the original temperament it seems like something's off a bit. I suspect that the values are added twice or something like that. I've not had time to really dig into the code yet but I wanted to give you a little heads up on the problem.

oleg68:
@larspalo 

Seems I fixed it. The problem was the ODF had PitchTuning but there was not a preset existing.

larspalo (approved):
@oleg68 Excellent work! I really like that the odf pitch tuning lines are no longer filled into the manual tuning spin controls. I've tested the build from your fork and it works exactly as it should no matter what I've tried. Also I like that the pitch changes from the overall controls now work just fine either in original temperament or in any selected but are kept separate from the organsettings dialog tuning values. Thanks for your work and all the effort you've put into improving this aspect of GO! This is really a great (and in my mind very important) improvement! I happily approve!

oleg68:

rousseldenis (approved):

Issue #1375, opened 2023-02-07
Feature request: version number in cmb file


willeke1234:
The fields in the cmb file can change from version to version. Can the version number be added at the start? Something like

```
Version=3.9.4-1
```

(Yes, I know I shouldn't edit cmb files, but adding MIDI events can be automated that way.)

oleg68:
Done as #1388

PR #1388, opened 2023-02-14
Added storing the GrandOrgueVersion key in the Organ Settings file https://github.com/GrandOrgue/grandorgue/issues/1375

oleg68:
Release: #1375


larspalo (approved):
@oleg68 Do you have any intended use for this? Some kind of conversion if we alter things?

oleg68:
I thougt to make a check if the settings have been saved with a new version of GO and to put a warning message. But not now.

rousseldenis (approved):

PR #1380, opened 2023-02-10
Added capability of redefining ReleaseLength and IgnorePitch on each level of Organ Settings

oleg68:
I added new Organ Settings `ReleaseLength` and `IgnorePitch` available for setting at each level: Organ, Windchest, Rank and Pipe.

This PR includes all capabilities of #774, so that PR may be closed after this.

PR #1401, opened 2023-02-23
Fixed crash on loading an incorrect organ

oleg68:
#1380 caused the GO crash on attempt of loaging a wrong ODF.

The reason was GOFrame::LoadOrgan called SetOrganModificationListener of unitialised organ.

### DIFF
--- a/src/grandorgue/GOFrame.cpp
+++ b/src/grandorgue/GOFrame.cpp
@@ -545,6 +545,13 @@ void GOFrame::Init(wxString filename) {
   clean.Cleanup();
 }
 
+void GOFrame::AttachToOrganController(bool isToAttach) {
+  GOOrganController *pOrgan = GetOrganController();
+
+  if (pOrgan)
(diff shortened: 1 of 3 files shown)


src/grandorgue/GOFrame.cpp
```diff
@@ -545,6 +545,13 @@ void GOFrame::Init(wxString filename) {
   clean.Cleanup();
 }
 
+void GOFrame::AttachToOrganController(bool isToAttach) {
```
rousseldenis, line 548: Just wondering why calling a function that attachs with a parameter like that. Either the function name means something else, either the parameter does.

rousseldenis: Maybe a comment should be welcome

oleg68: Renamed to AttachDetachOrganController

rousseldenis (approved):

larspalo (approved):

Issue #1386, opened 2023-02-12
Feature request: to have label without background image


eturpault:
Currently in GrandOrgue a label must have necessarily a background image (built-in image defined with the attribute DispImageNum or external image defined by the attribute Image). This request is to add the possibility to define a label without background image, so only displaying the text defined in the attribute Name.
Some Hauptwerk sample sets have labels displayed without background image. In order to render them in GrandOrgue, OdfEdit has to use a built-in backgroupd, which is not beautiful.
In the image below on the left what should be obtained thanks to this feature request, on the right what is currently obtained with GO and a sample set converted from HW : 
![label without background](https://user-images.githubusercontent.com/51839171/218321538-8d27d7f0-2d08-438f-adf4-f69fdc878d62.png)

rousseldenis:
@eturpault Isn't this possible with using images with no background property (and transparency) ?

eturpault:
Yes theorically it is a feasible solution that I had in mind, the problem that I see with this solution would be for the placement of the text. It is placed by GO at the center of the image area, it may be difficult to manage the placement with the different lengths and sizes of text inside a transparent image which has a unique size. The goal of this request is also to avoid to have to place in the sample set an additional image which is not natively included. I think that it should be easier to make a change in GO SW to simply display a text without associated background image.
In OdfEdit I used a similar solution to have silent pipe ([this discussion](https://github.com/GrandOrgue/grandorgue/issues/1385#issue-1581303556)) by using a "transparent" sound sample, this is easier to manage that having to place graphically a text with variable length and size within a transparent image which has a fixed size. But I don't like this solution which obliges to add an external wav file in the sample set folder.

larspalo:
There are more than one way to implement this in GO as a feature (not considering the workaround already possible for a sample set creator to just add a transparent image).

We can fairly easily just add a new label style (DispImageNum) with number 13 that's just a blank transparent .png. The attributes of width, height etc for a GUILabel is already in place to control the exact size and extent and flow of text.

Another possibility that requires much more work is to allow naming the Image as TRANSPARENT and create the wxBitmap on the fly according to specifications of width, height etc. Or if they are blank calculate it from the text extent of the label Name.

I'm leaning more towards the first option as it's very easy to implement...

oleg68:
@larspalo I like the second option more, because implementing the first one would require us to create a size management that would not be necessary with other 12 images.

larspalo:
[...]

PR #1402, opened 2023-02-23
Added support of labels without background image by specifying DispImageNum=0 https://github.com/GrandOrgue/grandorgue/issues/1386

oleg68:
Resolves: #1386

This PR allows to create a Label object without a background image.

If the label object has `DispImageNum=0`, then GO creates a label with a transparent image. By default it's size is 80x25 and it may be overriden with `Width` and `Heigh` keys.

### DIFF
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -8826,7 +8826,7 @@ group label font of the panel.
           <listitem>
             <para>
 	      (integer 0 - 12, default: 1) Builtin bitmap set to use. 0 means
-	      the lable does not have a background image.
+	      that the lable does not have a background image.
 	    </para>
           </listitem>
         </varlistentry>
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -8825,8 +8825,9 @@ group label font of the panel.
           <term>DispImageNum</term>
           <listitem>
             <para>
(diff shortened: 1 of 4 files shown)


help/grandorgue.xml
```diff
@@ -8825,8 +8825,9 @@ group label font of the panel.
           <term>DispImageNum</term>
           <listitem>
             <para>
-(integer 1 - 12, default: 1) Builtin bitmap set to use.
-     </para>
+	      (integer 0 - 12, default: 1) Builtin bitmap set to use. 0 means
+	      the lable does not have a background image.
```
larspalo, line 8829: that the label doesn't have a background image.

larspalo (approved):
Works for me!

help/grandorgue.xml
```diff
@@ -8826,7 +8826,7 @@ group label font of the panel.
           <listitem>
             <para>
 	      (integer 0 - 12, default: 1) Builtin bitmap set to use. 0 means
-	      the lable does not have a background image.
+	      that the lable does not have a background image.
```
larspalo, line 8829: label

larspalo (approved):

help/grandorgue.xml
oleg68: Fixed

rousseldenis (approved):

PR #1409, opened 2023-02-25
Updates copyright headers

oleg68:
See also #1414

### DIFF
--- a/src/build/CMakeLists.txt
+++ b/src/build/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Copyright 2006 Milan Digital Audio LLC
-# Copyright 2009-2022 GrandOrgue contributors (see AUTHORS)
+# Copyright 2009-2023 GrandOrgue contributors (see AUTHORS)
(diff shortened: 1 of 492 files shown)


larspalo (approved):

rousseldenis (approved):
I think I've already talked about this.

To avoid a great commit diff each year, I suggest to modify the year only when the concerned file is modified.

This is more coherent as some files are not modified (on code side) through some years.

That's my two cents on this from my other projects contribution experience

oleg68:
Then we will forget to modify the copyright.

Is there any pre-commit hook that checks for the correct year?

rousseldenis:
> Then we will forget to modify the copyright.
> 
> Is there any pre-commit hook that checks for the correct year?

That can exist if a template can be provided as configuration. I can look at

rousseldenis:
> Then we will forget to modify the copyright.

The year is not crucial for it. License kind (AGPL/GPL... )more.

rousseldenis:
I found this referenced on pre-commit site : https://github.com/Lucas-C/pre-commit-hooks#insert-license

PR #1420, opened 2023-02-27
Using soundEngine throught GOSoundStateHandler instance instead of GOOrganController

oleg68:
This is the naext PR toward removing dependency on GOOrganController from model.

1. GOEventDistributor::PreparePlayback passes a pointer to GOSoundEngine to all GOSoundStateHandler instances.
2. GOSoundStateHandler subclasses now can call GOSoundEngine methods directly instead of using GOOrganController proxies.
3. These proxies are no more in use, so they are removed.


larspalo:
@oleg68 Something has happened with the DefaultToEngaged=Y sounds not getting triggered with this PR. For instance the motor (blower) effects that should be triggered at startup of an organ doesn't anymore. The stop controls are visibly on but to trigger them one must first disable them and then activate the stop again.

oleg68:
Seems the motor sound attempts to be playing before PreparePlayback is called. I'll search more.

oleg68:
@larspalo The reason of this impact was the effect sound was played from inside of GODrawStop::PreparePlayback(). But GOEventDistributor may calls PreparePlayback() of different objects (in this case, GODrawStop and GOSoundingPipe) in the some orbitrary order, so GOSoundingPipe::GOSoundStateHandler::p_SoundEngine is initialised after GODrawStop::PreparePlayback().

For solving this problem I partially reverted #1412 and restored GOSoundStateHandler::StartPlayback() that is calles after PreparePlayback() of all objects. Then I moved the effect playing action from GODrawStop::PreparePlayback to GODrawStop::StartPlayback.

oleg68:

larspalo (approved):

larspalo:
@rousseldenis Please approve this PR as it fixes some issues.

rousseldenis (approved):

PR #1421, opened 2023-02-28
Moved GOEventDistributor::ControlChanged to GOEventHandlerList::SendControlChanged

oleg68:
Toward to removing dependencies on GOOrganController from model I removed the method GOEventDistributor::ControlChanged and replaced it with GOEventHandlerList::SendControlChanged.

### DIFF
--- a/src/grandorgue/control/GOButtonControl.cpp
+++ b/src/grandorgue/control/GOButtonControl.cpp
@@ -146,7 +146,7 @@ void GOButtonControl::Display(bool onoff) {
     return;
   m_sender.SetDisplay(onoff);
   m_Engaged = onoff;
-  m_OrganController->ControlChanged(this);
+  m_OrganController->SendControlChanged(this);
 }
 
 bool GOButtonControl::IsEngaged() const { return m_Engaged; }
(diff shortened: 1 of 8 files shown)


oleg68:

larspalo:
@oleg68 As I tested this PR, I noticed again a strange thing with the default to engaged noise effects. They start fine, but the first time I de-activate them nothing happens. The control is visibly switched off but the noise itself isn't. It requires another activation and de-activation before it's working as it should.

oleg68:
@larspalo after some testing I found that
1. The bug you mentioned was also existing before this PR. It's reason that the GORank::PreparePlayback was clearing velocities had been set when the stop had been engaged
2. I can not answer where it was introduced, but [the last commit](https://github.com/GrandOrgue/grandorgue/pull/1420/commits/f30ccd81eabd27f60150b55c01144049aa85c23d) in #1420 also fixes this problem.

So after merging this PR and #1420 there won't be the bug.

larspalo:
> So after merging this PR and #1420 there won't be the bug.

Ok, I don't doubt that. The functionality in #1420 does work as it should. I'll approve this so that we can move forward.

larspalo (approved):

rousseldenis (approved):

PR #1427, opened 2023-03-07
Removed dependency on GOOrganController from GOMidiReceiver

oleg68:
And changed to GOOrganModel

NO GO behavior should be changed

### DIFF
--- a/src/core/midi/GOMidiReceiverEventPatternList.h
+++ b/src/core/midi/GOMidiReceiverEventPatternList.h
@@ -41,8 +41,13 @@ public:
 
   GOMidiReceiverType GetType() const;
 
-  unsigned GetEventCount() const;
-  GOMidiReceiverEventPattern &GetEvent(unsigned index);
+  unsigned GetEventCount() const { return m_events.size(); }
+  const GOMidiReceiverEventPattern &GetEvent(unsigned index) const {
+    return m_events[index];
+  }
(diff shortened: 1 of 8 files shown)


src/core/midi/GOMidiReceiverEventPatternList.h
```diff
@@ -41,8 +41,13 @@ class GOMidiReceiverEventPatternList {
 
   GOMidiReceiverType GetType() const;
 
-  unsigned GetEventCount() const;
-  GOMidiReceiverEventPattern &GetEvent(unsigned index);
+  unsigned GetEventCount() const { return m_events.size(); }
+  const GOMidiReceiverEventPattern &GetEvent(unsigned index) const {
+    return m_events[index];
+  }
+  GOMidiReceiverEventPattern &GetEvent(unsigned index) {
```
larspalo, line 48: @oleg68 Any specific reason to have a const and non const that does exactly the same thing? I thought the normal procedure in such cases is that the non const should call the const version and then cast away the const... But I don't know your reasons behind it, so therefore I ask.

larspalo: @oleg68 No, don't mind it if it's still necessary. I just reacted on it as I saw the inline change and the doubling caught my eye.

larspalo (approved):

rousseldenis (approved):

PR #1430, opened 2023-03-08
Refactored  preparing OS scripts for linux

oleg68:
Due the planning end of support Ubuntu-18 on github I rewrote the prepare-ubuntu shell scrripts that
1. Do not dublicate the same package list for different processes architectures. They should reuse the same basic script
2. Can run on any debian-based system of any version. Earlier they were linked to ubuntu-18 only.

Now `build-scripts/for-linux/prepare-debian-based.sh` is a general preparation script with a single parameter - target processor architecture. If it is omitted, then the current processor architecture is assumed.

All other prepare-debian-based.sh scripts call  `build-scripts/for-linux/prepare-debian-based.sh` with a necessary parameter value.


larspalo (approved):

oleg68:
@rousseldenis could you approve this PR? It is necessary for ability to build on Github since 2023-04-01.

rousseldenis (approved):

PR #1432, opened 2023-03-09
Unified build scripts for windows

oleg68:
For unification of build scripts I
1. Renamed build-scripts/for-win64/prepare-ubuntu20.sh to prepare-debian-based.sh because it works on any debian-based linux.
2. I extraced the mingw-specific, but not depending on GrandOrgue part of build-scripts/for-win64/build-on-linux.sh to a separate script set-mingw-vars.sh that can be easy reused in other projects requiring cross-compilation for windows on linux.


larspalo (approved):

oleg68:
@rousseldenis could you approve this PR? It is necessary for ability to build on Github since 2023-04-01.

rousseldenis (approved):

PR #1436, opened 2023-03-10
Removed old cmake files

oleg68:
This PR removes a clone of the standard cmake `BundleUtilities.cmake` file and it's usage.

This clone was added by Martin at https://github.com/GrandOrgue/grandorgue/commit/2b9e327bc93229dc6f0ae981e7db94fde0916f46 in 2016 and was modified later because of some cmake bugs.

But these bugs seems have already been fixed in cmake, so I couldn't find any differencies in the resulting osx and windows bundles with and without the clone. So this clone is no more necessary.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1437, opened 2023-03-10
Bugfix/build on osx

oleg68:
After changing something in CMakeLists.txt, I experienced some difficults when building GrandOrgue for OSx: https://github.com/oleg68/GrandOrgue-official/actions/runs/4384854530

The reason was using the variable name `target` that may be conlicting with other places.

I renamed this variable to `bundledtarget`.


larspalo (approved):

oleg68:
@rousseldenis could you approve this PR? It is necessary for implementing the yaml combinations features.

rousseldenis (approved):

PR #1458, opened 2023-04-03
Added capability of importing yaml combinations https://github.com/GrandOrgue/grandorgue/issues/1195

oleg68:
Resolves: #1195

This is the last PR on export/import yaml combination

Now the menu entry ``File->Import Combinations``
1. Open a file selection dialog at Combination/ChurchName directory (where Export Combinations creates files).
2. By default lists the `.yaml` file in this directory, but the user is able to choose the `.cmb` file filter instead.
3. Supports importing combinations from both the new yaml and from the old .cmb (settings) file formats.
4. For importing combinations from the old .cmb files the user has to navigate the file open dialog to the Export/Import (`Settings`) directory.

### DIFF
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -420,13 +420,27 @@ currently in use is <emphasis>deleted</emphasis>.
           <indexterm>
             <primary>Import Combinations</primary>
           </indexterm>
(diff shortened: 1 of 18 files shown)


rousseldenis (approved):
Functional review. Working fine. Great!

larspalo (approved):

PR #1460, opened 2023-04-05
Added the filename to an error message when the file not found

oleg68:
If an ODF or a built-in panel references to an image filename that does not exist, GO displays an error message without a filename, so troubleshooting is quite difficult.

I added the filename to the error message.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1461, opened 2023-04-05
Added new label images for displaying the current combination file name https://github.com/GrandOrgue/grandorgue/issues/1196

oleg68:
This is the first PR related to #1196.

For adding capability of navigating across combination yaml-files, the sequencer panel has to have a label large enough fo displaying the file name. I added that label images in this PR.

No GO behaviur should be changed (unless the new label image numbers 13 and 14 are not used in ODF) .


larspalo:
Honestly, don't you think we should get some new(er) images instead of enlarged versions of the old ones I did way back... It would be really nice to have some label images that tile well, so that the actual label dimension would be possible to set as needed.

larspalo:
What about something like this?
SeamlessLabel1
SeamlessLabel2
The first is somewhat like parchment paper and the second is perhaps towards the green screen types. They should tile fairly well so that different sizes could be specified without the corners creating problems.

oleg68:
@larspalo Yes, I like your images. But a gray one is needed too.

larspalo:
Something like this?
SeamlessLabel3

oleg68:
@larspalo Yes, it is.

Could you make a PR with your labels instead of this one?

larspalo:
I'll do that.

oleg68:
Replaced with #1465.

PR #1467, opened 2023-04-08
Moved notification on combination change from combinations to the controls and the setters

oleg68:
Earlier any combination (either General or Divisional) notified OrganController on change.

That
1. Contradicted the model/controller paradigm, where the model should only return change status, but not to call the controller
2. Did not allow to separate changing in combinations from changing the organ setting

In this PR 
1. I removed calling SetOrganNotified from GOCombination::PushLocal().
2. I introduced the method GOSetter::NotifyCmbPushed() .
3. I aded calling GOSetter::NotifyCmbPushed() to the code where GOCombination::PushLocal() is called from.

### DIFF
--- a/src/grandorgue/combinations/GOSetter.cpp
+++ b/src/grandorgue/combinations/GOSetter.cpp
@@ -435,14 +435,17 @@ void GOSetter::Load(GOConfigReader &cfg) {
     cfg, wxT("SetterGeneralNext"), _("Next"));
 }
 
+void GOSetter::NotifyCmbChanged() {
+  // Temporary we mark the organ modified when a combination is changed for
+  // the user would save the preset.
+  // But we intend to split combinations and organ settings, so in the future
+  // we will only mark the combinations as modified, not the organ settings
+  m_OrganController->SetOrganModified();
+}
+
 void GOSetter::NotifyCmbPushed(bool isChanged) {
-  if (isChanged && IsSetterActive()) {
-    // Temporary we mark the organ modified when a combination is changed for
-    // the user would save the preset.
-    // But we intend to split combinations and organ settings, so in the future
-    // we will only mark the combinations as modified, not the organ settings
-    m_OrganController->SetOrganModified();
-  }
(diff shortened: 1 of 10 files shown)


src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -923,7 +939,7 @@ void GOSetter::Crescendo(int newpos, bool force) {
     else
       m_CrescendoExtraSets[oldIdx].clear();
     ++m_crescendopos;
-    m_crescendo[newIdx]->Push(
+    changed |= m_crescendo[newIdx]->Push(
```
larspalo, line 942: This is in need of some serious comments in my opinion. I do understand the use of bitwise OR to not short circuit the cumulative changes to changed and the need to always have the right hand part executed in this case, but it's more than likely not at all intuitive to someone reading the code to understand why we're using bitwise OR here rather than logical OR.

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -932,9 +948,10 @@ void GOSetter::Crescendo(int newpos, bool force) {
 
     const unsigned newIdx = m_crescendopos + m_crescendobank * CRESCENDO_STEPS;
 
-    m_crescendo[newIdx]->Push(
+    changed |= m_crescendo[newIdx]->Push(
```
larspalo, line 951: Same as above.

larspalo (approved):
Functionally works as it should in my tests. Though, you may consider adding some comments where I made some notations.

src/grandorgue/combinations/GOSetter.cpp
oleg68: I would replace it with `||=` but, unfortunally, this operator does not exist in the c++ syntax. So I replaced it with ``changed = changed || _Expression_``

oleg68:
@larspalo seems I fixed the `Delete` and `Insert` buttons behavior.

rousseldenis (approved):

PR #1468, opened 2023-04-08
Added capability of specifying a size of the label

oleg68:
In order to using lars' label images #1465  in the sequencer panel I need a capability of specifying a label size different from the label size.

It is a capability for future use in #1196. No GO behavior should b changed now.

### DIFF
--- a/src/grandorgue/gui/GOGUILabel.cpp
+++ b/src/grandorgue/gui/GOGUILabel.cpp
@@ -63,6 +63,8 @@ static const wxString WX_BITMAP_LABEL_FMT = wxT(GOBitmapPrefix "label%02d");
 void GOGUILabel::InitBackgroundBitmap(
   unsigned x,
   unsigned y,
+  unsigned w,
+  unsigned h,
   wxString imageFileName,
   unsigned imageNum,
   const wxString &imageMaskFilename) {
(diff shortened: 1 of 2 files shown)


src/grandorgue/gui/GOGUILabel.cpp
```diff
@@ -202,6 +207,22 @@ void GOGUILabel::Load(GOConfigReader &cfg, wxString group) {
...
+    cfg.ReadInteger(
+      ODFSetting,
+      group,
+      wxT("Height"),
+      1,
+      m_metrics->GetScreenHeight(),
+      false,
+      0),
     cfg.ReadStringTrim(ODFSetting, group, WX_IMAGE, false),
     cfg.ReadInteger(ODFSetting, group, WX_DISP_IMAGE_NUM, 0, 12, false, 1),
```
larspalo, line 227: Could you please also change this line to allow up to image 15 instead of only 12?

larspalo (approved):
Ok, but please change on the line I noted too, so that we don't need another PR for that...

src/grandorgue/gui/GOGUILabel.cpp
oleg68: Done

larspalo (approved):
Thanks!

rousseldenis (approved):

PR #1475, opened 2023-04-14
Added capability of nanigating over, loading and saving the YAML combinations files on the Combination Setter panel https://github.com/GrandOrgue/grandorgue/issues/1196

oleg68:
Resolves: #1196

This PR adds capability of navigating over the YAML combination files, loading and saving them in the current combinations directory. This directory by default is the same directory Expoer/Import Combinations menu items work  with. But importing YAML combinations from any other place with the ``File->Import Combinations`` menu switches the current combination directory.

I also enlarged the font of the current setter position as has been asked in #1473.

This is the last PR planned for 3.11.0, so I'm just creating an issue-release request as well.


help/grandorgue.xml
```diff
@@ -7228,6 +7314,25 @@ attributes of a label.
...
+          <term>SetterCurrFileName</term>
+          <listitem>
+            <para>
+	      Represents the current combinations file name of the setter.
+	      The section must contain the GUI attributes of a label.
+	    </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>Refresh, SetterPrevFile, SetterNextFile, SetterLoadCmb, SetterSaveCmb</term>
```
larspalo, line 7327: There was already a Save button on the setter panel since before (that saved to a preset/settings file, right?). Why have you silently removed it from visibility by moving it to the general controls? Since the setter element Save is already available and could potentially have been used in sample sets, I understand that it shouldn't be removed lightly - but now you have a button with the same old name on the setter panel that does something slightly different. That could potentially be somewhat confusing. I think that changing to "Load cmb" and "Save cmb" as texts could be clearer - just as you renamed the old Save to "Save settings".

src/grandorgue/gui/GOGUISequencerPanel.cpp
```diff
@@ -39,106 +43,152 @@ GOGUIPanel *GOGUISequencerPanel::CreateSequencerPanel(GOConfigReader &cfg) {
...
+  PosDisplay->SetFontSize(14);
   panel->AddControl(PosDisplay);
 
+  GOGUILabel *const currFileNameDisplay = new GOGUILabel(
+    panel, m_OrganController->GetLabel(GOSetter::KEY_CURR_FILE_NAME));
+
+  currFileNameDisplay->Init(
+    cfg, GOSetter::GROUP_CURR_FILE_NAME, 170, 58, wxEmptyString, 15, 380, 40);
+  currFileNameDisplay->SetFontSize(14);
+  /*
```
larspalo, line 56: Perhaps remove the  comments that contain dead code?

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -371,6 +401,13 @@ void GOSetter::Load(GOConfigReader &cfg) {
   m_buttons[ID_SETTER_INSERT]->Init(cfg, wxT("SetterInsert"), _("Insert"));
   m_buttons[ID_SETTER_DELETE]->Init(cfg, wxT("SetterDelete"), _("Delete"));
 
+  m_buttons[ID_SETTER_REFRESH]->Init(cfg, GROUP_REFRESH, _("Refresh"));
+  m_buttons[ID_SETTER_PREV_FILE]->Init(cfg, GROUP_PREV_FILE, _("Prev File"));
+  m_buttons[ID_SETTER_NEXT_FILE]->Init(cfg, GROUP_NEXT_FILE, _("Next File"));
+  m_buttons[ID_SETTER_LOAD_CMB]->Init(cfg, GROUP_LOAD_CMB, _("Load"));
```
larspalo, line 407: Load cmb

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -371,6 +401,13 @@ void GOSetter::Load(GOConfigReader &cfg) {
...
+  m_buttons[ID_SETTER_SAVE_CMB]->Init(cfg, GROUP_SAVE_CMB, _("Save"));
```
larspalo, line 408: Save cmb

src/grandorgue/gui/GOGUISequencerPanel.cpp
```diff
@@ -39,106 +43,152 @@ GOGUIPanel *GOGUISequencerPanel::CreateSequencerPanel(GOConfigReader &cfg) {
 
   GOGUILabel *PosDisplay
     = new GOGUILabel(panel, m_OrganController->GetLabel(wxT("Label")));
-  PosDisplay->Init(cfg, wxT("SetterCurrentPosition"), 350, 10);
+  PosDisplay->Init(cfg, wxT("SetterCurrentPosition"), 325, 10);
+  PosDisplay->SetFontSize(14);
```
larspalo, line 47: I'm perfectly ok with this, but in the long run I think we should have another way of exposing the font and font size for both label and button elements so that the user can customize them from within GO.

larspalo (changes requested):
In my tests it works beautifully! I have a few comments about words you might think about changing. If you don't like my suggestion of load and save cmb, at least consider another text that makes their purpose obvious. Also, the symmetry of the setter panel is now a little off, but I don't have any better ideas on how to improve it, so for now I think that the value of your work is far beyond such small issues. Great work!

help/grandorgue.xml
rousseldenis: Even I agree on that, shouldn't we use full term, so "Save combinations"?

larspalo: @rousseldenis Yes, it would be even better, but I think that the space on the drawstop might put a limit on what can be written, but maybe if the words are put on two lines with appropriate textbreakwidth set...

oleg68:
>  the symmetry of the setter panel is now a little off, but I don't have any better ideas on how to improve it, 

Some ideas:
1. To remove the `Refresh` button. I think it will be in use very seldom.
2. To remove the `Save` button, because `File->Export Combinations` does the same thing and even more. But saving from the setter panel only with MIDI migh be more commode.
3. To move some button to the last line where there are some empty space.

help/grandorgue.xml
oleg68: > That could potentially be somewhat confusing. I think that changing to "Load cmb" and "Save cmb" as texts could be clearer - just as you renamed the old Save to "Save settings".

But `cmb` may be confused with the file name extension `.cmb` that is used for organ settings. ``Load yaml`` and ``Save yaml`` would not be understandable for most of non-technical people. Might be ``Load File`` and ``Save File``, because ``Prev File`` and ``Next File`` are present already?

oleg68: > Even if I agree on that, shouldn't we use full term, so "Save combinations"?

That long label wouldn't fit the button size.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Added capability of nanigating over, loading and saving the YAML combinations files on the Combination Setter panel https://github.com/GrandOrgue/grandorgue/issues/1196
```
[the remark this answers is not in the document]
oleg68: Fixed

help/grandorgue.xml
```diff
@@ -7228,6 +7314,25 @@ attributes of a label.
...
+          <listitem>
+            <para>
+	      Refresh/PrevFile/NextFile/Load/Save buttons of the setter. The
+	      section must contain the GUI attributes of a button (displayed as
+	      button per default).
```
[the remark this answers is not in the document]
oleg68: Replaced with `drawstop`

src/grandorgue/gui/GOGUISequencerPanel.cpp
oleg68: Fixed

src/grandorgue/combinations/GOSetter.cpp
oleg68: I changed to "Load file" and "Save file". If you agree, give me a sign, and I will update the screenshot for help

src/grandorgue/combinations/GOSetter.cpp
oleg68: Same as above

src/grandorgue/gui/GOGUISequencerPanel.cpp
oleg68: Is not resizing the whole panel sufficient?

And now it is possible to change it in odf with

```
[SetterCurrentPosition]
FontSize=16
```

src/grandorgue/combinations/GOSetter.cpp
larspalo: @oleg68 That's ok with me, I understand that it's difficult to fit the Load combinations and Save combinations. So if @rousseldenis doesn't have a better idea that fits, lets go with Load file and Save file, but maybe the Id constants should be changed too then.

src/grandorgue/combinations/GOSetter.cpp
larspalo: Sure is!

larspalo:
> Some ideas:
> 
>     1. To remove the `Refresh` button. I think it will be in use very seldom.
> 
>     2. To remove the `Save` button, because `File->Export Combinations` does the same thing and even more. But saving from the setter panel only with MIDI migh be more commode.
> 
>     3. To move some button to the last line where there are some empty space.

For your first point: The times when a refresh might be needed is if the combination lists in current directory change names, new ones are made or when lists are deleted. Maybe, the refresh could be done automatically if any such events occur? How did you plan future management of such operations?

For the second point: Keep that Save file option available by MIDI, if the combinations are not reworked so that no saves need to be actively done (instead the current list always update/save back to disc any time a combination in it changes...). If such a change would be made, then we indeed could remove the save button. Then a New File (or list) button might be more useful.

For the third, it would look really great if the current combination number and the combination list label would be centered on the panel. But maybe we first need to re-think exactly what buttons we want/need on the setter panel.

src/grandorgue/combinations/GOSetter.cpp
oleg68: You are right. I replaced to FILE everywhere

oleg68:
> For your first point: The times when a refresh might be needed is if the combination lists in current directory change names, new ones are made or when lists are deleted. Maybe, the refresh could be done automatically if any such events occur? How did you plan future management of such operations?

I'm not planning any automatically handling file modifications. It is not so actual on organ consoles.

> For the second point: Keep that Save file option available by MIDI, if the combinations are not reworked so that no saves need to be actively done (instead the current list always update/save back to disc any time a combination in it changes...). If such a change would be made, then we indeed could remove the save button. 

Sometimes the errorneous combination changes occur and auto saving them wouldn't be good.

> Then a New File (or list) button might be more useful.

Entering the new file name would be very difficult with midi, so I'd leave this capability to `Export Combinations`.

> For the third, it would look really great if the current combination number and the combination list label would be centered on the panel. But maybe we first need to re-think exactly what buttons we want/need on the setter panel.

For now I'd leave the layout unchanged.

larspalo:
> Sometimes the errorneous combination changes occur and auto saving them wouldn't be good.

Yes, but this is indeed how most real life combination systems work. As soon as a combination is set it is stored back to disc. There is usually no option to save this list, it just written as soon as anything changes. But we need not go into such details right now.

> Entering the new file name would be very difficult with midi, so I'd leave this capability to Export Combinations.

Right, this is good enough for now. But in the future I think we should change/expand the "Export Combinations" to instead be "Manage Combinations", where the user can handle the lists more gracefully so that there wont be a need to use an external file manager for handling list creation/renaming/deleting etc.

help/grandorgue.xml
```diff
@@ -7228,12 +7314,31 @@ attributes of a label.
      </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>SetterCurrFileName</term>
```
larspalo, line 7318: CurrFileName

help/grandorgue.xml
```diff
@@ -7228,12 +7314,31 @@ attributes of a label.
...
+          <listitem>
+            <para>
+	      Represents the current combinations file name of the setter.
+	      The section must contain the GUI attributes of a label.
+	    </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>SetterRefreshFiles, SetterPrevFile, SetterNextFile, SetterLoadFile, SetterSaveFile</term>
```
larspalo, line 7327: RefreshFiles, PrevFile, NextFile, LoadFile, SaveFile

larspalo (changes requested):
The keys to the new setter elements needs to be changed in the help.

help/grandorgue.xml

larspalo (approved):

rousseldenis (approved):

PR #1483, opened 2023-04-20
Update Hungarian translation

kerkovits:
New strings were introduced in recent versions.


larspalo (approved):
I cannot understand Hungarian at all, but every effort to improve the translations has my approval in good faith nevertheless.

oleg68 (approved):
I can not check the translation, so I approve it as is.

Issue #1484, opened 2023-04-20
Number of WindchestGroup objects is limited to 50


eturpault:
The current implementation of GrandOrgue limits the number of WindchestGroup objects at 50.
Since v2.2 OdfEdit converts to GrandOrgue ODF the volume sliders/enclosures of Hauptwerk sample sets. This can generate more than 50 WindchestGroup objects in the GO ODF with some sample sets.
Example : the Dom Bedos sample set of Sonus Paradisi for which 79 WindchestGroup objects are generated by OdfEdit : one slider to tune the volume of each stop, and each stop being divided in two windchests (left and right), plus the tracker volume slider and blower volume slider.
Could the maximum number of WindchestGroup objects be increased to 100 for example ?

eturpault:
With this change I get no more an error for the WindchestGroup objects number of the Dom Bedos sample set, thank you.
However now I get another error (not visible before), it is with the Enclosure objects number which exceeds the maximum value (50) as well with Dom Bedos, there are 58 enclosures built by OdfEdit (one enclosure per stop + enclosures for blower noise volume and tracker noise volume).
@oleg68 could you please also increase to 999 the maximum number of enclosures ? Do you want I open another issue for that ?

PR #1487, opened 2023-04-22
Fixed opening an organ with more than 50 windchest groups https://github.com/GrandOrgue/grandorgue/issues/1484

oleg68:
Resolves: #1484

I increased the windchest limit to 999


larspalo (approved):
Nothing against this per se. However, looking at it from the perspective of future swellbox modeling, it's possible that we'll need to separate pure amplitude adjustment (again - in the future) from a more complex swellbox model since they might be used for completely different things/purposes and both might be desired.

oleg68:

rousseldenis (approved):

Issue #1485, opened 2023-04-20
YOffset attribute in Manual section cannot have negative value


eturpault:
The attributes YOffset_KEYTYPE and Key999YOffset can be set at a value from 0 to 500. Negative Y offset value is not possible.
This doesn't permit to display correctly keyboards sharp keys of some Hauptwerk sample sets.

For example with the Prytanée sample set of Sonus Paradisi : 
keyboard (on the left the expected display, on the right the obtained display with YOffset=0 instead of -10) :  
![comparatif claviers prytanée](https://user-images.githubusercontent.com/51839171/233500695-ae186270-8ab8-487a-b04f-43867c96fd27.png)
pedal (on the left the expected display, on the right the obtained display with YOffset=0 instead of -56) :  
![comparatif pédalier prytanée](https://user-images.githubusercontent.com/51839171/233501648-831d6ebd-4cfc-4cd6-b2d1-7732f98eab3d.png)

Same issue with for example Buckeburg and Noordbroek sample sets of Sonus Paradisi.

Could the YOffset accepted range be extented to negative values : -100 for example ?

eturpault:
Thanks @oleg68 for the implementation, I confirm that the keys can now be all displayed as expected with the sample sets that I tested.

PR #1488, opened 2023-04-22
Added support of negative y-axis adjustment of manual keys https://github.com/GrandOrgue/grandorgue/issues/1485

oleg68:
Resolves: #1485

I allowed to specify negative values for YOffset_KEYTYPE and Key999YOffset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Added support of negative y-axis adjustment of manual keys https://github.com/GrandOrgue/grandorgue/issues/1485
 - Fixed opening an organ with more than 50 windchest groups https://github.com/GrandOrgue/grandorgue/issues/1484
 # 3.11.0 (2023-04-17)
 - Added capability of navigating over, loading and saving the YAML combinations files on the Combination Setter panel https://github.com/GrandOrgue/grandorgue/issues/1196
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -8695,7 +8695,7 @@ sharp key overlapped with the previous key.
           <term>YOffset_<emphasis>KEYTYPE</emphasis></term>
           <listitem>
             <para>
-(integer 0 - 500, default: 0) This value is can be used to adjust the
+(integer -500 - 500, default: 0) This value is can be used to adjust the
 Y coordinate of the current key.
      </para>
           </listitem>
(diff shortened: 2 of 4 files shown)


larspalo (changes requested):
@oleg68 You do realize that the issue in #1485 is caused by not understanding how things can be done in GrandOrgue and being willing to adapt to it? It's happening when someone is just looking at how things are done in hw and expect GO to behave in exactly the same way, when the things already could be done by just doing it the opposite way with the already existing implementation... (offsetting the naturals instead and possibly need to re-calculate some positions from the hw offsets to get the same exact placement). But in the end, as such, I have nothing really against the possibility to adjust the y value of key positioning in the negative direction.

However, what I **really** do want you to change is the phrasing of the title of the PR and the line in changelog so that it simply states that it's to allow negative y-axis adjustment of manual keys. Just so that it's clear that this is not a feature to just imitate Hauptwerk behavior.

eturpault:
In Hauptwerk format, each key is placed in an absolute manner with LeftXPosPixels and TopYPosPixels attributes. There are no offset attributes.
To convert this HW absolute placement in GO format, the TopYPosPixels value of the first key of the keyboard (usually a natural key) can be used to set the PositionY attribute of the Manual, then each next key of the keyboard is placed relatively to this PositionY value with the YOffset attribute.
In order to use only positive YOffset values, it should be necessary to parse the TopYPosPixels values of all the keys of the keyboard to memorize the lowest Y value, then use it to set PositionY and parse again the keys of the keyboard to place them relatively to PositionY.
The first approach is more simple to implement and faster to execute that the second one.

larspalo:
> The first approach is more simple to implement and faster to execute that the second one.

Totally beside the point, and for a computer/app of **very little** concern. The point is: the desired effect was already possible to achieve, you just didn't want to do it that way! That said, I'm not at all against implementing negative y-axis adjustments as a feature in GO, in this case just the phrasing of the motivation for it.

eturpault:
The change being simple to make in GO, yes I would prefer it to be done in GO rather than in OdfEdit. And it is good to take advantage of this kind of opportunity to enlarge the configuration possibilities of the GO ODF format.

oleg68:
> owever, what I **really** do want you to change is the phrasing of the title of the PR and the line in changelog so that it simply states that it's to allow negative y-axis adjustment of manual keys

Fixed

larspalo (changes requested):
The display works perfectly, however the mouse rectangle seems to not really have changed from the default position to the new displayed position with this PR (whereas with the older positive YOffset it is). Thus, when clicking on the offset position or above on the bitmap, nothing happens. Also, the displayed keychange for the on state only updates to the not offset default position. Perhaps check the bounding box calculations?

larspalo:
[TestYOffsetManuals.zip](https://github.com/GrandOrgue/grandorgue/files/11311802/TestYOffsetManuals.zip)

oleg68:
@larspalo I fixed processing mouse events

larspalo (approved):
Now it works as expected!

oleg68:

rousseldenis (approved):

Issue #1489, opened 2023-04-23
YAML combinations reset default couplers


kerkovits:
The Niederzwehren sampleset at http://binauralpipes.com/Niederzwehren/Kassel-Niederzwehren.zip has couplers between the first manual and the other two. As this is a coupling manual without own stops, the couplers are always engaged, and are not visible to the user.

I tried exporting YAML combinations, and imported that later on another computer. The hidden couplers became disengaged, and the lower manual produced no sound. The issue is not present if I save a CMB, and load it with 'import combinations'.

oleg68:
@larspalo 

> The switches are listed in the [Manual999] section so they should definitely be stored in the divisionals. Also, I think that the visible switches should be the ones controlling the back end stops even if the stops themselves are not displayed or stored in the combinations - they should activate as they are controlled by the switches nevertheless.
> To sum it up: we can expect that since Norrfjärden has the CombinationsStoreNonDisplayedDrawstops=N in the [Organ] section, the combinations should not store any of the non displayed drawstops, but all the displayed drawstops (switches) should really be stored. But to be fair, this "bug/feature" has existed for a long time as I tested a very old GO build and it shows exactly the same behavior.

We should specify a precise definition, what `Displayed` mean? Displayed where? On the main panel only? Or on any other panels?

Why should we assume that any switch mentioned in  [Manual999] section is displayed? It is not obvious.

larspalo:
@oleg68 The Displayed for the structural elements of an organ is a relic from old times where the main panel was specified in the structural section. With the new panel format the main panel is [Panel000] and all structural elements default to a displayed false - but are displayed as needed on [Panel000]. So, in the old format they are not displayed - but in the new format they are displayed as they exist on [Panel000].

The reason why it's this complicated is because of keeping compatibility with older versions of specification. Also the current implementation is not fully adapted to deal with the new way of separating the displayed items as anything that is defined on the panels. The structural section is not intended to be used for any kind of display related stuff in the new format.

Any logic checking if a drawstop is displayed or not must take into account multiple things. The simple case is if Displayed=Y in the structural section like in the old format, and this works since ancient times. For the new format GO should check all of the panels (or at the very least [Panel000]) and see if an element is displayed there, if so, then they should be considered displayed even if the value of Displayed is false in the structural section.

[...]

PR #1492, opened 2023-04-28
Fixed processing yaml combinations on organs with invisible couplers https://github.com/GrandOrgue/grandorgue/issues/1489

oleg68:
Organ settings have three possible states for any combination elements: 1 - enabled, 0 - disabled and -1 - not to touch. But yaml combination file format doesn't distinguish between 0 and -1. So earlier after loading from a yaml combination files all invisible elements were 0 (disabled).

After yhis PR GO sets 0 for elements not mentioned in the yaml file only if they are visible.


larspalo:
I don't remember if I've checked if you implemented the Setter element Full option with the new combinations. Normally we have the 1, 0 or -1 states from odf, but in GO we also have the Full setter element that, when enabled, will save every element regardless of odf settings. Did you take that into consideration for the new yaml combinations?

oleg68:
@larspalo Yes, setting combinations with `Full` is a workaround of this bug when the invisible element is enabled. But when this element is disabled, it is still not saved in yaml combinations, so after restoring it's state still won't be changed. 

I'd leave this inact because I'm not sure that such cases exist.

larspalo (commented):
When testing this PR with the build from your repo I cannot set any combinations at all in GO with a sample set like Norrfjärden. No combination is set nor saved. With simpler sample sets without any invisible couplers it works fine.

oleg68:
> I cannot set any combinations at all in GO with a sample set like Norrfjärden. No combination is set nor saved

Seems there was the same issue before this PR. Have you tested 3.11.0-1?

oleg68:
Having done more tests with Norrfjärden, I noticed, that now switches are not saved in the combinations unless `Full` is engaded. I created a separate issue #1498.

Talking about this PR, there is still some inconsistance of yaml combinations with `Full` behavior. If you save a combination with Full engaged, store it in a yaml file, engage more stops (switches), then reload the saved combination from yaml, then extra stops won't disabled.

A possible solution would be storing a flag `SavedAsFull` in the GOCombination and writing it to yaml. Then, on import the yaml file, regarding all absent elements as `0` if they are visible (as presents in this PR) **or if `SavedAsFull` is set** (is to be done).

@larspalo do you agree and insist on this implementation or do you think that current PR would enough?

larspalo:
> So the reason is that all visible switches have StoreInGeneral=false and StoreInDivisional=false. It is more an ODF-related issue rather a GO-combination system related.

Ok, so the sample set has some limitations imposed on what is saved in the combinations, but such settings as you quote above doesn't really exist. I can ease up the current limitations in next update of the sample set. But for now, what seems to be a bit problematic is that even the displayed drawstops (switches) are not saved in the combinations. Does that seem right to you? The switches are listed in the [Manual999] section so they should definitely be stored in the divisionals. Also, I think that the visible switches should be the ones controlling the back end stops even if the stops themselves are not displayed or stored in the combinations - they should activate as they are controlled by the switches nevertheless.

To sum it up: we can expect that since Norrfjärden has the CombinationsStoreNonDisplayedDrawstops=N in the [Organ] section, the combinations should not store any of the non displayed drawstops, but all the displayed drawstops (switches) should really be stored. But to be fair, this "bug/feature" has existed for a long time as I tested a very old GO build and it shows exactly the same behavior.

Regarding this PR, I think we might as well just move on with this and fix the issues/inconsistencies that creep up as new PRs, do you agree with that?

larspalo:
@oleg68 I think I get it now. Since the [Switch999] doesn't have the Displayed property set, it defaults to N, and the current implementation fails to check that the element (switch) actually **is** displayed on the main panel according to the new format.

oleg68:
> But for now, what seems to be a bit problematic is that even the displayed drawstops (switches) are not saved in the combinations. Does that seem right to you? The switches are listed in the [Manual999] section so they should definitely be stored in the divisionals. Also, I think that the visible switches should be the ones controlling the back end stops even if the stops themselves are not displayed or stored in the combinations - they should activate as they are controlled by the switches nevertheless.

I agree with your conclusions, but it is a subject of #1498,. Lets continue this discusion there.

> Regarding this PR, I think we might as well just move on with this and fix the issues/inconsistencies that creep up as new PRs, do you agree with that?

Before this PR Norrfjärden combinations could be successfully saved in yaml and loaded again with `Full` engaged. Affer this PR Niederzwehren combinations start being saved even without `Full`, but saving Norrfjärden with `Full` has been broken.

You haven't answer my question, whether we should restore the `Full` behavior, so I assume `Yes` and I'm going to do it. If you disagree, put a comment here.

oleg68:
@larspalo I've finished reworking the saving/restoring combination with `Full` engaged as I descrived before. Please check the current implementation now.

Saving Norrfjärden combinations will be fixed in #1498. A workaround of saving them with `Full` engaged should work after this PR.

larspalo (approved):
As far as I've tested it, this PR works fine now. With full enabled everything is saved correctly and can be restored from the combination file. Small things to think about is that GO will not have Full as active by default so if one wants to change/add combinations saved with Full later one need to enable it again - but this is kind of expected even if it might confuse someone that forgets it. Also I noted that the built in divisional buttons are not reset when new combinations are loaded, but that is of course another issue.

rousseldenis (approved):

PR #1517, opened 2023-05-08
Removed usage of GOOrganController from GOButtonControl and derivatives

oleg68:
This PR is a big step towards removing usage of GOOrganController.

It switches GOButtonControl to use GOOrganModel instead of GOOrganController.

Because there are a lot of direct and indirect subclasses of GOButtonControl, it allowed to stop using  GOOrganController there.


larspalo:
@oleg68 There are two very similar branches: refactor/button-model and refactor/button-model-origin with the same title. Which one did you really intend?

oleg68:
> @oleg68 There are two very similar branches: refactor/button-model and refactor/button-model-origin with the same title. Which one did you really intend?

I deleted `refactor/button-model-origin`.

larspalo (approved):

rousseldenis (approved):

PR #1527, opened 2023-05-14
Removed usage of GOOrganController from combinations/model

oleg68:

### DIFF
--- a/ide-projects/NetBeans12/nbproject/configurations.xml
+++ b/ide-projects/NetBeans12/nbproject/configurations.xml
@@ -693,41 +693,6 @@
             tool="1"
             flavor2="8">
         <ccTool flags="3">
-          <incDir>
-            <pElem>../../src/core</pElem>
-            <pElem>/usr/lib64/wx/include/gtk3-unicode-3.0</pElem>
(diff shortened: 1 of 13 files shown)


larspalo (approved):
Works as far as I've been able to test it (with a local build). However, I do note a difference in behavior between the builtin divisionals and the odf defined divisionals when a odf general, built in general (or a sequencer setter combination) is pushed. The builtin divisional is still visibly active whereas the odf defined divisional will be disabled (goes to not active state). When switching between the builtin divisionals and the odf divisionals both are still visibly active all the time. Is this something intended, or what?

oleg68:
@larspalo 

> I do note a difference in behavior between the builtin divisionals and the odf defined divisionals when a odf general, built in general (or a sequencer setter combination) is pushed. The builtin divisional is still visibly active whereas the odf defined divisional will be disabled (goes to not active state). When switching between the builtin divisionals and the odf divisionals both are still visibly active all the time. Is this something intended, or what?

It is not intended behaviour, but it was also before this PR. Please create a separate issue.

oleg68:

rousseldenis (approved):

Issue #1419, opened 2023-02-27
RPM warnings (missing manpages, unused dependencies)


kerkovits:
I tried to create a custom build of GrandOrgue. The RPM package builder has some warnings (though the build seems to work correctly):
```
grandorgue.x86_64: W: unused-direct-shlib-dependency /usr/bin/GrandOrguePerfTest /usr/bin/../lib64/librtaudio.so.6
grandorgue.x86_64: W: unused-direct-shlib-dependency /usr/bin/GrandOrguePerfTest /usr/bin/../lib64/librtmidi.so.5
grandorgue.x86_64: W: unused-direct-shlib-dependency /usr/bin/GrandOrguePerfTest /usr/bin/../lib64/libjack.so.0
grandorgue.x86_64: W: unused-direct-shlib-dependency /usr/bin/GrandOrguePerfTest /usr/bin/../lib64/libportaudio.so.2
grandorgue.x86_64: W: no-manual-page-for-binary GrandOrguePerfTest
grandorgue.x86_64: W: no-manual-page-for-binary GrandOrgueTool
```
The last two lines indicate the need for `man` pages for the two command-line utilities. I think that it would be useful to include them, because even a regular user knows next to nothing about these utilities.

I have absolutely no idea what the first four lines about unused dependencies mean.

oleg68:
How do you run the RPM package builder?

I do it with ``cpack -G RPM`` and I cann't see these warnings even with verbose running:

```
[oleg@oleg2 current]$ cpack -G RPM -V
CPack: Enable Verbose
CPack Verbose: Read CPack config file: 
CPack Verbose: Read CPack configuration file: /home/oleg/my-projects/grandorgue/GrandOrgue/build/current/CPackConfig.cmake
CPack Verbose: Specified generator: RPM
CPack Verbose: Use generator: cmCPackRPMGenerator
CPack Verbose: For project: grandorgue
CPack: Create package using RPM
CPack Verbose: [RPM] requested component grouping = ONE_PER_GROUP
CPack Verbose: Remove toplevel directory: /home/oleg/my-projects/grandorgue/GrandOrgue/build/current/_CPack_Packages/linux/RPM
CPack: Install projects
CPack: - Run preinstall target for: GrandOrgue
CPack: - Install project: GrandOrgue []
CPack Verbose: Install configuration: "Debug"
[...]

PR #1535, opened 2023-05-21
Added GrandOrgueTool and GrandOrguePerfTest man pages for linux https://github.com/GrandOrgue/grandorgue/issues/1419

oleg68:
I added man pages for two additional command line utilities: `GrandOrgueTool` and `GrandOrguePerfTest`. Also I renamed the source of the GrandOrgue man page.

### DIFF
--- a/resources/CMakeLists.txt
+++ b/resources/CMakeLists.txt
@@ -49,6 +49,13 @@ endforeach()
 # ========================
 # Install resources
 # ========================
+function(add_man_page EXECUTABLE_NAME)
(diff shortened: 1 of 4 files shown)


larspalo (commented):
In all three of the *Man.xml files there's an email reference to the old developers mailing list on Sourceforge. This is not really appropriate anymore, I think. However, I'm not certain what alternative email address we should use...

oleg68:
> In all three of the *Man.xml files there's an email reference to the old developers mailing list on Sourceforge. This is not really appropriate anymore, I think. However, I'm not certain what alternative email address we should use...

I spent some time for research, whether we can specify the Github URL instead of the email. The result is *no*, we can't.

Maybe it is better to remove the email at all?

larspalo:
> Maybe it is better to remove the email at all?

If an email address is not mandatory, then I think we can remove it until we have a good alternative solution.

larspalo (approved):

rousseldenis (approved):

Issue #1550, opened 2023-05-27
MacOS + sound card usb = invalid number of channels 


dewal5:
I had no problem with this card on PC.

It's not only this specific sound card. I tried two others and give me the same problem/error. 
"Open of the audio stream for Pa: Core Audio: USB Audio CODEC:
failed: Invalid numbers of channels."
"the device does not support the requested channel count"

I change channels with no result. 

On built speakers/card works great. 
On built in monitor/screen works great

oleg68:
@dewal5 

1. Download and install the last version of GrandOrgue.
2. Post the File->Settings->Audio screenshot here.
3. Try to use `RtAudio` instead of `PortAudio`.
4. Post here the exact error message from the Log window.

dewal5:
Hello 
I've the last version of GrandOrgue ✅
Set to RtAudio ✅
RtAudio doesn't work either. Believe me I tried almost everything 

![Zrzut ekranu 2023-05-31 o 09 20 15](https://github.com/GrandOrgue/grandorgue/assets/93423039/721b40b4-d480-4674-83f1-5dbe05f7e569)
![Zrzut ekranu 2023-05-31 o 09 20 10](https://github.com/GrandOrgue/grandorgue/assets/93423039/eeab099a-22f5-45c7-9c4d-cbc3e4253ffa)
![Zrzut ekranu 2023-05-31 o 09 19 56](https://github.com/GrandOrgue/grandorgue/assets/93423039/06dd924c-aecb-40ce-b882-b62a236dda6b)

![Zrzut ekranu 2023-05-31 o 09 22 25](https://github.com/GrandOrgue/grandorgue/assets/93423039/a14327da-86b5-44a6-8043-c1ab9879ec54)

dewal5:
Any idea?

oleg68:
This error is rased inside CoreAudio, so it is the same with RtAudio and PortAudio.

The reason is GrandOrgue tries to open the audio device with two channels (as they are confirured in Audio Settings), but your usb audio interface has a different number of channels.

What audio device do you have?

dewal5:
Behringer uca200. Also with H2n zoom using as sound card

oleg68:
Try to set Sample Rate to 48000.

dewal5:
I've already tried

oleg68:
A better workaround is to install `qjackctl`, setup and start jack and use 'Jack: Native Output'.

oleg68:
[...]

PR #1569, opened 2023-06-08
Fixed "The devive does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550

oleg68:
Resolves: #1550

The problem was caused by duplicating device name in MacOS for USB audio devices

This PR
- Adds checking for the output channel count when matching devices by name
- Eliminates double searching a device by name: earlier both create and Open were searching the devices, now create searches the device and remember it's number that is used by Open

Both `GOSoundPortaudioPort` and `GOSoundRtPort` are taken in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Fixed "The devive does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550
+- Fixed "The device does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550
 # 3.12.1 (2023-06-06)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Fixed "The devive does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550
```
rousseldenis, line 1:
```suggestion
- Fixed "The device does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550
```

rousseldenis (approved):
Minor comment

larspalo (approved):

PR #1576, opened 2023-06-13

oleg68:

It is jost refactoring. No GO controller should be changed.

### DIFF
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -211,6 +211,7 @@ GOProperties.cpp
 GOFrame.cpp
 GODocument.cpp
 GOPanelView.cpp
+GOVirtualCouplerController.cpp
 )
 
 if (USE_INTERNAL_ZITACONVOLVER)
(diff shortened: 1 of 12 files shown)


larspalo:
When I test this PR I get lots of errors in the log about faulty combination posts that all relate to Couplers. Even though the errors are logged, the combinations seem to work anyway.

oleg68:
@larspalo I managed to reproduce the log messages and fixed two (old) bugs. But the extra log messages still appear when loading old presets. The reason was a reference to not-existent couplers from the preset.

After saving a new preset the errors disappear.

oleg68:
@larspalo The error messages you mentioned relates to the virtual couplers, that do not exist in ODF and are provided by the `Couplers` panel.

Now the virtual couplers are created *after* initialisation of General and Divisional templates, so this PR breaks saving virtual coupler states in the combinations. It need more works to restore this capability.

oleg68:
For fixing saving virtual couplers I need at first merging #1576

PR #1581, opened 2023-06-19

oleg68:
This is a new PR instead of #1576

It is just refactoring. No GO controller should be changed.

After merging #1580 the capability of saving virtual coupler states are saved correctly with this PR.


oleg68:
Wrong branch

Issue #1498, opened 2023-04-30
GrandOrgue does not store visible switches in combinations by default


oleg68:
I cannot set any combinations at all in GO with a sample set like Norrfjärden. No combination is set nor saved. With simpler sample sets without any invisible couplers it works fine.

_Originally posted by @larspalo in https://github.com/GrandOrgue/grandorgue/pull/1492#pullrequestreview-1407021965_

oleg68:
Storing switches in combinations is controlled by GOSwitch::IsDisplayed(). Seems that this method returns a wrong result for Norrfjärden.

oleg68:
@larspalo 

> The switches are listed in the [Manual999] section so they should definitely be stored in the divisionals. Also, I think that the visible switches should be the ones controlling the back end stops even if the stops themselves are not displayed or stored in the combinations - they should activate as they are controlled by the switches nevertheless.
> To sum it up: we can expect that since Norrfjärden has the CombinationsStoreNonDisplayedDrawstops=N in the [Organ] section, the combinations should not store any of the non displayed drawstops, but all the displayed drawstops (switches) should really be stored. But to be fair, this "bug/feature" has existed for a long time as I tested a very old GO build and it shows exactly the same behavior.

We should specify a precise definition, what `Displayed` mean? Displayed where? On the main panel only? Or on any other panels?

Why should we assume that any switch mentioned in  [Manual999] section is displayed? It is not obvious.

oleg68:
> @oleg68 The Displayed for the structural elements of an organ is a relic from old times where the main panel was specified in the structural section. With the new panel format the main panel is [Panel000] and all structural elements default to a displayed false - but are displayed as needed on [Panel000]. So, in the old format they are not displayed - but in the new format they are displayed as they exist on [Panel000].

> The reason why it's this complicated is because of keeping compatibility with older versions of specification. Also the current implementation is not fully adapted to deal with the new way of separating the displayed items as anything that is defined on the panels. The structural section is not intended to be used for any kind of display related stuff in the new format.

Any logic checking if a drawstop is displayed or not must take into account multiple things. The simple case is if Displayed=Y in the structural section like in the old format, and this works since ancient times. For the new format GO should check all of the panels (or at the very least [Panel000]) and see if an element is displayed there, if so, then they should be considered displayed even if the value of Displayed is false in the structural section.

[...]

PR #1582, opened 2023-06-19

oleg68:
This is a new PR instead of https://github.com/GrandOrgue/grandorgue/pull/1576

It is just refactoring. No GO controller should be changed.

After merging https://github.com/GrandOrgue/grandorgue/pull/1580 the capability of saving virtual coupler states are saved correctly with this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Changed the default value of the CombinationsStoreNonDisplayedDrawstops ODF settings to false
 - Fixed capability of running on MacOs 11.3
 - Fixed "The device does not support requested channel count" error when using an USB audio card on MacOS https://github.com/GrandOrgue/grandorgue/issues/1550
 # 3.12.1 (2023-06-06)
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -7150,10 +7150,10 @@ state of divisional couplers.
           <term>CombinationsStoreNonDisplayedDrawstops</term>
           <listitem>
             <para>
-(boolean, default: true) determines, if the state of 
-invisible objects (on the main panel) is stored in divisionals,
-generals and the setter.
-     </para>
+	      (boolean, default: true) determines, if the state of read-only
+	      objects (that is calculated based on the state of another objects)
+	      is stored in divisionals, generals and the setter.
+	    </para>
           </listitem>
         </varlistentry>
         <varlistentry>
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -7150,7 +7150,7 @@ state of divisional couplers.
           <term>CombinationsStoreNonDisplayedDrawstops</term>
           <listitem>
             <para>
-	      (boolean, default: true) determines, if the state of read-only
+	      (boolean, default: false) determines, if the state of read-only
 	      objects (that is calculated based on the state of another objects)
 	      is stored in divisionals, generals and the setter.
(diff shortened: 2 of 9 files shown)


larspalo (changes requested):
With this PR the CombinationStoreNonDisplayedDrawstops suddenly becomes necessary as it won't default to true as before. I suggest that you change it back to default to true instead of being a mandatory line in the odf. There might be more issues, but this is the first I immediately found.

oleg68:
@larspalo 

> With this PR the CombinationStoreNonDisplayedDrawstops suddenly becomes necessary as it won't default to true as before. I suggest that you change it back to default to true instead of being a mandatory line in the odf.

At #1498 the semantics of `CombinationsStoreNonDisplayedDrawstops` was changed totally.

Now it relates to being read-only instead of the visibility. Earlier there were problems with determining visibility, but now there are no more issues with it.

After fixed the problem with storing read-olny elements at https://github.com/GrandOrgue/grandorgue/pull/1582/files#diff-e96a2e520f7cdeb61d46c74147ca677d9f681b3590651d84a5cd82b912d230fe the whole set of rad-only drawstops started spaming in yaml, so I switched it's default to `false`.

I changed this default to `false` because n

Do you know an example of ODF where CombinationsStoreNonDisplayedDrawstops=Y is really necessary? Otherwise I'd leave the new default.

larspalo:
The new PR breaks among others Burea Church and Norrfjärden Church sample sets. The thing is that this PR doesn't just change the default value, it makes the line necessary to include in the odf, whereas earlier one could omit it and trust the default value to kick in. I suggest to just revert the default value to true and keep the CombinationStoreNonDisplayedDrawstops as not necessary to have included.

Personally, I'd even say that it's not even necessary to keep CombinationStoreNonDisplayedDrawstops as a parameter in the odf (should be depracated and kept only for legacy reasons) since the displayed property is no longer the deciding factor - the read only property is now taking its place.

Anyway, the important thing is that it should not be needed to include in the odf, since that does indeed change GO behavior.

oleg68:
> I suggest to just revert the default value to true and keep the CombinationStoreNonDisplayedDrawstops as not necessary to have included.

I marked this parameter as not mandatory, but kept the default value as `false`.

larspalo:
@oleg68 Yes, that works for me. Only thing left then is that the help should also be updated to state that the default value is false now like so:

```
CombinationsStoreNonDisplayedDrawstops
(boolean, default: false) determines, if the state of invisible objects (on the main panel) is stored in divisionals, generals and the setter. 
```

Also, perhaps a note should be added that this is actually only used for old style panel format .organ files (if even that) since that mechanism has been changed to actually work with how the new panel format does things. I leave that up to you though.

oleg68:
> @oleg68 Yes, that works for me. Only thing left then is that the help should also be updated to state that the default value is false now like so:
> 
> ```
> CombinationsStoreNonDisplayedDrawstops
> (boolean, default: false) determines, if the state of invisible objects (on the main panel) is stored in divisionals, generals and the setter. 
> ```

Fixed.

> Also, perhaps a note should be added that this is actually only used for old style panel format .organ files (if even that) since that mechanism has been changed to actually work with how the new panel format does things.

Won't fix, because `CombinationsStoreNonDisplayedDrawstops` behavior does not depend on the panel format style.

larspalo (approved):

rousseldenis (approved):

Issue #1416, opened 2023-02-26
Organ Setting dialog does not have the Help button


oleg68:
There is a lot of information about of Organ Settings in the help, but unlike other dialogs, it doesn't provide a simple way for navigating with the `Help` button.

PR #1591, opened 2023-07-04
Fixed absence of the Help button on the Organ Setting dialog https://github.com/GrandOrgue/grandorgue/issues/1416

oleg68:
Resolves: #1416

This PR adds the `Help` capability to the Organ Settings Dialog.

For this GOOrganDialog is made as a subclass of `GODialog`. An intermediate subclass `GOSimpleDialog` (in opposit to `GOTabbedDialog`) has been added.

Using a common base `GODialog` allowed to remove a lot of code from `GOOrganDialog` itself.


rousseldenis (approved):

larspalo (approved):
Ok, it works, but at least on Linux, the help window is actually behind the organ settings - which makes it kind of less useful. It would be better if the style of the help was to be always on top.

Issue #1415, opened 2023-02-26
Organ Settings dialog takes the whole height of the screen.


oleg68:
When I open Audio/Midi -> Organ Setting, the dialog spans the whole screen vertically, and it can not be decreased.

ClarkeBrunt:
Using v3.10.0-1, I'm not finding that Audio/Midi -> Organ Setting spans whole screen vertically. It's rather small vertically for me, e.g. just showing first 4 items in the tree for the demo organ.

Whereas Audio/Midi -> MIDI Objects dialog is rather large vertically - on one of my screens, it always comes up with the buttons at the bottom off the bottom edge of the screen.

Neither of these dialogs seems to remember size/position. I can resize and move around, but if they are closed and re-opened, they go back to the default.

oleg68:
The issue is on linux

PR #1594, opened 2023-07-07
Fixed size of the Organ Settings Dialog https://github.com/GrandOrgue/grandorgue/issues/1415

oleg68:
Resolves: #1415

I've reworked the layout of the Organ Settings dialog in this PR
1. The size of the Object Tree is not more taken in account when determining the minimal size.
2. The Object tree is now expanded only by one level (windchests) instead of the full expanding when the dialog is open.
3. I removed the ``Collapse Tree`` button because it was needed only as a workaround of the full expanding.
4. I alligned all the buttons in one line.
5. The initial size of the dialog became much less.
6. I removed the explicit Scroll Bar because it is not present in other dialogs (ex. Settings). But I added the `SetLayoutAdaptationMode` call for automatic adding scrollbars of low resolution screens.


larspalo:
Ok, so two things I'm concerned about:

1. The width of the tree is way too small to really be useful. One will need to increase the window size to see anything relevant in it. I suggest that the proportions should be changed so that the tree will occupy, at minimum, equal amount of space as the rest of the dialog by default and then if the window size is increased get all the increased size.
2. The help is still buried under the dialog so that it's basically useless on Linux. The help should be on top if it's activated.

Besides that, I don't have any more objections for now (but in the future, we might need to incorporate many more features in this dialog that doesn't exist yet, which will change the size requirements considerably).

oleg68:
> 1. The width of the tree is way too small to really be useful. One will need to increase the window size to see anything relevant in it. I suggest that the proportions should be changed so that the tree will occupy, at minimum, equal amount of space as the rest of the dialog by default and then if the window size is increased get all the increased size.

Now the window can be resized by the user and the whole of the increased size goes to the tree. Do you want
1. To make the initial size of the tree some wider?
2. To resize the right side of the dialog as well of the object tree? I'd not make it because there is no any "resizable" elements there.

> 2. The help is still buried under the dialog so that it's basically useless on Linux. The help should be on top if it's activated.

Yes, I'm discovering this issue now.

> Besides that, I don't have any more objections for now (but in the future, we might need to incorporate many more features in this dialog that doesn't exist yet, which will change the size requirements considerably).

May be we'll make this dialog tabbed in the future?

larspalo:
1. Yes, only the left side (tree should be wider), I suggest equally much as the right side initially and exactly the same behavior as presently when the window is enlarged.
2. No. The right side is perfect as it is.

Great if you fix the help window too. I've not tested it with a more recent wxWidgets version though. If it works fine out of the box on a more recent wxWidgets, I think you could leave it as it is. If 3.2.2+ also exhibit the issue it needs to be fixed.

Possibly it could be tabbed. What I've got in mind is en equalizer type of intonation control that we possibly could use both for individual pipe intonation (bass, middle, treble controls + their cutoff frequencies) and perhaps utilize the same mechanisms for enclosures on the windchests. But this is of course for other PRs...

oleg68:
@larspalo I managed to fix the covering  the help window

oleg68:
> 1. I suggest equally much as the right side initially

I enlarged the object tree by something, but not as large as the right side. I'd evit making the minimal size of the dialog too large.

larspalo:
Ok, that size is at least a decent starting point for the tree. The only thing I did note were some warnings that doesn't seem so critical as their words imply...

```
(GrandOrgue:100359): Gtk-WARNING **: 10:05:28.410: Attempting to add a widget with type GtkBox to a GtkWindow, but as a GtkBin subclass a GtkWindow can only contain one widget at a time; it already contains a widget of type GtkBox

(GrandOrgue:100359): Gtk-CRITICAL **: 10:05:28.454: gtk_cairo_should_draw_window: assertion 'GDK_IS_WINDOW (window)' failed
```

I didn't note anything wrong with the behaviour of the dialog layout when I re-sized it, so if you're not concerned with the warning above, I'll approve.

oleg68:
> (GrandOrgue:100359): Gtk-WARNING **: 10:05:28.410: Attempting to add a widget with type GtkBox to a GtkWindow, but as a GtkBin subclass a GtkWindow can only contain one widget at a time; it already contains a widget of type GtkBox
> 
> (GrandOrgue:100359): Gtk-CRITICAL **: 10:05:28.454: gtk_cairo_should_draw_window: assertion 'GDK_IS_WINDOW (window)' failed
> ```

It is as a side effect of the changehttps://github.com/GrandOrgue/grandorgue/pull/1594/commits/ab0bbf681a1770ff8039139b48c761e5914e3358

Without calling `Create()` there is no gtk criticals, but the help is covered by the dialog.

I'll spend some more time for research.

oleg68:
@larspalo I managed to eliminate gtk criticals.

For this reproduced the `two step dialog construction` in `GOSimpleDialog` like thete is in `wxPropertySheetDialog`. It allows the dialog not to cover the help window without any gtk criticals.

I'm not going to add more changes in this PR.

src/grandorgue/dialogs/GOOrganDialog.cpp
```diff
@@ -336,49 +343,29 @@ GOOrganDialog::GOOrganDialog(
...
-  scroll->SetScrollbars(0, 5, 0, 15);
-  mainSizer->Add(scroll, 1, wxEXPAND);
+  LayoutWithInnerSizer(mainSizer);
+}
 
+bool GOOrganDialog::TransferDataToWindow() {
   FillTree();
-  Load();
-  LayoutWithInnerSizer(mainSizer);
+  // Load();
```
larspalo, line 367: Is this commended code intended to be used?

src/grandorgue/dialogs/GOOrganDialog.cpp
```diff
@@ -840,12 +828,16 @@ void GOOrganDialog::FillTree(wxTreeItemId parent, GOPipeConfigNode &config) {
     e = m_Tree->AppendItem(parent, config.GetName(), -1, -1, data);
   for (unsigned i = 0; i < config.GetChildCount(); i++)
     FillTree(e, *config.GetChild(i));
-  m_Tree->Expand(e);
+  // m_Tree->Expand(e);
```
larspalo, line 831: Is the commented code expected to be used?

larspalo (commented):
When the organ dialog is opened here on Ubuntu 20.04 with the build from your repo it's not sized at all to begin with. See screenshot. I can adjust the size manually but I also see that the ok buttons are displayed under the text- and spinctrls until the size is large enough. (second screenshot)
Skärmbild från 2023-07-20 19-40-23
Skärmbild från 2023-07-20 19-52-00

oleg68:
@larspalo I tested with wxwidgets 3.2, and everything looks all right. May be it does not work with wx 3.0. I'll do more testing.

oleg68:
@larspalo I added a call to `GetSizer()->SetSizeHints(this);` and I tested on ubuntu 20.04:

Screenshot from 2023-07-20 22-22-58

Could you test [the build](https://github.com/oleg68/GrandOrgue-official/actions/runs/5615156874) on your system?

src/grandorgue/dialogs/GOOrganDialog.cpp
oleg68: No. FillTree calls m_Tree->SelectItem() that calls Load(). 

I removed the line.

src/grandorgue/dialogs/GOOrganDialog.cpp
oleg68: No. It was before, but I removed it.

larspalo (approved):
Now it works fine with initial size, help window and proportions! I like it. Thanks for your effort!

oleg68:
@larspalo the difference in the dialog size between wx 3.0 and 3.2 might be relate to #1600: 3.2 automatically adjusts the dialog size basen on the element sizes, but 3.0 requires an explicit call to `SetSizeHints()`.

larspalo:
@oleg68 Eventually, with time, wx3.0 will be obsolete, but for now I think it's great that the code can work fine on both 3.0 and 3.2!

oleg68:

rousseldenis (approved):

PR #1606, opened 2023-07-22
Fix GCC 12 compiler warnings

Garfield96:
This PR fixes several compiler warnings generated by GCC 12 similar too:
```
src/grandorgue/sound/GOSoundAudioSection.cpp:261:14: warning: array subscript <unknown> is outside array bounds of ‘stereoSample’ {aka ‘GOInt24 [0][2]’} [-Warray-bounds]
  261 |       + input[pos + 1][0]
      |         ~~~~~^
```

### DIFF
--- a/src/grandorgue/sound/GOSoundAudioSection.cpp
+++ b/src/grandorgue/sound/GOSoundAudioSection.cpp
@@ -242,7 +242,7 @@ inline void GOAudioSection::MonoUncompressedLinear(
 template <class T>
 inline void GOAudioSection::StereoUncompressedLinear(
   audio_section_stream *stream, float *output, unsigned int n_blocks) {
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Fix GCC 12 compiler warnings
```
oleg68, line 1: Please remove this entry because it is not interesting end users.

oleg68 (changes requested):
Thus change should not be listed in CHANGELOG

CHANGELOG.md
Garfield96: I removed it.

oleg68 (approved):

PR #1607, opened 2023-07-22
Improve concurrency handling

Garfield96:
With C++11, `volatile` should no longer be used for synchronization, and atomics should be used instead. Therefore, I replaced `volatile` in most places. In addition, I made some variables atomic, which didn't use `volatile` before but were altered by different threads, which could have led to undefined behavior. In two places a condition variable was used without checking the condition in a loop. This can be problematic during spurious wakeups, where the condition is not fulfilled.


src/grandorgue/sound/GOSound.cpp
```diff
@@ -319,7 +319,7 @@ bool GOSound::AudioCallback(
   GOSoundOutput *device = &m_AudioOutputs[dev_index];
   GOMutexLocker locker(device->mutex);
 
-  if (device->wait && device->waiting)
+  while (device->wait && device->waiting)
```
oleg68, line 322: Why did you change the behavior?

src/grandorgue/sound/scheduler/GOSoundScheduler.cpp
```diff
@@ -34,15 +34,9 @@ void GOSoundScheduler::Clear() {
   Unlock();
 }
 
-void GOSoundScheduler::Lock() {
-  m_ItemCount = 0;
-  __sync_synchronize();
-}
+void GOSoundScheduler::Lock() { m_ItemCount = 0; }
```
oleg68, line 37: Could you move this short function to the header file

src/grandorgue/sound/scheduler/GOSoundScheduler.cpp
```diff
@@ -34,15 +34,9 @@ void GOSoundScheduler::Clear() {
...
 
-void GOSoundScheduler::Unlock() {
-  m_ItemCount = m_WorkItems.size();
-  __sync_synchronize();
-}
+void GOSoundScheduler::Unlock() { m_ItemCount = m_WorkItems.size(); }
```
oleg68, line 39: Could you move this short function to the header file?

oleg68 (changes requested):

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Improve concurrency handling
```
oleg68, line 1:
```suggestion
- Improved concurrency handling
```

rousseldenis:
@Garfield96 Could you also enable github actions on your repository in order to be able to test easily ? Thanks

CHANGELOG.md
Garfield96: Fixed

src/grandorgue/sound/scheduler/GOSoundScheduler.cpp
Garfield96: Done

src/grandorgue/sound/GOSound.cpp
Garfield96: From my understanding, this code should wait until the condition is met, e.g. you want to wait until https://github.com/GrandOrgue/grandorgue/blob/bf5144d9dc34d6ae4daa85c01fd9bec81d8526db/src/grandorgue/sound/GOSound.cpp#L194-L198 was executed. However, condition variables suffer from spurious wakeups (https://en.wikipedia.org/wiki/Spurious_wakeup) resulting in the function call returning before the condition is fulfilled. Therefore, conditions awaited using condition variables should always be checked in a loop.

Garfield96:
@rousseldenis I apologize for the delay but the last week was a little busy.

I didn't quite get what you mean by enabling Github Actions. Github Actions are enabled in my fork, but they would only run if I would create a PR there. I configured pre-commit locally, therefore there should be no more formatting issues.

src/core/threading/GOThread.cpp
```diff
@@ -32,6 +32,6 @@ void GOThread::Stop() {
   Wait();
 }
 
-bool GOThread::ShouldStop() { return load_once(m_Stop); }
+bool GOThread::ShouldStop() { return m_Stop; }
```
oleg68, line 35:
```suggestion
bool GOThread::ShouldStop() { return m_Stop.load(); }
```

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.cpp
```diff
@@ -144,7 +144,7 @@ void GOSoundGroupWorkItem::Finish(bool stop, GOSoundThread *pThread) {
     GOMutexLocker locker(
       m_Mutex, false, "GOSoundGroupWorkItem::Finish", pThread);
 
-    if (locker.IsLocked() && m_Done != 3)
+    while (locker.IsLocked() && m_Done != 3)
```
oleg68, line 147:
```suggestion
    while (locker.IsLocked() && m_Done != 3 && !pThread->ShouldStop())
```

because GOCondition::WaitOrStop() may return when the condition is not met but the thread is marked for stopping. Your code will never finish in this case.

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.h
```diff
@@ -25,8 +27,8 @@ class GOSoundGroupWorkItem : public GOSoundWorkItem, public GOSoundBufferItem {
   GOMutex m_Mutex;
   GOCondition m_Condition;
   unsigned m_ActiveCount;
-  unsigned m_Done;
-  volatile bool m_Stop;
+  std::atomic_uint m_Done;
+  std::atomic_bool m_Stop;
```
oleg68, line 32: Because you have changed the types to atomic, you also have to add `load` and `store` everywhere with these variables in  src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.cpp

src/grandorgue/sound/scheduler/GOSoundOutputWorkItem.h
```diff
@@ -26,7 +27,7 @@ class GOSoundOutputWorkItem : public GOSoundWorkItem, public GOSoundBufferItem {
   GOSoundReverb *m_Reverb;
   GOMutex m_Mutex;
   unsigned m_Done;
-  volatile bool m_Stop;
```
oleg68, line 29: Please add `load` and `store` in the cpp file

src/grandorgue/sound/scheduler/GOSoundReleaseWorkItem.h
```diff
@@ -23,7 +25,7 @@ class GOSoundReleaseWorkItem : public GOSoundWorkItem {
   GOSoundSimpleSamplerList m_List;
   atomic_uint m_WaitCnt;
   atomic_uint m_Cnt;
-  volatile bool m_Stop;
```
oleg68, line 26: Please add load and store in the cpp file

src/grandorgue/sound/scheduler/GOSoundScheduler.cpp
oleg68:
```suggestion
void GOSoundScheduler::Lock() { m_ItemCount.store(0); }
```

src/grandorgue/sound/scheduler/GOSoundScheduler.cpp
oleg68:
```suggestion
void GOSoundScheduler::Unlock() { m_ItemCount.store(m_WorkItems.size()); }
```

src/grandorgue/sound/scheduler/GOSoundTouchWorkItem.h
```diff
@@ -17,7 +19,7 @@ class GOSoundTouchWorkItem : public GOSoundWorkItem {
 private:
   GOMemoryPool &m_Pool;
   GOMutex m_Mutex;
-  bool m_Stop;
```
oleg68, line 20: Please add load and store in the cpp file

src/grandorgue/sound/scheduler/GOSoundWindchestWorkItem.h
```diff
@@ -21,7 +23,7 @@ class GOSoundWindchestWorkItem : public GOSoundWorkItem {
   GOSoundEngine &m_engine;
   GOMutex m_Mutex;
   float m_Volume;
-  bool m_Done;
```
oleg68, line 24: Please add load and store in the cpp file

oleg68 (changes requested):

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.h
Garfield96: Accessing an atomic variable will automatically result in atomic loads and stores:

- Normal read access: https://en.cppreference.com/w/cpp/atomic/atomic/operator_T
- Normal write access: https://en.cppreference.com/w/cpp/atomic/atomic/operator%3D

I can add it nevertheless, if you consider it useful for readability.

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.cpp
Garfield96: Thanks. I extended the condition, but I had to add an additional check, whether `pthread` equals `nullptr`. This can be the case if `GOSoundOutputWorkItem::Exec()` is executed.

Is there a similar edgecase which must be considered for the `while` loop change in `GOSound.cpp` (https://github.com/GrandOrgue/grandorgue/pull/1607/files#diff-bea4aa67494cafccf86c6eca3b365a7f6c6d677b9ebd7fcac50748e5ce5204f4R322)?

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.h
oleg68: > I can add it nevertheless, if you consider it useful for readability.

Yes, it would be super. Readability is an important issue.

src/grandorgue/sound/scheduler/GOSoundGroupWorkItem.cpp
oleg68: > Thanks. I extended the condition, but I had to add an additional check, whether `pthread` equals `nullptr`. This can be the case if `GOSoundOutputWorkItem::Exec()` is executed.

Yes, checking for pthread would also be useful.

> Is there a similar edgecase which must be considered for the `while` loop change in `GOSound.cpp` (https://github.com/GrandOrgue/grandorgue/pull/1607/files#diff-bea4aa67494cafccf86c6eca3b365a7f6c6d677b9ebd7fcac50748e5ce5204f4R322)?

Yes, these changes are welcome.

Garfield96:
@oleg68 I added load and store function calls.

rousseldenis:
@oleg68

src/grandorgue/sound/scheduler/GOSoundScheduler.h
```diff
@@ -20,12 +20,12 @@ class GOSoundScheduler {
...
   std::vector<GOSoundWorkItem **> m_WorkItems;
   atomic_uint m_NextItem;
-  unsigned m_ItemCount;
+  atomic_uint m_ItemCount;
   unsigned m_RepeatCount;
   GOMutex m_Mutex;
 
-  void Lock();
-  void Unlock();
+  void Lock() { m_ItemCount = 0; }
```
oleg68, line 27:
```suggestion
  void Lock() { m_ItemCount.store(0); }
```

src/grandorgue/sound/scheduler/GOSoundScheduler.h
```diff
@@ -20,12 +20,12 @@ class GOSoundScheduler {
...
+  void Unlock() { m_ItemCount = m_WorkItems.size(); }
```
oleg68, line 28:
```suggestion
  void Unlock() { m_ItemCount.store(m_WorkItems.size()); }
```

oleg68 (changes requested):

oleg68:
@Garfield96 please resolve the conflict with CHANGELOG.md

Issue #1599, opened 2023-07-08
Switches are not distributed among manuals in a yaml file


oleg68:
When a user exports combinations from a new-style panel organ to a yaml file, all switches are saved globally instead of under manuals. Example:

```
generals:
  0:
    switches:
      001: Sousbasse 16'
      002: Bourdon 8'
      007: Montre 8'
      016: Hautbois 8'
  1:
    switches:
      001: Sousbasse 16'
      002: Bourdon 8'
      003: Contrebasse 16'
      007: Montre 8'
      008: Prestant 4'
      014: Bourdon 8'
      018: Gamba 8'
      022: II/I
```

So it is unusable for a user to read these combinations.

oleg68:
> It's kind of expected that switches should be listed globally for generals as they are never referenced from a manual for them

Why must not the switches in generals be referenced from a manual?

larspalo:
@oleg68 Because the switches, like the tremulants are global objects (they can affect more than one manual).

oleg68:
> @oleg68 Because the switches, like the tremulants are global objects (they can affect more than one manual).

But in most samplesets they belong only to a single manual for each.

So I'd look at the usage counter: if the switch is used only in one manual, it woud be referenced from the manual, otherwise it would be global.

oleg68:
> Switches in the generals always had their own section of reference

I understand that this is the current behavior, but I cann't understand why the distribution inside an internal file format (.cmb) must not be changed?

> previously stored combinations should still be working as before

Yes, of cause. I want to load both global and manual switches from .cmb, but save into the new file only as manual swithes (if they are referenced from only one manual).

larspalo:
Sure. Another option of selection is to list the switches that (in the odf) are listed under the manuals (for getting the divisionals to work) also under the manuals for the generals in the yaml file. Then the odf designer can select which switches (logically) should be shown under each manual that way.

oleg68:
@larspalo could you explain more your option? Now the ODF designer already can select switches in each manuals. The problem is that it is ignored in cmb/yaml files now.

larspalo:
Currently, the switches listed under the manuals are only there for getting divisionals to work. But couldn't you use that listing for the yaml combinations to also list/sort the switches set in a general to be shown under the manual (that they are listed under)? Then the switch enumeration under the manual would also serve a logical purpose.

[...]

PR #1618, opened 2023-08-01
Added associating of a switch with a manual https://github.com/GrandOrgue/grandorgue/issues/1599

oleg68:
This is a second PR related to #1599.

As was discussed there, not all switches remain global. If a switch is referenced exactly from one manual, then it becomes owned by this manual.

This PR introduces two GOSwitch properties: `AssociatedManualN` and `IndexInManual` and the method `GOSwitch::AssociateWithManual`. It is called when the GOManual is read from ODF.

This PR also renames `GOManual::m_switch_ids` to `m_GlobalSwitchIds`.

The new properties are not used yet, so no GO behavior should be changed.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1619, opened 2023-08-01
Increased maximum value of Pipe999LoopCrossfadeLength and Pipe999ReleaseCrossfadeLength. Now they are 3000 https://github.com/GrandOrgue/grandorgue/issues/1612

### LINKED ISSUE #1612: Pipe999LoopCrossfadeLength and Pipe999ReleaseCrossfadeLength maximum value
The attributes Pipe999LoopCrossfadeLength and Pipe999ReleaseCrossfadeLength are limited respectively to 120ms and 200ms maximum in the ranks. 
Some sample sets made for Hauptwerk can have such attributes going up to respectively 173ms and 305ms according to my observations.
I am not able to say how important is this attribute on the realism of the pipes sound rendering, but would it be worth / feasible to increase these maximum values to respectively 200 ms and 400ms for example so that OdfEdit could convert these attributes without having to clip them ?

oleg68:
Resolves #1612

Waits:
- [x] #1634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Increased maximum value of Pipe999LoopCrossfadeLength and Pipe999ReleaseCrossfadeLength. Now they are 3000 https://github.com/GrandOrgue/grandorgue/issues/1612
 # 3.12.3 (2023-08-14)
 - Fixed saving manual switches as global in yaml https://github.com/GrandOrgue/grandorgue/issues/1599
 - Fixed convolution enabling warrning in the Settings dialog https://github.com/GrandOrgue/grandorgue/issues/1617
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -10312,7 +10312,7 @@ specified in samples.
           <term>Pipe999LoopCrossfadeLength</term>
           <listitem>
             <para>
-(integer 0 - 120, default: 0) Crossfade length in ms between loop start and loop
+(integer 0 - 3000, default: 0) Crossfade length in ms between loop start and loop
 end. A cross fade requires enough samples before the start of the loop.
             </para>
           </listitem>
@@ -10321,7 +10321,7 @@ end. A cross fade requires enough samples before the start of the loop.
           <term>Pipe999ReleaseCrossfadeLength</term>
           <listitem>
             <para>
-(integer 0 - 200, default: 0) Crossfade length in ms between loop and the
+(integer 0 - 3000, default: 0) Crossfade length in ms between loop and the
 release (or other attacks). 0 means automatic selection.
             </para>
           </listitem>
--- a/src/grandorgue/model/GOSoundingPipe.cpp
+++ b/src/grandorgue/model/GOSoundingPipe.cpp
@@ -198,13 +198,13 @@ void GOSoundingPipe::Load(
     false,
     -1.0);
   m_LoopCrossfadeLength = cfg.ReadInteger(
-    ODFSetting, group, prefix + wxT("LoopCrossfadeLength"), 0, 120, false, 0);
+    ODFSetting, group, prefix + wxT("LoopCrossfadeLength"), 0, 3000, false, 0);
   m_ReleaseCrossfadeLength = cfg.ReadInteger(
     ODFSetting,
     group,
     prefix + wxT("ReleaseCrossfadeLength"),
     0,
-    200,
+    3000,
     false,
     0);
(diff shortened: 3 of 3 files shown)


larspalo:
I trust (or not really, that's why I'm asking) that you have tested what happens: 

- if loop is shorter than crossfade length
- if beginning of loop is earlier in audio stream than crossfade length
- if total length of attack is shorter than crossfade length
- if attack sample is shorter than release crossfade length

oleg68:
@larspalo Thank you for testcases. I'll test them.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Fixed maximum value of Pipe999LoopCrossfadeLength and Pipe999ReleaseCrossfadeLength. Now they are 3000 https://github.com/GrandOrgue/grandorgue/issues/1612
```
larspalo, line 1: not "Fixed"! It should read:

Increased maximum...

larspalo (changes requested):
First off, this is **not** a bug fix, it's possibly at best an enhancement of an existing feature. I've tested the functionality with my mentioned use cases and in all LoopCrossfadeLength cases GO does the right thing and aborts the load/open of the organ with a log error if the values are dangerously wrong. With the ReleaseCrossfadeLength that's not so critical since GO is simply fading in the release and provide 0 padded audio data if necessary in either direction. But it does create interesting effects as it delays the release when the values are exaggerated. The max values with this PR are most likely way too large for any realistic usage, but I won't object as it won't cause problems either.

oleg68:
> But it does create interesting effects as it delays the release when the values are exaggerated

Could you bring a test case for it?

CHANGELOG.md
oleg68: > not "Fixed"! It should read:
> 
> Increased maximum...

Should this change will be delayed until 3.13.0 as it is not a bug?

larspalo: If that suits your overall planning. I won't mind anyway.

oleg68: I planned it include to 3.12.3 as I considered this (two strong restriction) as a bug. But if we reclassify it as enchancement, we may not include this change to the bugfix-only version 3.12.3.

So we either write `Fixed` and include this change to 3.12.3 or we write `Increased` and don't include it to 3.12.3.

What woud you decision be?

larspalo: Delay it then. We should get it out fairly quickly anyway.

larspalo:
> just describe it in more details

What happens when the release crossfade gets really big, like the max value of 3000, is that when you release the note it continues to sound for some time after release until you can hear it slowly fade away. It seems like all of the release transients get lost in the process, as the attack is so slowly faded out (that it covers all the important parts of the release). The audio quality is also somewhat degraded by such a slow fade.

larspalo:
Here is a small working example that you can test to hear the effect of increasing the release crossfade length. The samples are not done to be beautiful, it's just so that one can really hear the effect clearly.

[TestReleaseXfade.zip](https://github.com/GrandOrgue/grandorgue/files/12268109/TestReleaseXfade.zip)

larspalo:
@oleg68 Have you tested the example above and realized that the "strong restrictions" actually were quite reasonable from the audio perspective, at least certainly when it comes to the crossfade length of the release. Anything over 250 is really not usable, in my opinion, as it removes too much of the release transients and introduces too much delay.

As for the loop crossfades I can say that normally a 25 ms x-fade should take care of a reasonably well aligned loop without problems when it only have minor clicks to begin with. Above 120 ms issues normally start creeping in and the fading itself can be audible. Anything above 120 I only use if I'm trying to correct faults in the audio data to save the sample (when it's already not really usable, so it cannot become worse by the fading).

CHANGELOG.md
oleg68: Fixed

oleg68:
@larspalo I've tested the example set and I couldn't find anything wrong. A large crossfade causes a longer release and it is an expected behavior.

Removing some release "percussion" is not a stopper: it brings a choice to the ODF creator.

So I'd leave the current behavior with 3000.

larspalo (approved):
> Removing some release "percussion" is not a stopper: it brings a choice to the ODF creator.

I never said it was a stopper that the release transients would be eliminated, nor that the release would be audible delayed. I said that, in my opinion, anything over 250 ms release crossfade length is unsuitable for realistic usage. Anyone with at least somewhat sensible ears should be able to hear that.

rousseldenis (approved):

Issue #1035, opened 2022-02-14
Dialog position and sizes not maintained


paolo2504:
hi developers, this not big issue just a nice to have request maybe.
"Settings" and "Load" dialogs sizes and positions are not maintained. On reopening, they revert to defaults.
Is it possible to save size and position ?
tks.

paolo2504:
Just to show this behaviour...

https://user-images.githubusercontent.com/50132661/154731722-18f00eb1-45e6-4b7d-bae4-fcd871ff41e5.mp4

oleg68:
Not only the organ selection dialog, but the positions of any dialog was never saved. Every time the dialog opens in it's original position.

So this issue is a feature request for saving all dialog positions as well as the panel positions.

eturpault:
With which Linux distribution the dialog position and sizes are maintained with GO v3.13.0-1 ?
Or is this evolution applied to the dialog windows only and not to the panel windows ?
With Ubuntu the panels position and size are not maintained, is it normal ?

oleg68:
@eturpault 

This issue relates only dialogs. Panels arestill maintained asthey were maintained before.

Sizes are maintained in any linux. But saving positions work only with xorg windows manager and doesn't work with wayland due the restriction of this window manager.

PR #1625, opened 2023-08-04
Extracted saving panel size and position code into the separate class GOGUISizeKeeper

oleg68:
I've started working on #1035 (saving GOGUISizeKeeper dialog positions).

Before the all code of saving and sizing was in `GOPanelView` and `GOGUIPanel`. It worked only with GO panels and it couldn't be used for dialogs.

This PR extracted this code to the new class `GOGUISizeKeeper`, that will be used also for dialogs.

NB: if a panel was saved as maximized, it cann't be restored on gnome desktop on recent linux versions as like it didn't work before this PR.


oleg68:
@larspalo, @rousseldenis could you approve this PR?

rousseldenis (approved):

larspalo (approved):

PR #1630, opened 2023-08-09
Fixed saving manual switches as global in yaml https://github.com/GrandOrgue/grandorgue/issues/1599

oleg68:
Resolves: #1599 

After this PR the switches referenced only from one manual start saving to .cmb and .yaml as manual switches instead of global ones.

When importing from .cmb and .yaml, both kinds of the switch references (global or manual) are supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed saving manual switches as global in yaml https://github.com/GrandOrgue/grandorgue/issues/1599
 - Fixed convolution enabling warrning in the Settings dialog https://github.com/GrandOrgue/grandorgue/issues/1617
 - Fixed "Release Length is not valid" error when applying changes for several Organ Settings objects at once https://github.com/GrandOrgue/grandorgue/issues/1601
 - Fixed saving all combinations as full
--- a/src/grandorgue/combinations/model/GOCombination.cpp
+++ b/src/grandorgue/combinations/model/GOCombination.cpp
@@ -106,7 +106,8 @@ void GOCombination::SetStatesFromYaml(
       maxElementNumber = r_OrganModel.GetTremulantCount();
       break;
     case GOCombinationDefinition::COMBINATION_SWITCH:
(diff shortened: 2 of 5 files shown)


rousseldenis (approved):

larspalo:
I do get a lot of warnings that the switches have wrong name(s) when importing older combinations, so it's not really silent in the import though!

The warnings look like:

tor 10 aug 2023 21:21:19: 21:21:19: Varning: Fel namn "Octava Bass 4'" istället för "Super Octava 2'" hos switch 007

which approximately translates to

Warning: Wrong name "Octava Bass 4'" instead of "Super Octava 2'" at switch 007

larspalo:
I should perhaps add that this is when importing from older yaml combinations. The older cmb seem to be not so fussy about it.

oleg68:
@larspalo Could you tell me what is the sample set and post here the old yaml file?

larspalo:
@oleg68 It's Norrfjärden...

[TestingNfj1.zip](https://github.com/GrandOrgue/grandorgue/files/12321902/TestingNfj1.zip)

I think that the combinations work though, and the fact that the internal implementation has changed since the file was made sure explains the warnings. Honestly, I wouldn't make too much of it if you're not worried about it, so if you're ok with it after having examined the file I'll approve.

oleg68:
@larspalo 

It looks like the old GO version placed incorrect switch names to the yaml for banked divisionals (from global switches instead of the manual ones). This bug is fixed with this PR also.

So the warnings are about wrong switch names, they are correct.

larspalo:
> So the warnings are about wrong switch names, they are correct.

@oleg68 Yes, that's what I thought too. So this PR corrects the switch names from what was actually wrong before. Do you think this PR is good enough that we can let it pass even thought it might cause warnings?

oleg68:
> Do you think this PR is good enough that we can let it pass even thought it might cause warnings?

Yes, I do.

larspalo (approved):

PR #1636, opened 2023-08-16
Renamed GOPanelView to GOGUIPanelView and moved to the gui directory

oleg68:
Working with #1035 I noticed that the `GOPanelView` class is close to other GUI classes.

So this PR renames `GOPanelView` to `GOGUIPanelView` and moved to the `gui` directory.


larspalo:
When testing this PR I've noticed a few crashes when closing GO, however it doesn't seem to be related to this PR at all (and I can make it happen with a current build too), but it's likely worth examining more closely.

[CrashReportClose20230818.txt](https://github.com/GrandOrgue/grandorgue/files/12401701/CrashReportClose20230818.txt)

oleg68:
> When testing this PR I've noticed a few crashes when closing GO, however it doesn't seem to be related to this PR at all (and I can make it happen with a current build too), but it's likely worth examining more closely.

It looks like some GUI element of the main toolbar is deleted twice: one times explicitly and one times as part as a child of the toolbar window.

Could you give me reproducing steps?

larspalo:
> Could you give me reproducing steps?

Sure, it seems to happen every time both with a debug build and a release build on (Ubuntu 20.04) Linux.

- Start GrandOrgue from a terminal either with

```
gdb ./GrandOrgue
run
```
or
`pasuspender ./GrandOrgue`
or
`./GrandOrgue
`
- Open File->Settings and change from using PortAudio Jack to other device ALSA (or the other way around: from ALSA device to Jack).
- Ok and close Settings dialog
- Close GrandOrgue from File menu and watch the message in terminal or gdb

larspalo (approved):
As the crash I mentioned earlier isn't related to this PR, I approve.

rousseldenis (approved):

oleg68:
> * Open File->Settings and change from using PortAudio Jack to other device ALSA (or the other way around: from ALSA device to Jack).
> * Ok and close Settings dialog
> * Close GrandOrgue from File menu and watch the message in terminal or gdb

I couldn't reproduce the crash: GO closes correctly.

Could you post the screenshot of File->Settings->Audio before the crash?

larspalo:
@oleg68 This is using a standard Ubuntu 20.04 and a local build. I've re-built both the release and debug version of GO from latest official source and it crashes at closing GO even without having done anything at all after startup, it really doesn't matter if I use PortAudio or RtAudio. It just crashes at closing. I also managed to get a crash while changing from PortAudio to RtAudio in the Settings dialog, but that's another thing, I guess.

Screenshots and crash reports from gdb: 
[GoCrash20230826.zip](https://github.com/GrandOrgue/grandorgue/files/12444952/GoCrash20230826.zip)

PR #1646, opened 2023-09-05
Added saving dialog positions and sizes https://github.com/GrandOrgue/grandorgue/issues/1035

oleg68:
Resolves: #1035

This PR adds saving positions and sizes of all dialogs that inherit `GODialog`.

The sizes are saved to the organ setting file as well as to the GrandOrgue config.

Notes:
- Some window managers (ex. Gnome/Wayland) do not allow applications to maintain window positions.
- Not all dialogs inherit `GODialog` now. Please create separate issues/PRs about these dialogs.


oleg68:
> While testing this PR I've had multiple crashes:
> 
> * when changing sample rate and bit size in the Settings dialog (audio related)
> * when closing GrandOrgue, just like I reported earlier, but this is to be expected as it has happened before
> * when opening the OrganSettings dialog after having imported settings from an exported settings file

@larspalo  Could you check the same actions with [the version before this PR](https://github.com/GrandOrgue/grandorgue/actions/runs/6082537418)? If the crashes stilll exist then they are not related with this PR and they should be addressed to a separate issue.

larspalo:
@oleg68 The crash at close exist on your mentioned build (and was detected earlier) so it's not related to this PR.

However with this PR I get frequent crashes at opening different dialogs with a message of "corrupted size vs. prev_size" and a warning that a process was terminated with signal 6, or a crash with no other message than a warning that a process was terminated with signal 11. I'll try a local debug build of this PR and see if that can help narrowing down what causes the problem.

larspalo:
I can consistently crash GO with a debug build of this PR. I

- open a sample set
- import settings
- open Organ Settings dialog

I attach the gdb report and the settings file:
[CrashReport20230908.zip](https://github.com/GrandOrgue/grandorgue/files/12556706/CrashReport20230908.zip)

I do note that the only dialog saved is the Settings dialog. When I try opening the Organ Settings dialog (that doesn't exist in the [DialogSizes] section) GO always crashes.

oleg68:
> I attach the gdb report and the settings file: [CrashReport20230908.zip](https://github.com/GrandOrgue/grandorgue/files/12556706/CrashReport20230908.zip)
> 
When I try opening the Organ Settings dialog (that doesn't exist in the [DialogSizes] section) GO always crashes.

Thank you for the crash report. I fixed this.

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1392, opened 2023-02-17
Separate MIDI events for Stop ON and Stop OFF


jrsorgan:
Stop control from an organ console appears to work only one way.  Stop can be activated with a midi B0 11 0F but will not turn off with B0 10 0F  Am I missing something in the setup or is this a current bug.
I am very new to GO and any help is appreciated.
Thank you

larspalo:
Did you configure the MIDI send events?

jrsorgan:
Midi stop control from the console is not configurable.  The stop tab sends a midi B0 10 0F to turn the stop on in GO.  The tab off sends a midi B0 11 0F which is not recognized by GO.

larspalo:
Sounds like a complex midi detection might be needed instead of the normal simple event detection (done from the same right-click dialog). My midi message knowledge is a bit rusty, but the messages you cite seems to be referring to two different controllers.

jrsorgan:
I have tried the complex midi detection process and that doesn't work either.  I loaded a demo of HW4.2 and it works fine but I prefer GO.  I did a quick search for this problem and it appears others are having the same problem.  Maybe one of the programmers will take a look and see if there is a bug.  Thanks for your suggestions.

larspalo:
But GO recognizes both events, right? What if you add both events for the same stop (alternating)?

jrsorgan:
Yes GO recognizes both events.  I'm not sure how to add 2 events for the same stop.  If you can point me in the right direction?\
J

larspalo:
Just add one and then click the "New" button for the other.

jrsorgan:
I tried your suggestion using Listen for Event and Detect complex MIDI setup.  I'm using MIDI Medic to send the stop on  B0 11 0F  and  stop off B0 10 0F  messages.  I can get the stop to turn on fine.  Can't get it to turn off.  The console is at church but I captured the tab on/off messages with MIDI-OX.  I will keep trying.

jrsorgan:
I would appreciate it if someone could test my stop on, stop off problem.  So far most people have said that it will work.  I have been unable to get the stop off midi message to work.  I have tested at home using MIDI Medic to send the commands.  I have tested it at church with the console.  Stop on always works just fine.  Stop off does not.

larspalo:
[...]

PR #1654, opened 2023-09-14
Clarified code of GOMidiEventRecvTab

oleg68:
This is the first PR related to #1392.

1. Added some comments in the GOMidiEventRecvTab header
2. Introduced the new variable GOMidiEventRecvTab::m_ReceiverType for eliminating calling m_midi.GetType()
3. Removed extra returns from GOMidiEventRecvTab::DetectEvent()

It is just refactoring. No GO behavior should be changes.

### DIFF
--- a/src/grandorgue/dialogs/midi-event/GOMidiEventRecvTab.h
+++ b/src/grandorgue/dialogs/midi-event/GOMidiEventRecvTab.h
@@ -38,6 +38,7 @@ private:
 
   GOMidiReceiverBase *m_original;
   GOMidiReceiverEventPatternList m_midi;
+  GOMidiReceiverType m_ReceiverType;
   GOMidiListener m_listener;
   GOChoice<GOMidiReceiverMessageType> *m_eventtype;
   wxChoice *m_eventno, *m_channel, *m_device;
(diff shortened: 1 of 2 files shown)


oleg68:
@rousseldenis @larspalo could you approve this PR because it delays completing #1392

src/grandorgue/dialogs/midi-event/GOMidiEventRecvTab.h
```diff
@@ -38,6 +38,7 @@ class GOMidiEventRecvTab : public wxPanel,
 
   GOMidiReceiverBase *m_original;
   GOMidiReceiverEventPatternList m_midi;
+  GOMidiReceiverType m_ReceiverType;
```
larspalo, line 41: I'm somewhat concerned about this. I understand the supposed "simplification", but GOMidiReceiverType is a typedef:ed enum that's buried in another header file (src/core/midi/GOMidiReceiverEventPatternList.h) that all breath of old c style coding that should be structured in a better way, in my opinion. To me, this "clarification" is not really so "clear" as there is no class or separate file that can be easily search for to find the defined type.

oleg68: You are right. But extracting GOMidiReceiverType is a subject of a separate pr #1664.

oleg68:
@larspalo @rousseldenis could you approve this PS because it blocks #1392 ?

rousseldenis (approved):

larspalo (approved):

PR #1664, opened 2023-09-21
Extracted enums from GOMidiReceiverEventPatternList.h to separate header files

oleg68:
As @larspalo mentioned in https://github.com/GrandOrgue/grandorgue/pull/1654#discussion_r1332084229, I moved enum definitions from `GOMidiReceiverEventPatternList.h` to separate header files.


larspalo:
@oleg68 I think this is ok, but the build seems to have failed.

oleg68:
@larspalo there were some temporary problems in GitHub. The build has successed upon retry.

rousseldenis (approved):

larspalo (approved):

Issue #1660, opened 2023-09-21
AppImage creates multiple copies of the demo organ


ahall41:
Every time you launch GrandOrgue as an Appimage (after the first), you get an error e.g.
`Thu 21 Sep 2023 17:49:48 BST: 17:49:48: Error: can't open file '/tmp/.mount_grandoWSlhCy/usr/share/GrandOrgue/packages/demo-4232D4C616E4B99145537F02B34D342D4E3E3808.orgue' (error 2: No such file or directory)`
Together with another copy of the demo organ. And continue to do so until you switch to "Start without any organ".

oleg68:
Fixed with #1681 and #1682

PR #1681, opened 2023-10-14
Fixed error messages after multiple runs of GrandOrgue ftom Appimage with a demo organ https://github.com/GrandOrgue/grandorgue/issues/1660

oleg68:
This is the first PR related to #1660

Because running GrandOrgue from applimage causing unpacking it to a temporary directory, the demo Organ may be registered multiple time with different paths. Old paths are not more valid, because the old directories have already been deleted.

Earlier opening an organ from a package were only by ArchiveId and the path was not taken in account. It caused trying to open the archive from all paths had been registered and logging error messages on each invalid path.

This PR adds passing the archive path to GOArchiveManager::LoadArchive that makes unnecessary looking over invalid paths. So this PR *eliminates logging extra error messages*.

Multiple demo organs registered are still an issue and will be fixed with another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed error messages after multiple runs of GrandOrgue ftom Appimage with a demo organ https://github.com/GrandOrgue/grandorgue/issues/1660
 # 3.13.0 (2023-10-11)
 - Implemented option to send MIDI Note Off as 0x8X or 0x9X with velocity 0 https://github.com/GrandOrgue/grandorgue/issues/1640
 - Added capability of control buttons with Control Change MIDI events with different keys but same values using "Bx Ctrl Change Fixed On Value Toggle" and "Bx Ctrl Change Fixed Off Value Toggle" https://github.com/GrandOrgue/grandorgue/issues/1392
--- a/src/core/archive/GOArchiveCreator.cpp
+++ b/src/core/archive/GOArchiveCreator.cpp
@@ -66,7 +66,9 @@ bool GOArchiveCreator::AddPackage(const wxString &path) {
       return true;
     }
   m_packageIDs.push_back(a);
-  GOArchive *archive = m_Manager.LoadArchive(a->GetID());
+
+  GOArchive *archive = m_Manager.LoadArchive(a->GetID(), path);
+
   if (!archive) {
     wxLogError(_("Failed to load organ package %s"), path.c_str());
     return false;
--- a/src/core/archive/GOArchiveManager.cpp
+++ b/src/core/archive/GOArchiveManager.cpp
@@ -108,25 +108,31 @@ bool GOArchiveManager::ReadIndex(GOArchive *archive, bool InstallOrgans) {
   return true;
 }
 
-GOArchive *GOArchiveManager::LoadArchive(const wxString &id) {
-  for (unsigned i = 0; i < m_OrganList.GetArchiveList().size(); i++) {
-    const GOArchiveFile *file = m_OrganList.GetArchiveList()[i];
-    if (file->GetID() != id)
-      continue;
-    GOArchive *archive = OpenArchive(file->GetPath());
-    if (!archive)
-      continue;
-    if (!ReadIndex(archive, archive->GetArchiveID() != id)) {
-      delete archive;
-      continue;
+GOArchive *GOArchiveManager::LoadArchive(
+  const wxString &id, const wxString &archivePath) {
+  GOArchive *pResArchive = nullptr;
+
+  for (const GOArchiveFile *file : m_OrganList.GetArchiveList()) {
+    const wxString &filePath = file->GetPath();
+
+    if (
+      file->GetID() == id
+      && (archivePath.IsEmpty() || filePath == archivePath)) {
+      GOArchive *archiveProbe = OpenArchive(filePath);
+
+      if (archiveProbe) {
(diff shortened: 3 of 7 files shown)


larspalo:
@oleg68 Would it be an option to just remove the demo sample set from the AppImage build?

larspalo:
@oleg68 Actually, earlier we had a separate build artifact for only the demo sample set to download. Perhaps it would be a good idea to remove the default inclusion of the demo sample set from all builds if not explicitly set to be included and have a separate target for only the demo sample set (or even just include a link to it as it doesn't change that often and could be just downloaded once...).

What this would do is to reduce the sheer amounts of different demo sample sets some of us have floating around when we're testing different builds... Installation of the demo sample set would be a choice and exist only in one instance (if not the user decides otherwise when building from source).

oleg68:
@larspalo I object to remove the demo from appimage because i is intended for using by not advanced users for whom it is difficult to download a separate sample set and to open it with GrandOrgue.

The next PR will automatically remove unexisting organs that were opened from a temporary directory. So unpacking GO under the temporary directory will solve your problem of testing multiple versions.

larspalo:
> So unpacking GO under the temporary directory will solve your problem of testing multiple versions.

No, it won't. Every time I download an archive of GO to test a PR a new demo sample set is registered with it.

> I object to remove the demo from appimage because i is intended for using by not advanced users for whom it is difficult to download a separate sample set and to open it with GrandOrgue.

One doesn't "Open" a package in GO, it's "Loaded". If it's too difficult to download and place the .orgue package in the package folder for GO and then start GO - the intelligence of people in the world is really going downhill...

larspalo:
@oleg68 Just to prove my point, here is a screenshot of the organ list from GrandOrgue on my computer.

Skärmbild från 2023-10-14 17-33-00

The fact that most of the demo sample sets listed are actually deleted but GO doesn't report them as missing is a bug in itself...

I usually clean up to list from time to time, but it's quite un-necessary that we include the demo sample set in **any** build that is not in fact a **release**.

oleg68:
> The fact that most of the demo sample sets listed are actually deleted but GO doesn't report them as missing is a bug in itself...

I hope this bug is fixed with #1682. Earlier the existance was checked by ArchiveId rather than by path, so GO considered an absent archive as existing because there was another existing archive with the same id.

oleg68:
> > So unpacking GO under the temporary directory will solve your problem of testing multiple versions.
> 
> No, it won't. Every time I download an archive of GO to test a PR a new demo sample set is registered with it.
> 

#1682 will help you if you unpack GO under the temp dir.

> > I object to remove the demo from appimage because i is intended for using by not advanced users for whom it is difficult to download a separate sample set and to open it with GrandOrgue.
> 
> One doesn't "Open" a package in GO, it's "Loaded". If it's too difficult to download and place the .orgue package in the package folder for GO and then start GO - the intelligence of people in the world is really going downhill...

The package folder does not exist before GO runs for the first time. The current behaviour is the most simple: to download appimage, to set it executable, to double click and to play without any extra steps. Necessarity of downloading a sample set adds complexity. So I'd leave the demo organ packaged in appimage.

larspalo:
> So I'd leave the demo organ packaged in appimage.

Ok, but it should be removed from all others, and instead be made available as an optional download.

larspalo:
> https://github.com/GrandOrgue/grandorgue/pull/1682 will help you if you unpack GO under the temp dir.

A true solution must work no matter what directory is used. Each version of an archive will be extracted to another base directory by default. An executable, archive (.tar.gz, .zip or whatever really) should not be required to be extracted to a specific place, it is the .orgue package that should be placed in a specific place...

oleg68:
> > So I'd leave the demo organ packaged in appimage.
> 
> Ok, but it should be removed from all others, and instead be made available as an optional download.

I ageree with removing the demo from .zip and .tar.gz, because they usually used for resting where GO has already installed.

But I still insist on presenting the demoin all installation packages, because they are overwritten by the next installation and no multiplication occurs.

But changing shipping the demo is a subject of a separate issue. @larspalo do you have any reasons why not to get rid of error messages when missing archives exist, but the current organ is loaded from an existing one?

larspalo (approved):
@oleg68 What I have the most problem with is the approach of trying to hide a bigger issue like this. The correct solution, which you know, is that organ packages should always be placed in the package directory that GrandOrgue searches, where any duplicates never could occur anyway (because they would be overwritten). During an installation, the demo package should be moved to the correct directory which should be created if not existing. So if the installation process would do the correct thing then no problem would arise.

But more generally, I think that this full path to the organ packages is actually a very good idea! As I mentioned in https://github.com/GrandOrgue/grandorgue/discussions/1544#discussioncomment-5999920 it would be really great if we can reference to samples existing within packages for re-using without needing to un-pack them first. As I mention and show in that post, this is partially possible already but since the paths are not complete the risk is present that conflicting "paths" can occur when trying to access the samples if multiple organs are used that have the same/similar naming structure. I've not had the time to investigate this but I do see a potential here!

oleg68:
@larspalo 

There are lots of places in GO where organ packages are referenced by archive id instead of by path. This PR eliminates one of them - loading an organ, #1682 fixes another place - checking the organ for existence.

But, of cause, a dependency between organ packages are still by id rather than by full path and now I don't have any ideas how to improve it. Lets continue the discussion in the discussion area, not here.

PR #1682, opened 2023-10-14
Added deregistering organs in the temporary directory that do not more exist https://github.com/GrandOrgue/grandorgue/issues/1660

oleg68:
This is the second and the last PR related to #1660

Now when GrandOrgue starts, it automatically removes registration of all organs that were under the temporary directory and are not more exist. It solves the problem of multiplication of the Demo organ when GO is rinning from Appimage.


larspalo (approved):
Yes, with this PR the non existing organs are correctly detected and listed as missing.

rousseldenis (approved):

Issue #1696, opened 2023-10-29
Organ fails to start with invalid configuration


radionerd:
While trying to understand the format of sysex messages to my LCDs which seem to vary in text start position I suspect that I configured the metronome displays (BPMinute or BPBar) of PiteaMHS badly. Now the organ fails to start with the following log message displayed:

> Sun 29 Oct 2023 15:54:08: 15:54:08: Error: Out of range value at section 'SetterCurrentPosition' entry 'MIDISendLength001': 32

Selecting the panel tab, where I would expect to see the Metronome panel, a blank rectangle is displayed. This prevents me from updating/correcting the configuration so I am stuck. I cannot even determine what the bad configuration item is. The organ was running nicely before I closed and restarted it.

![Screenshot from 2023-10-29 16-31-08](https://github.com/GrandOrgue/grandorgue/assets/3194568/742b3422-b937-4be4-9862-85619d2beca1)

Ideally I would like to be able to restore the organ configuration to something that starts without losing the unrelated receive and send event configurations. The problem occurred running grandorgue-3.12.3-1.linux.aarch64, and persists with grandorgue-3.13.0-1.linux.aarch64. This seems like a bug to me.

The setup is:
- Console:  Homebrew USB connected running on STM32F103C8T6 Blue pill in active development
- Computer: Raspberry Pi4 4GB
- OS: Ubuntu Linux 22.04
- Organ: Pitea MHS
- GO Versions 3.13.0-1.linux.aarch64 and 3.12.3-1

Any help, or suggestions for further investigations much appreciated. Thanks for such a wonderful project.

oleg68:
A workaround: delete the old .cmb file from the GrandOrgue/Data folder, start GrandOrgue and make the configuration from scratch. Don't use 32 as the midi message length due the bug #1686.

A more complex workaround: unpack the text file from the .cmb file, change the MIDISendLength001 value to 31 and replace the packed .cmb file with the plain text file.

radionerd:
Thanks Oleg. Your first work around works well. I can now open the Pitea organ once more and continue my console development. Would you be able to suggest an app for opening the cmb file? My ubuntu graphical extraction tool does not seem up to the job.

I'm inclined to leave my issue open as it underlines several issues:

 - Only the time and date of the .cmb file might offer a clue as to its purpose or its possible role in a workaround.
 - We should be pedantic about the configurations that we save
[...]

PR #1707, opened 2023-11-04
Fixed loading an organ when some configuration entry is out of range https://github.com/GrandOrgue/grandorgue/issues/1696

oleg68:
Resolves: #1696

This PR changes handling of ``Out of range`` configuration errors dramatically.

Earlier loading the organ was finishing. Now the default value is used for this configurantion entry and loading the organ continues. An error message appears in the Log window.

I'm not sure that the new behavior does not cause unattended side effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed loading an organ when some configuration entry is out of range https://github.com/GrandOrgue/grandorgue/issues/1696
 - Fixed crash when trying to load a sample set with a truncated wave file https://github.com/GrandOrgue/grandorgue/discussions/370
 - Fixed crash on closing an organ https://github.com/GrandOrgue/grandorgue/issues/1678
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):

oleg68:

larspalo:
@oleg68 I don't like this at all. The out of range checking is there for very good reasons for the .odf. If, and I mean only really if, we should apply such "default value instead when out of range" strategy, it should really be limited to only "some configuration entries" and not all of them. If for instance a too big switch number is referenced as a GUI element it would default to the first (switch) instead.

Personally, I think that this new loose out of range checking would perhaps be ok for .cmb settings mostly, that are supposed to have been written by GO itself, the .odf should be strictly correct to begin with.

oleg68:
@larspalo could you approve this PR?

larspalo (approved):

Issue #1677, opened 2023-10-09
rtaudio and rt midi released version 6.х


Pro-pra:
with system rtaudio-6.0 rtmidi-6.0 build fail
```
src/grandorgue/sound/ports/GOSoundRtPort.cpp:34:12: error: 'RtAudioError' does not name a type; did you mean 'RtAudioErrorType'?
   34 |   } catch (RtAudioError &e) {

```

with internal rtaudio and rtmidi build fail:
```
grandorgue-3.12.3-1/src/grandorgue/midi/ports/GOMidiRtPortFactory.h:9:10: fatal error: rtmidi/RtMidi.h: No such file or directory
    9 | #include "rtmidi/RtMidi.h"
      |          ^~~~~~~~~~~~~~~~~
```

Pro-pra:
my frend send this patch, may be this help?
https://abf.io/import/grandorgue/blob/rosa2023.1/grandorgue-3.13.0-fixes-for-rtaudio6.patch

svillemot:
A patch has also been submitted to the Debian bug tracking system, see https://bugs.debian.org/1051559

oleg68:
Could you submit a normal PR here instead of the patch?

rousseldenis:
@oleg68 IMHO, this patch applies only if we update RtAudio to 6.0 version. So, it should be embedded too in the PR (if we want to update it of course)

Pro-pra:
i'm not programmer, sorry. no PR from me.

larspalo:
> IMHO, this patch applies only if we update RtAudio to 6.0 version. So, it should be embedded too in the PR (if we want to update it of course)

@rousseldenis @oleg68 Yes, this is correct. The error handling in RtAudio, as well as a few other things, have changed significantly since the version GO currently uses. If we change anything in the code then the commit that the submodule of the library currently point to must be updated (ie. a later version of RtAudio will be used).

However, since I've recently tested this for LoopAuditioneer I can say that it's not without issues to go to the latest RtAudio release. Currently I cannot get the DirectSound api working with the cross compile to windows, but this might be specific to that project... After changing a few things in my code I got the 6.0.1 version working under Linux without issues.

PR #1713, opened 2023-11-06
Switched to RtAudio 6.0

oleg68:
Resolves: #1677

This PR.
1. Switches to RtAudio and RtMidi 6.0
2. Changed error handling to compatibility with RtAudio 6.0
3. Changed the device identification from the index to the device id.


larspalo (approved):
While testing this PR on Ubuntu 20.04 I've had an extreme amount of crashes/hangs when trying to use RtAudio devices other than pulse (which has quite poor performance). However, since this is kind of expected as RtAudio doesn't really like my usb audio device (Alesis iO2) I approve (and will continue to use PortAudio instead...).

rousseldenis (approved):

PR #1742, opened 2023-12-13
Eliminated a lot of copying wxString instances when reading setting files

oleg68:
This PR eliminates lots of copying wxString instances when reading configuration, .cmb and odf files for

1. Increasing performance
2. Simplifying debugging by eliminating extra calls of copy constructor

It is just optimisation. No GO behavior should be changed.


rousseldenis (approved):
LGTM

larspalo (approved):

Issue #1724, opened 2023-11-18
Not enough samples for a crossfade


eturpault:
Now that the attribute Pipe999LoopCrossfadeLength is set by OdfEdit when making conversion from HW to GO (in link with the issue #1612), I get the following error with 5 wav files when loading the Prytanée sample set of Sonus Paradisi :
"Not enough samples for a crossfade"

Here are the properties of these 5 wav files and the value Pipe999LoopCrossfadeLength set for them (they are all at 48kHz 24 bits and contain at least 3 loops) : 
- 9.54 seconds, 458155 samples, Pipe999LoopCrossfadeLength=118
- 9.60 seconds, 460912 samples, Pipe999LoopCrossfadeLength=120 
- 9.22 seconds, 442764 samples, Pipe999LoopCrossfadeLength=188
- 9.19 seconds, 441154 samples, Pipe999LoopCrossfadeLength=190
- 9.35 seconds, 448938 samples, Pipe999LoopCrossfadeLength=188

For example the first file is this one :  [061-c#_050.zip](https://github.com/GrandOrgue/grandorgue/files/13400614/061-c._050.zip)

I see in the GO source code that this error is generated in the file GOSoundAudioSection.cpp : 
```
        if (start_seg.start_offset < fade_len)
          throw(wxString) _("Not enough samples for a crossfade");
```
Is the error triggered because one loop of the wav file is starting too early ? They are all starting after 2 seconds, it is not so early. Else what can be causing that ?
I get no more error by reducing the Pipe999LoopCrossfadeLength values of these pipes at 45 for example.

oleg68:
@eturpault I can not reproduce the problem with the sample you provided.

Could you make a test .organ file with only one sample that would reproduce this issue?

eturpault:
@oleg68 : Here is a mini sample set permitting to reproduce the issue with one sample.
In GO settings "Loop loading" must be set to "All loops" and "Attack loading" to All, else the issue is not reproducible.
[Issue1724_demo.zip](https://github.com/GrandOrgue/grandorgue/files/13625880/Issue1724_demo.zip)

oleg68:
This error comes from another sample: `061-c#.wav`. It has six loops:

![loops](https://github.com/GrandOrgue/grandorgue/assets/11586506/a66f6e8d-8307-41b3-8267-4addf7b2fa5e)

One of them starts with 5265, that is too early, because the crossfade requires 118*48=5664 samples before the loop start point.

[...]

PR #1751, opened 2023-12-19
Renamed some methods added the const qualifier and added some comments

oleg68:
This is the first PR related to #1724

This PR makes the code a little bit clear
1. Addes the `const` qualifier to lots of GetXXX() methods
2. Added the `override` qualifier to lots of methods
3. Renamed some methods
4. Added some comments
5. Moved some short implementations from .cpp to .h
6. GOAudioSection::Setup(): extracted the same code from the both branches of if/else

### DIFF
--- a/src/grandorgue/sound/GOSoundAudioSection.cpp
+++ b/src/grandorgue/sound/GOSoundAudioSection.cpp
@@ -732,6 +732,9 @@ void GOAudioSection::Setup(
         loop_length * m_BytesPerSample,
         (end_length - copy_len) * m_BytesPerSample);
       if (fade_len > 0)
+        // TODO: Remove the parameter names from the comment and reduce the
+        // number of parameters of DoCrossfade that the call would be easy
+        // readable without additional comments
         DoCrossfade(
           end_seg.end_data,                  // dest
           MAX_READAHEAD,                     // dest_offset
--- a/src/grandorgue/sound/GOSoundAudioSection.cpp
+++ b/src/grandorgue/sound/GOSoundAudioSection.cpp
@@ -668,11 +668,13 @@ void GOAudioSection::Setup(
     /* Setup the loops and find the amount of data we need to store in the
      * main block. */
     unsigned min_reqd_samples = 0;
+
     for (unsigned i = 0; i < loop_points->size(); i++) {
       audio_start_data_segment start_seg;
       audio_end_data_segment end_seg;
       const GOWaveLoop &loop = (*loop_points)[i];
       unsigned fade_len = crossfade_length;
+
       if (loop.m_EndPosition + 1 > min_reqd_samples)
         min_reqd_samples = loop.m_EndPosition + 1;
 
@@ -683,24 +685,19 @@ void GOAudioSection::Setup(
         = 1 + end_seg.end_offset - start_seg.start_offset;
       unsigned end_length;
 
+      if (fade_len > end_seg.end_offset - start_seg.start_offset)
+        throw(wxString) _("Loop too short for crossfade");
+
+      if (start_seg.start_offset < fade_len)
+        throw(wxString) _("Not enough samples for a crossfade");
+
+      // calculate the fade segment size and offsets
       if (end_seg.end_offset - start_seg.start_offset > SHORT_LOOP_LENGTH) {
-        if (fade_len > end_seg.end_offset - start_seg.start_offset)
-          throw(wxString) _("Loop too short for crossfade");
-
-        if (start_seg.start_offset < fade_len)
-          throw(wxString) _("Not enough samples for a crossfade");
-
         end_seg.transition_offset
           = end_seg.end_offset - MAX_READAHEAD - fade_len + 1;
         end_seg.read_end = end_seg.end_offset - fade_len;
         end_length = 2 * MAX_READAHEAD + fade_len;
       } else {
-        if (fade_len > end_seg.end_offset - start_seg.start_offset)
-          throw(wxString) _("Loop too short for crossfade");
-
-        if (start_seg.start_offset < fade_len)
-          throw(wxString) _("Not enough samples for a crossfade");
-
         end_seg.transition_offset = start_seg.start_offset;
         end_seg.read_end = end_seg.end_offset;
         end_length = SHORT_LOOP_LENGTH + MAX_READAHEAD;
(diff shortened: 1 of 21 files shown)


src/grandorgue/sound/GOSoundProviderWave.cpp
```diff
@@ -278,7 +278,7 @@ unsigned GOSoundProviderWave::GetFaderLength(unsigned MidiKeyNumber) {
   return fade_length;
 }
 
-void GOSoundProviderWave::LoadFromFile(
+void GOSoundProviderWave::LoadFromSampleFiles(
```
larspalo, line 281: I'd instead suggest: LoadSamplesFromFile(

Reason being, the plural you add indicates that multiple files are handled here.

src/grandorgue/sound/GOSoundProviderWave.h
```diff
@@ -114,7 +117,11 @@ class GOSoundProviderWave : public GOSoundProvider {
   unsigned GetFaderLength(unsigned MidiKeyNumber);
 
 public:
-  void LoadFromFile(
+  /**
+   * Load all attack and release samples from corresponding .wav files or from
+   *an archive
+   **/
+  void LoadFromSampleFiles(
```
larspalo, line 124: void LoadSamplesFromFile(

src/tools/GOPerfTest.cpp
```diff
@@ -102,7 +102,7 @@ void GOPerfTestApp::RunTest(
         ainfo.release_end = -1;
         ainfo.loops.clear();
         attack.push_back(ainfo);
-        w->LoadFromFile(
+        w->LoadFromSampleFiles(
```
larspalo, line 105: w->LoadSamplesFromFile(

src/grandorgue/sound/GOSoundProviderWave.h
```diff
@@ -114,7 +117,11 @@ class GOSoundProviderWave : public GOSoundProvider {
   unsigned GetFaderLength(unsigned MidiKeyNumber);
 
 public:
-  void LoadFromFile(
+  /**
+   * Load all attack and release samples from corresponding .wav files or from
+   *an archive
```
larspalo, line 122: * an archive

src/grandorgue/sound/GOSoundProviderWave.h
```diff
@@ -89,7 +89,10 @@ class GOSoundProviderWave : public GOSoundProvider {
     unsigned channels,
     bool compress);
 
-  void ProcessFile(
+  /* Load attack or/and release samples from one wav file or from an archive
```
larspalo, line 92: /* Load attack **and/or** release

larspalo (changes requested):
I've got nothing against the functionality. But, as the intention is to bring clarity to the code, I have added a few suggestions that I think should be improved upon, or omitted, since they, in your PR, doesn't really do what you intend them to do.

src/grandorgue/sound/GOSoundAudioSection.cpp
```diff
@@ -734,15 +733,15 @@ void GOAudioSection::Setup(
...
-          end_seg.end_data,
-          MAX_READAHEAD,
-          (const unsigned char *)pcm_data,
-          start_seg.start_offset - fade_len,
-          pcm_data_channels,
-          m_BitsPerSample,
-          fade_len,
-          loop_length,
-          end_length);
+          end_seg.end_data,                  // dest
```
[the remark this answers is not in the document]
oleg68: When I discovered #1724 it was too difficult for me to understand how `DoCrossfade` worked and to trace wthat value were passed to each parameter because this method had lots of parameters. So I want to mark just the parameter names for making this job easier. So I'd leave them in comments

src/grandorgue/sound/GOSoundAudioSection.cpp
```diff
@@ -734,15 +733,15 @@ void GOAudioSection::Setup(
...
+          MAX_READAHEAD,                     // dest_offset
+          (const unsigned char *)pcm_data,   // src
+          start_seg.start_offset - fade_len, // src_offset
+          pcm_data_channels,                 // channels
+          m_BitsPerSample,                   // bits_per_sample
+          fade_len,                          // fade_length
+          loop_length,                       // loop_length
```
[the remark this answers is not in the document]
oleg68: The same as above.

src/grandorgue/sound/GOSoundProviderWave.cpp
oleg68: This methods really may load samples from multiple files: attacks, releases. So the plural is really necessary here.

oleg68: I changed the names to `LoadFromMultipleFiles` and `LoadFromOneFile`

src/grandorgue/sound/GOSoundProviderWave.h
oleg68: The same as above

src/tools/GOPerfTest.cpp
oleg68: The same as above

src/grandorgue/sound/GOSoundProviderWave.h
oleg68: Fixed

src/grandorgue/sound/GOSoundProviderWave.h
oleg68: Fixed

src/grandorgue/sound/GOSoundAudioSection.cpp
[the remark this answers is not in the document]
larspalo: If they should remain, they should explain that what you're trying to say now. For instance something like:

// dest is the name of the receiving parameter that is a pointer to the actual output sample data to use

src/grandorgue/sound/GOSoundAudioSection.cpp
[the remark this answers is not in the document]
larspalo: Likewise, for instance:

// loop_length is the total length of the loop in samples

src/grandorgue/sound/GOSoundProviderWave.cpp
larspalo: Ok, but still it's Samples in plural! So

LoadSamplesFromFiles

would be better as a description.

larspalo:
@oleg68 Yes, your last versions are much better!

larspalo:
@oleg68 If you're not intending to expand upon the parameter comments, then at least add a TODO to remove the comments that are only temporary work additions when the final commit that resolves the issue is due. But better would be to really add the explanations to the comments that would perhaps be useful in the future!

oleg68:
@larspalo 

The description of parameters should be at the method specification, not at the method call.

Some other programming langualges (ex. Oracle pl/sql, Kotlin etc) allow not only positional notation but also naming notation of passing parameters. It makes the calling code more readable.

Unfortunally c++ does not support naming notation. Putting the parameter names to the comments is the only possible workaround. 

So I insist on parameter names in the comments. I hope it will help other future developers.

larspalo:
> So I insist on parameter names in the comments. I hope it will help other future developers.

And I insist that they are better than they are currently.

rousseldenis:
> The description of parameters should be at the method specification, not at the method call.

Agree on that

oleg68:
> > So I insist on parameter names in the comments. I hope it will help other future developers.
> 
> And I insist that they are better than they are currently.

OK. I added a TODO of removing the parameter name comments in the future after reducing the number of parameters of `DoCrossfade`

larspalo:
@oleg68 Ok, fair enough. Except that the comments are prime examples of poor comment practice, I never saw the need of them to begin with as when hovering over the function call in Eclipse IDE I could see the parameters directly as a reference. Don't know if it works the same in Netbeans though...

larspalo (approved):
We'll just have to remember this later.

oleg68:
> Don't know if it works the same in Netbeans though

Netbeans does not show the parameter names.

> We'll just have to remember this later.

I have already started working on it, but it is not so simple.

larspalo:
@oleg68 Sure, more or less every IDE behaves differently. This is what I see:

Skärmbild från 2023-12-26 21-23-26

> I have already started working on it, but it is not so simple.

I understand that. Thanks for your efforts!

rousseldenis (approved):

PR #1764, opened 2024-01-02
Release 3.13.3

oleg68:
After

- [ ] #1763


oleg68:
@larspalo  @rousseldenis could you approve this pr?

rousseldenis (approved):

larspalo:
Would it be possible to add another task just to update the copyright year in all files within the GrandOrgue project? As we're preparing a new release, and it's actually the year that the release happens on that should be written as copyright year - not the year when the actual code itself was conceived, this should be done before a release.

oleg68:
@larspalo We discussed it last year when we were updating the copyright headers. And we decided that we would update only the headers of the files that would be changed.

larspalo:
@oleg68 Perhaps, I don't have that much of a memory of it. However, reading https://www.gnu.org/licenses/gpl-howto.html it's pretty clearly written under the copyright notice:

> The copyright notice should include the year in which you finished preparing the release

Thus, if a release is finished 2024 the notice should state that too. I'd consider every file that's part of the GO project and included in the release (not an external dependency) to be covered by that statement. (If you read the linked page above it's also clearly stated that it should be for each source file)

If something was done poorly last year, that's no excuse to continue doing it.

oleg68:
@larspalo Please take a look at https://github.com/GrandOrgue/grandorgue/pull/1409#pullrequestreview-1320292767

I'm not a legal expert, so I argue neither pro nor conta the yearly copyright update, so let's come to any agreement.

@rousseldenis Do you still inssist not to update the copyrights in all files yearly?

larspalo:
@oleg68 @rousseldenis I don't want to delay next release unnecessarily, so therefore I've created the #1767 PR that at least changes the visible copyright year in the application. Oleg has already approved of it, and as soon as it is merged I'll also approve of this PR so that we can go on.

However, I really like Olegs suggestion (as I understand it - though I could be wrong) in the mentioned thread #1409. If there could be some kind of automatic update of the headers with each new release for each new year, that would be perfect so that we don't need to think about it.

larspalo (approved):

PR #1781, opened 2024-01-23
Revise Hungarian translation

kerkovits:
- Added translation of new strings.
- Very minor corrections in existing strings.

@briff: Át tudná nézni a fordítást, mint ígérte?

### DIFF
--- a/po/hu.po
+++ b/po/hu.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: GrandOrgue\n"
 "Report-Msgid-Bugs-To: ourorgan-developers@lists.sourceforge.net\n"
-"POT-Creation-Date: 2023-04-20 14:49+0200\n"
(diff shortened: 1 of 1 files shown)


oleg68:
@briff could you check this translation?

rousseldenis:
On other projects, we used to auto approve such things as indeed you need to find a reviewer that speaks the language.

We could maybe envisage a dedicated interface for translations like weblate.

rousseldenis (approved):

larspalo:
@oleg68 @rousseldenis I'm also in favor of approving any translations trusting that other users that do speak the language will provide feedback if anything is wrong. However, as @rousseldenis mentioned with the extension issue, we might need to check which strings are available for translation so that it won't break any functionality.

larspalo (approved):

oleg68 (approved):

PR #1782, opened 2024-01-25
Simplified code of GOPipeConfig and GOPipeConfigNode

oleg68:
This PR:
- Reorders the members and methods of `GOPipeConfig` and `GOPipeConfigNode`
- Introduces template methods `SetSmallMember` and `SetLargeMember`
- Implements all setters with the new template methods

### DIFF
--- a/src/grandorgue/model/pipe-config/GOPipeConfig.cpp
+++ b/src/grandorgue/model/pipe-config/GOPipeConfig.cpp
@@ -10,9 +10,6 @@
 #include "config/GOConfigReader.h"
 #include "config/GOConfigWriter.h"
 
-#include "GOPipeConfigListener.h"
-#include "GOPipeUpdateCallback.h"
-
 GOPipeConfig::GOPipeConfig(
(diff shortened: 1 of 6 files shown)


larspalo (approved):
Personally I'm not that fond of macro usage for function like content (GOPipeConfig.h line 54 and forward) but it seems to work. Is there any definite advantage over writing out the full function content?

oleg68:
> Personally I'm not that fond of macro usage for function like content (GOPipeConfig.h line 54 and forward) but it seems to work. Is there any definite advantage over writing out the full function content?

There are two functions with the same code but with different passing of parameters. Using a macro allows to writhe the code only once.

larspalo:
> Using a macro allows to writhe the code only once.

Yes, I understand how preprocessor directives works by substituting the text. What I asked was if there was any real advantage (beyond not having to write out the body of the function with copy/paste) of using it? In this case, luckily, the macro is found just above the usage of it, which makes it somewhat easier to see/read/understand...

oleg68:

rousseldenis (approved):

Issue #1772, opened 2024-01-10
GO crash on startup


Psi58:
[GOcrash240110.zip](https://github.com/GrandOrgue/grandorgue/files/13888240/GOcrash240110.zip)
New for me, flagging it in case it is a more widespread issue.

Previously (3.12) I had GO running happily on 2 laptops, both running W10.
Installed 3.13, works fine (almost...*) on one, but on the older machine crashes after momentarily showing the GO menu bar.
I've tried installing as admin, cleaning/updating everything on the machine (including "tidy up leftovers" in BC Uninstaller) - no change.
Completely uninstalled GO, tried to reinstall 3.12 (which previously worked) - same issue.

This machine is maxed out with only 128GB SSD and 16GB RAM, so it just may be a resources issue? I've got a more beefy replacement machine on order, so hopefully this isn't a long-term problem for me. But I'm raising it in case whatever the cause is affects other users.

Relevant eventviewer log items attached. They don't mean much to me...

*: 1 sampleset that previously loaded OK now crashes while loading. Still investigating this.

oleg68:
Could you post here a crash dump file? Usually it appears in the `AppData\Local\CrashDumps` subdirectory of the home user directory.

Psi58:
Sure. However, it wasn't in that location (Windows Error Reporting was turned off). 
I've turned it on, and now have the attached in C:\ProgramData\Microsoft\Windows\WER. Is this what you need?
[WER.zip](https://github.com/GrandOrgue/grandorgue/files/13893034/WER.zip)
Thanks.

Larioant:
@oleg68 hi, the same problem happens to me. I attach .dmp file  
[GrandOrgue.exe.9592.dmp](https://github.com/GrandOrgue/grandorgue/files/13917320/GrandOrgue.exe.9592.dmp)

JPIL04:
bonjour,   Même problème....Crash au démarrage . je lance GO, la barre de menu s'affiche , ainsi que la fenêtre avec l'avancement du chargement....et au bout d'une vingtaine de secondes, tout disparait - arrêt de GO avant même que le premier fichier WAV ne se charge. 
Cela est arrivé une diziane de fois , depuis, je crois, la version 12 ou 13....J'utilise l'orgue de Alessandria de Piotr Grabowski. Je n'avais jamais eu de problème avant . Le premier plantage remonte vers le mois de avril, mai 2023 a peu près !
[...]

PR #1783, opened 2024-01-26
Fixed crash on loading an organ with a crescendo in Add mode https://github.com/GrandOrgue/grandorgue/issues/1772

oleg68:
This PR fixes one case of #1772: when there was a crescendo in the add mode.

The reason was the m_ElementStates was not allocated when the crescendo was loaded from a .cmb file.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1784, opened 2024-01-26
Replaced int with GOBool3 in GOCombination and subclasses

oleg68:
Thist PR reduces the memory usage of combinations: from 4 bytes to one byte for each element state.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #1788, opened 2024-01-28
Fixed crash with rtaudio/asio https://github.com/GrandOrgue/grandorgue/issues/1772

oleg68:
This PR fixed a RtAudio bug of incompatibility with thr latest Asio SDK that caused a crash #1772

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed crash with rtaudio/asio https://github.com/GrandOrgue/grandorgue/issues/1772
 - Added support of specifying different crossfade length for additional attack and release files https://github.com/GrandOrgue/grandorgue/issues/1760
 - Fixed wrong filename of the midi recorder file with Hungarian language settings https://github.com/GrandOrgue/grandorgue/issues/1644
 - Fixed unability to select the Hungarian language in the Organ Settings dialog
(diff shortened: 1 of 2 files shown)


larspalo:
@oleg68 Have you submitted the RtAudio fix upstream?

oleg68:
> @oleg68 Have you submitted the RtAudio fix upstream?

No, I haven't. Because RtAudio ships with it's own clone of asio sdk and they avoided this bug by modifying the asio code.

I don't like this way because
- RtAudio has not the most recent asio sdk version
- it contradicts to the ASIO License Agreement 2.4:

> In case the Licensed Software Developer Kit is modified, it shall not be published, sold or
distributed without agreement of Steinberg

So I fixed this bug by modifying RtAudio instead of ASIO. Modyfying RtAudio is not prohibited by [it's license](https://github.com/thestk/rtaudio/blob/b4f04903312e0e0efffbe77655172e0f060dc085/LICENSE):

> including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software

So I wanted to submit the fix upstream, at first I would need to ask the Rt maintainer to remove the asio clone. I'm not ready to do this now.

larspalo (approved):

rousseldenis (approved):

Issue #1602, opened 2023-07-10
Add a --justgui parameter for console panels development


rnditdev:
Hello everyone,

One common theme among samplesets developers is the difficulty of development of the graphic parts of a console. Could a --justgui flag or similar be developed, where GO just loads the panels (And probably shows a watermark indicating this mode)?
Ideally, when the program is in that mode, a function key reloads the whole GUI, or ideally, GO detects the file changes and reloads accordingly.

Best.

rousseldenis:
@rnditdev Thanks for this suggestion.

Maybe --editmode could be a better wording.

However, you can maybe look at https://github.com/larspalo/GOODF development as even if I haven't tested yet, it suggests GUI helper is present in the tool.

@larspalo

larspalo:
@rnditdev @rousseldenis With the latest versions it's now possible to display the actual panel layout in GOODF. Still it's necessary to set the parameters (x, y or row/col) for position control separately for the gui elements as they cannot (yet) be dragged on the panel to desired position.

I should also perhaps mention that GOODF is still very much in development and bugs are likely to exist...

oleg68:
GrandOrgue is not intended to be a graphic console design tool.

@larspalo @rousseldenis should we move in this direction? Or it is better to suggest to use another tools like GOODF and OdfEdit for this?

larspalo:
@oleg68 I think that external tools like GOODF or OdfEdit can handle both organ and graphical panel construction (odf) so that GrandOrgue can stay focused on the VPO rendering and playing experience. Also, it's already possible to set the global sample loading in GO to not load any samples if the user so desires.

jefferycrowley:
I have created several GO ODF using new material and Old HW Sample sets.   I have also produced a full Hauptwerk.xml file, but this is a long and tedious task and all of the features of HW are not yet possible in GO.   Hauptwerk in early development promised a GUI to assist with compilation of graphic images to Hauptwerk.  It has never happened.   

 GO, on the other hand is easier to compile both the ODF and the GUI.   It would be awesome to have a tool that we can use, but it is not crucial to the development of GO.

[...]

PR #1797, opened 2024-02-08
Added capability of loading only GUI panels without audio samples by specifying the "-g" switch from the command line https://github.com/GrandOrgue/grandorgue/issues/1602

oleg68:
Resolves: #1602

This PR
- Reworks parsing the command line
- Implements the `-h` (`--help`) option that displays all possible command line options
- Implements the `-g` (`--gui-only`) option that disables loading sound samples

### DIFF
--- a/src/grandorgue/GOApp.cpp
+++ b/src/grandorgue/GOApp.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2023 GrandOrgue contributors (see AUTHORS)
+ * Copyright 2009-2024 GrandOrgue contributors (see AUTHORS)
  * License GPL-2.0 or later
  * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
  */
@@ -14,12 +14,13 @@
 #include <wx/regex.h>
 #include <wx/stopwatch.h>
 
+#include "config/GOConfig.h"
+#include "sound/GOSound.h"
+
 #include "GOFrame.h"
 #include "GOLog.h"
 #include "GOStdPath.h"
-#include "config/GOConfig.h"
 #include "go_defs.h"
-#include "sound/GOSound.h"
 
 #ifdef __WXMAC__
 #include <ApplicationServices/ApplicationServices.h>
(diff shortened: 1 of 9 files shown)


src/grandorgue/GOApp.cpp
```diff
@@ -140,7 +169,7 @@ bool GOApp::OnInit() {
   SetTopWindow(m_Frame);
   m_Log = new GOLog(m_Frame);
   wxLog::SetActiveTarget(m_Log);
-  m_Frame->Init(m_FileName);
+  m_Frame->Init(m_FileName, m_IsGuiOnly);
```
rousseldenis, line 172: @oleg68 I don't like modes that are too strict. Do you think it should be a good idea to rename this m_LoadingMode? 

This should supports some headless mode, gui only mode, ... 

In a more flexible way

oleg68: Usually I defer adding complexity until it becomes necessary. So I'd leave the boolean `IsGuiOnly` now.

When we introduce another loading features we will rename it to m_LoadingFlags and we will change it's type to `unsigned`. But now I consider making a one-bit mask as unnecessary complexity.

larspalo: Personally I think this "no samples" just gui command line option is close to meaningless (especially considering the GUI displaying/editing capabilities of GOODF). The opposite, a headless mode, would in my opinion be much more usable for hardcore command line users who wish to build a GO driven organ console without any un-necessary resource consumption or screen/monitor.

Many GO users, even including some of the sample set developers, have a really hard time using the command line at all, believe it or not! More "normal" users would likely discover/use this option of loading just the gui if there would be a checkbox in the File->Settings instead that would affect the same thing when checked. Also much nicer to not have to re-start GO to enter/exit the "just gui" mode.

larspalo: @oleg68 @rousseldenis Sorry, scrap the last part of my comment above. It's already possible to achieve this "just-gui" from within the GO gui in the File->Settings dialog by setting the sample loading option to "Don't load" (instead of Stereo or Mono).

Still I cannot but question the value of this command line feature...

oleg68: @larspalo Using `Don't load` is not comfortable because this option is remembered so It needs to be reverted every time when a user wants to load an (another) organ with sound samples.

So I consider the `-g` command-line option as a lightweight solution for temporal disabling loading samples. 

I agree that capability of altering any program setting (not just this) from the command line would be a better solution, but it would require a lot of coding.

> Still I cannot but question the value of this command line feature

I think It is not big enouth for a lot of coding. But this implementation is very simple, so the value/cost relation allows to introduce it.

oleg68:
@larspalo @rousseldenis may be introducing a commond line option `-c ConfigFileName` for altering a config file instead of `-g` would be better?

src/grandorgue/GOApp.cpp
larspalo: > so It needs to be reverted every time when a user wants to load an (another) organ with sound samples.

And using commandline -g, GO needs to be quitted every time the user wants to load another organ with samples! It's exactly the same trouble - only worse and even less comfortable for any normal user!

But since you've already implemented it, and I don't think it will cause any harm, I'll agree to this rather pointless PR.

larspalo:
> may be introducing a commond line option -c ConfigFileName for altering a config file instead of -g would be better?

No, I think you can leave the -g as a "just-gui" option.

Specifying a -c configFile, or a -p presetNumber is certainly viable and could be relevant, but both should perhaps be part of another PR? Or do you want to expand upon this PR?

rousseldenis:
> @larspalo @rousseldenis may be introducing a commond line option `-c ConfigFileName` for altering a config file instead of `-g` would be better?

As @larspalo said, -g should be sufficent at this moment. Maybe, indeed, a specific config file could be an improvement.

oleg68:
> > may be introducing a commond line option -c ConfigFileName for altering a config file instead of -g would be better?
> 
> No, I think you can leave the -g as a "just-gui" option.
> 
> Specifying a -c configFile, or a -p presetNumber is certainly viable and could be relevant, but both should perhaps be part of another PR? Or do you want to expand upon this PR?

No, I don't. It will be a future enchancement.

larspalo (approved):

rousseldenis (approved):

Issue #1657, opened 2023-09-21
Feature Request: Enable CoupleToSubsequent for Coupler Manuals


ahall41:
I recently tried out Coupler Manuals. The idea is that you can turn your 2 manual organ into a virtual 3 or more manual organ?
One problem I noticed is that the Couplers for the manual don't appear to work - presumably because the relevant CoupleToSubsequent... option isn't set for the coupler?

oleg68:
@ahall41 

> Couplers for the manual don't appear to work

Could you explain what dont't work exactly?

ahall41:
Thank you for looking into this. This can be demonstrated on the demo organ
- Select  II/I coupler
- Create unison coupler for coupler manual 1 to first manual
- Play a note on coupler manual. This does not subsequently play the note on manual II.
I'm surprised this feature has been missing for so long, as (I think I'm right) the whole point of the coupling manuals is to play (say) a 3 manual organ on a 2 manual console. Fine, so long as you don't use any couplers.

ahall41:
Having thought about it, I'm not sure how useful this might be, as you can use the built in couplers rather than the organ couplers when the need arises ...

oleg68:
> Thank you for looking into this. This can be demonstrated on the demo organ
> 
> * Select  II/I coupler

Do you mean to press `II/I` button on the organ panel?

> * Create unison coupler for coupler manual 1 to first manual

1. How to do this?
2. What is the diference between `manual 1` and `the first manual`,

ahall41:
1. Start GrandOrgue and open the Demo organ
2. On the main panel ("GrandOrgue Demo V1") select II/I coupler
3. Open Panel -> Coupler -> Coupler Coupling Manual 2
4. Select 8 under First Manual
5. Open Panel -> Coupler Manuals and Volume
6. Select a key on Coupling Manual 2
7. The key is also selected on the main panel (manual I), but is not subsequently coupled to manual II
I hope this is clear ...

oleg68:
@larspalo @rousseldenis @ahall41 how may we fix this issue from the GUI point of view?

Now the coupler panel looks:
![CouplerPanel](https://github.com/GrandOrgue/grandorgue/assets/11586506/90a22859-4baa-42d7-9852-fcd1d29a97b3)

[...]

PR #1810, opened 2024-02-21
Added support of "Couple Through" mode of Virtual Couplers https://github.com/GrandOrgue/grandorgue/issues/1657

oleg68:
Resolves: #1657

This PR enables using recursive coupling for all virtual couplers as discussed in #1657


larspalo:
@oleg68 Seems to be working as intended, but I noticed that the "Couple Through" option is not triggering immediately. It's necessary to re-strike the key for it to take effect both for activation and de-activation. It's perhaps a minor thing and not many would even think anything about it, as normally you'd set the option once for an organ and then forget about it, but I think it would be a nicer behaviour if the couple through action would activate/de-activate immediately even on held keys.

src/grandorgue/GOVirtualCouplerController.cpp
```diff
@@ -123,13 +125,67 @@ void GOVirtualCouplerController::Init(
...
+        cfg,
+        wxString::Format(
+          wxT("SetterManual%03dCoupler%03dThrough"), srcManualN, dstManualN),
+        _("Couple Through"));
+      m_CoupleThroughPtrs[couplerSetKey] = pCoupleThrough;
     }
   }
 }
 
+static wxString WX_COOPLE_THROUGH = wxT("CoupleThrough");
```
larspalo, line 142: perhaps 'COUPLE' instead of 'COOPLE'

src/grandorgue/GOVirtualCouplerController.cpp
```diff
@@ -123,13 +125,67 @@ void GOVirtualCouplerController::Init(
...
+
+void GOVirtualCouplerController::Load(
+  GOOrganModel &organModel, GOConfigReader &cfg) {
+  Init(organModel, cfg);
+  for (auto e : m_CoupleThroughPtrs) {
+    GOCallbackButtonControl *pCoupleThrough = e.second;
+    bool isCoupleThrough = cfg.ReadBoolean(
+      CMBSetting, pCoupleThrough->GetGroup(), WX_COOPLE_THROUGH, false, false);
```
larspalo, line 150: see comment for line 142

src/grandorgue/GOVirtualCouplerController.cpp
```diff
@@ -123,13 +125,67 @@ void GOVirtualCouplerController::Init(
...
+  }
+}
+
+void GOVirtualCouplerController::Save(GOConfigWriter &cfg) {
+  for (auto e : m_CoupleThroughPtrs) {
+    GOCallbackButtonControl *pCoupleThrough = e.second;
+
+    cfg.WriteBoolean(
+      pCoupleThrough->GetGroup(),
+      WX_COOPLE_THROUGH,
```
larspalo, line 162: see comment for line 142

oleg68: Fixed

src/grandorgue/GOVirtualCouplerController.cpp
oleg68: Fixed

larspalo:
@oleg68 Yes, the typo is fixed. Now, what about the other issue I mentioned? The couple through mechanism is not activated/de-activated immediately when pushed (one has to re-strike the key for it to take effect). A nicer behaviour would be that it acted upon existing key presses immediately - it's a bit like what you mentioned in #1813 for the enclosure with zero minimum amplitude.

oleg68:
@larspalo I will try to fix it later.

oleg68:
@larspalo I implemented refreshing the key states when changing `Couple Through` during a key is pressed

src/grandorgue/model/GOCoupler.h
```diff
@@ -66,6 +72,10 @@ class GOCoupler : public GODrawstop {
     int dest_manual,
     GOCouplerType coupler_type);
   void Load(GOConfigReader &cfg, wxString group);
+
+  // send key states for all chaned couplers
```
larspalo, line 76: maybe a typo there?
for all chained couplers

larspalo (approved):
Works beautifully now.

rousseldenis (approved):

Issue #1818, opened 2024-02-26
Code Page mismatch between .organ and .yaml


ahall41:
There seems to be an issue on Windows when the .organ file contains non-ascii (unicode?) data in the labels. I have some imported files from HW which use (in this case) the wrong `'` character (see attached [Goch.zip](https://github.com/GrandOrgue/grandorgue/files/14402987/Goch.zip) Goch.organ - e.g. Stop011). On Linux if you save the combinations it uses the same character, but on Windows it seems to treat the file as ASCII rather than UTF8 [?]. As a consequence you can get a warning when transferring such files between Linux and Windows.

PR #1821, opened 2024-03-01
Always save exported combinations with UTF-8-BOM

nanoufo:
Resolves: #1818.
The problem occurred due to the utilization of the default system encoding for conversions between wxString and strings in YAML. yaml-cpp treats strings as a sequence of bytes and does not perform encoding conversions.

Now UTF-8 encoding is used for conversions. The exported file now begins with a UTF-8 BOM mark.
The UTF-8 BOM mark is used to differentiate documents exported from older GO versions (which use UTF-8 without BOM on Linux/Mac and one of the one-byte encodings on Windows) and to indicate to text editors to use UTF-8 encoding.

<s>While this change has not been tested on Windows yet, it is expected to function properly.</s> <s>Import is broken on Windows, investigating now.</s> Tested on Linux & Windows.

### DIFF
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -7,6 +7,7 @@
 
 #include "GOOrganController.h"
 
+#include <algorithm>
 #include <math.h>
 #include <wx/datetime.h>
 #include <wx/filename.h>
@@ -700,13 +701,16 @@ static std::vector<char> loadFileBytes(const wxString &file) {
 }
 
 static wxString loadFileTextWithEncodingDetection(const wxString &file) {
-  std::vector<char> content = loadFileBytes(file);
-  wxConvAuto conv(wxFONTENCODING_SYSTEM); // utf-8 if BOM is present, system otherwise
-  wxString contentAsText = wxString(&content[0], conv, content.size());
-  if (contentAsText.IsEmpty() && content.size() != 0) {
-    throw wxString::Format(_("Failed to guess file encoding: %s"), file);
+  std::vector<char> content = loadFileBytes(file);  
+  std::array<char, 3> utf8bom = {'\xEF', '\xBB', '\xBF'};
+  if (content.size() >= utf8bom.size() && std::equal(utf8bom.begin(), utf8bom.end(), content.begin())) {
+    // Newer GO versions export yaml files with UTF-8-BOM.
+    // wxConvAuto will detect BOM and decode as UTF-8.
+    return wxString(&content[0], wxConvAuto(), content.size());
+  } else {
+    // Use encoding that was used in older GO versions (system default)
+    return wxString(&content[0], *wxConvCurrent, content.size());
   }
-  return contentAsText;
 }
 
 void GOOrganController::LoadCombination(const wxString &file) {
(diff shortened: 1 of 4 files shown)


src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
   return isToImport;
 }
 
+static std::vector<char> loadFileBytes(const wxString &file) {
```
oleg68, line 693:
```suggestion
static std::vector<char> load_file_bytes(const wxString &fileName) {
```
due the GO naming conventions about out-of-class functions

src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
...
+  wxFileInputStream fileInputStream(file);
+  if (!fileInputStream.IsOk()) {
+    throw wxString::Format(_("Failed to open '%s'"), file);
```
oleg68, line 696: We need a more detailed error message: why?

src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
...
+  }
+  std::vector<char> content(fileInputStream.GetSize());
+  if (!fileInputStream.ReadAll(&content[0], content.size())) {
+    throw wxString::Format(_("Failed to read '%s'"), file);
```
oleg68, line 700: More detailed error message is needed

src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
...
+  }
+  return content;
+}
+
+static wxString loadFileTextWithEncodingDetection(const wxString &file) {
+  std::vector<char> content = loadFileBytes(file);
+  std::array<char, 3> utf8bom = {'\xEF', '\xBB', '\xBF'};
```
oleg68, line 707: Could you make a single constant and share it with line 647?

src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
...
+  if (
+    content.size() >= utf8bom.size()
+    && std::equal(utf8bom.begin(), utf8bom.end(), content.begin())) {
```
oleg68, line 710: If the file contains soma another BOM (not utf8), this method tries to use *wxConvCurrent instead of wxConvAuto()

src/grandorgue/GOOrganController.cpp
```diff
@@ -686,6 +690,33 @@ bool GOOrganController::IsToImportCombinationsFor(
...
+    throw wxString::Format(_("Failed to open '%s'"), file);
+  }
+  std::vector<char> content(fileInputStream.GetSize());
+  if (!fileInputStream.ReadAll(&content[0], content.size())) {
+    throw wxString::Format(_("Failed to read '%s'"), file);
+  }
+  return content;
+}
+
+static wxString loadFileTextWithEncodingDetection(const wxString &file) {
```
oleg68, line 705:
```suggestion
static wxString load_file_text_with_encoding_detection(const wxString &fileName) {
```

oleg68 (changes requested):

oleg68 (changes requested):

oleg68 (approved):

rousseldenis (approved):

Issue #1385, opened 2023-02-12
Feature request: to support silent attack pipe in a rank


eturpault:
In order to implement stop disengage or key release noise, it is currently necessary with GrandOrgue to define an attack pipe refering to a silent sample. This method is used by Piotr Grabowski and OdfEdit v2.1/2.2.
In order to avoid to have to include in the sample set a wav file with no sound inside (so to play a silent attack sample), this request is to add in GO the possibility to name a pipe attack as "SILENT" in a rank so that GO doesn't emit attack sound for this pipe and can emit release sound for the same pipe. The syntax should be : **Pipe999=SILENT**
Pipe999=DUMMY cannot be used in this case since the pipe release sample is dummy as well.

Example of pipe new definition in this case : 
```
Pipe001=SILENT
Pipe001ReleaseCount=1
Pipe001Release001=Noises\Releases\036-c.wav
```
instead of currently : 
```
Pipe001=Noises\SilentLoop.wav
Pipe001ReleaseCount=1
Pipe001Release001=Noises\Releases\036-c.wav
```

oleg68:
Should GO play the release sample in this case?

eturpault:
Yes correct. This should be a way to have a pipe which the attack is silent (not played) and the played release is the one given in the Pipe999Release999 attribute.
With the setting Pipe999=DUMMY I observed that the release sample of this pipe is ignored.

larspalo:
I think we should implement this in a better way. Looking at the real use case (effect stops like tracker action and stop noises) we should have a kind of hybrid between the Percussive=Y and Percussive=N. I think that another boolean setting could be added to the rank (or pipe) that specify that (if true) it _can_ have both attacks and releases but will play them back without any loops, basically just a percussive pipe _with_ release(s). We could name it "PercussiveRelease", "NoiseEffect", "UseRelease" or whatever. What that would do is enable us to keep both parts of the noise on and noise off in the same rank and avoid having to resort to tricks like multiple ranks, silent loops or looping the last sample of the attack etc. This should of course work both for tracker and stop noises (when only one single pipe is defined for the rank, or if multiple are for the former case).

eturpault:
Indeed in this use case the need is to play a percussive release sample without attack sample in the stop (the attack sample is managed in another stop).
[...]

PR #1828, opened 2024-03-06
Added capability of playing release of percussive pipes with Pipe999HasIndependentRelease=Y https://github.com/GrandOrgue/grandorgue/issues/1385

oleg68:
Resolves: #1385

This PR introduces the new Pipe999HasIndependentRelease and HasIndependentRelease keys. They are used only with Pipe999Percussive=Y and Percussive=Y.

The main two changes for HasIndependentRelease:
- GOSoundingPipe::VelocityChanged() starts playing a release sample instead of stopping playing the attack sample
- GOSoundEngine::StartSample() now can start a release sample as well as an attack sample depending on the `isRelease` parameter value

### DIFF
--- a/help/grandorgue.xml
+++ b/help/grandorgue.xml
@@ -9913,9 +9913,22 @@ on which the pipes of the are placed.
           <term>Percussive</term>
           <listitem>
             <para>
-(boolean, required) If false, the samples are played as is (without
-any loop/release handling)
-     </para>
+	      (boolean, required) If false, the samples are played as is
+	      (without any loop/release handling).
+	    </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>HasIndependentRelease</term>
+          <listitem>
+            <para>
+	      (boolean, default false or inherited from the Organ, Windchest or
+	      Rank). If true than release is played independently on the attack.
+	    </para>
+            <para>
+	      If HasIndependentRelease is true then Percussive must also be set
+	      to true.
+	    </para>
           </listitem>
         </varlistentry>
         <varlistentry>
@@ -9955,12 +9968,36 @@ attributes are:
       </para>
       <variablelist>
         <varlistentry>
+          <term>Pipe999</term>
+          <listitem>
+            <para>
+(string, required) Relative path to the sample WAV file of the first
+attack. It may be listed as REF:aa:bb:cc too. In that case, it means
+that this pipe is borrowed from manual aa, first rank of stop bb, the pipe cc.
+It may contain DUMMY, which defines a non-sounding placeholder.
+     </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
           <term>Pipe999Percussive</term>
           <listitem>
             <para>
 (boolean, default: rank percussive setting) If false, the samples are
 played as is (without any loop/release handling)
      </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>Pipe999HasIndependentRelease</term>
+          <listitem>
+            <para>
+	      (boolean, default: rank HasIndependentRelease setting). If true
+	      then the release is played independently on the attack.
+	    </para>
+            <para>
+	      If HasIndependentRelease is true then Pipe999Percussive must be
+	      also true and Pipe999ReleaseCount must be greater than 0.
+	    </para>
           </listitem>
         </varlistentry>
         <varlistentry>
@@ -9996,17 +10033,6 @@ number of cents (in addition to the organ/rank factor).
             <para>
 (integer 0 - 10000, default: 0) Delay introduced by the tracker for
 that pipe.
-     </para>
-          </listitem>
-        </varlistentry>
-        <varlistentry>
-          <term>Pipe999</term>
-          <listitem>
-            <para>
-(string, required) Relative path to the sample WAV file of the first
-attack. It may be listed as REF:aa:bb:cc too. In that case, it means
-that this pipe is borrowed from manual aa, first rank of stop bb, the pipe cc.
-It may contain DUMMY, which defines a non-sounding placeholder.
      </para>
           </listitem>
         </varlistentry>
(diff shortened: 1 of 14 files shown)


help/grandorgue.xml
```diff
@@ -9913,9 +9913,22 @@ on which the pipes of the are placed.
...
+	      (boolean, required) If false, the samples are played as is
+	      (without any loop/release handling).
+	    </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>HasIndependentRelease</term>
+          <listitem>
+            <para>
+	      (boolean, default false or inherited from the Organ, Windchest or
```
larspalo, line 9925: It's not really inherited from Organ or Windchest, right? It's only inherited from the Rank level and down to the pipe level, which is very much fine for me, but the wording of the help text should match reality I think.

help/grandorgue.xml
```diff
@@ -9913,9 +9913,22 @@ on which the pipes of the are placed.
...
+	      Rank). If true than release is played independently on the attack.
```
larspalo, line 9926: If true, releases are played independently of the attacks.

Both because this applies to multiple pipes and also since there can be more than one attack/release each pipe.

help/grandorgue.xml
```diff
@@ -9963,6 +9987,19 @@ played as is (without any loop/release handling)
      </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>Pipe999HasIndependentRelease</term>
+          <listitem>
+            <para>
+	      (boolean, default: rank HasIndependentRelease setting). If true
+	      then the release is played independently on the attack.
```
larspalo, line 9995: If true, then releases are played independently of attacks in this pipe.

larspalo (changes requested):

help/grandorgue.xml
larspalo: @oleg68 Technically, this PR works beautifully in any test I've thrown at it. As soon as the help text is corrected I'll approve.

help/grandorgue.xml
oleg68: Both `Percussive` and `HasIndependentRelease` may be also defined at the Organ or Windchest level. For some kind of instruments (ex. harpsichords) it may be useful.

larspalo: >  For some kind of instruments (ex. harpsichords) it may be useful.

Maybe. But as far as I've tested it it's not possible with this PR to write such lines to [Organ] or [Windchest999] sections, so I suggest just to change the help text like I mentioned above.

help/grandorgue.xml

help/grandorgue.xml

help/grandorgue.xml
larspalo: @oleg68 Ok, so now it would be possible to actually set the Percussive= and HasIndependentRelease= in the [Organ] and [Windchest999] sections and have it inherited to rank and pipe level. But it's not limited at that.

Both the `Percussive` and `HasIndependentRelease` now needs to be added to the [Organ] and [Windchest999] help sections too since they are available to set through the odf. Also, for the Windchest help section all the PitchTuning, PitchCorrection etc that comes with the GOPipeConfig must also be documented as they suddenly now are available for usage in an odf.

larspalo: @oleg68 Just to be clear, I'm not at all opposed to this development, I just want it well documented as it's a quite big change. I've also prepared GoOdf to be able to handle all the new properties, but I'll wait with pushing any changes until this PR is merged.

larspalo (changes requested):
In addition to my earlier comment:

`Both the Percussive and HasIndependentRelease now needs to be added to the [Organ] and [WindchestGroup999] help sections too since they are available to set through the odf. Also, for the Windchest help section all the PitchTuning, PitchCorrection etc that comes with the GOPipeConfig must also be documented as they suddenly now are available for usage in an odf.`

Also the Percussive help section for a rank would need adjustment, right?

oleg68:
@larspalo I documented the capability of specifying all PipeConfig items at the Windchest section and added Percussive and HasIndependentRelease to the Organ section.

larspalo (approved):
Thanks Oleg! This PR is actually doing a lot more than the title suggests. Not only can we now create different noise effects with much greater effiency (less stops/ranks needed) it's also much cleaner and obvious what happens with the attack and release as well as the need for a silent sample file is eliminated.

Another big change is the inheritance chain of Percussive (and HasIndependentRelease) that now go from Organ, WindchestGroup, Rank to Pipe which it didn't earlier. A side effect of this is the addition of odf settings for AmplitudeLevel, Gain, PitchTuning, PitchCorrection and TrackerDelay to WindchestGroups. The later have existed as cmb settings for a long time, but now we have it in the odf too.

All in all, I think this is a significant step forward. Thanks again!

oleg68:
I think the side effect requires a separare line in  changelog.md. I'll add it before merge

oleg68:
> Another big change is the inheritance chain of Percussive (and HasIndependentRelease) that now go from Organ, WindchestGroup, Rank to Pipe which it didn't earlier.

My dream is to put `IsTremulant` to the same inheritance chain: #878

larspalo:
> My dream is to put IsTremulant to the same inheritance chain

I know. But as I've explained at length in #878 it's inherently a mis-use of the rank feature to define different versions of states for a single pipe as different ranks - even if it's done by some it's still wrong! A rank is by definition a (single) line of actual physical pipes.

In the technical aspects of organ construction however, the tremulant usually do belong at the windchest level, and sometimes even at the main wind supply and thus affecting the whole organ - which is why one tremulant can "affect/be added to" many windchestgroups. Normally, the tremulant affects/disturbs the wind flow to all the pipes on the windchest where it's placed simultaneously. This works fine with the synthetic tremulant model even if that needs improvement too. For the sampled tremulant it's another story... Personally, I still feel that the Pipe999IsTremulant or even better Pipe999Attack999IsTremulant is a reasonable way to model sampled tremulants as different versions of the same pipe.

oleg68:

oleg68:

PR #1832, opened 2024-03-08
Rewrote GOSelectOrganDialog with the GOSimpleDialog

oleg68:
This is a preparatory PR for the next commit for #1663

This PR reimplementing the Organ Selection dialog with `GOSelectOrganDialog`.

As a result
- it receives the right icon
- it's position and size are stored
- it receives the `Help` button

### DIFF
--- a/src/grandorgue/GOFrame.cpp
+++ b/src/grandorgue/GOFrame.cpp
@@ -18,6 +18,7 @@
 #include <wx/sizer.h>
 #include <wx/spinctrl.h>
 #include <wx/splash.h>
+#include <wx/textctrl.h>
 #include <wx/toolbar.h>
 
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


larspalo:
@oleg68 Ok, the loading dialog size is remembered now, which is good. But the user selected width of the columns are not. It would be better to have a more dynamic layout that expand to fit available width instead of the absolute column widths currently used.

oleg68:
@larspalo I'm planning to implement saving the column sizes in a future PR (for both this dialog and Settings/Organs). This PR only replaces the diagog class and does nothing with the internal elements.

Do you agree that GO behavior has not become worse with this PR? If yes, please approve this PR for I could submit a next enchancement.

larspalo (approved):
I'll approve as this is clearly one step towards a better implementation and it's not worse than the current one!

oleg68:
@rousseldenis could your approve this pr?

rousseldenis (approved):

PR #1837, opened 2024-03-10
Fixed loading controller without GOSetter

rousseldenis:
As the controller methods should access to Modifiers like 'SetOrganModified', GOSetter object should be initialized without calling ReadConfigFile() method

Following https://github.com/GrandOrgue/grandorgue/pull/1268#discussion_r1500375471

### DIFF
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -123,6 +123,7 @@ GOOrganController::GOOrganController(
     m_MainWindowData(this, wxT("MainWindow")) {
   GOOrganModel::SetMidiDialogCreator(pMidiDialogCreator);
   GOOrganModel::SetModelModificationListener(this);
+  m_setter = new GOSetter(this);
   m_pool.SetMemoryLimit(m_config.MemoryLimit() * 1024 * 1024);
 }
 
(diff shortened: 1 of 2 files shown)


src/grandorgue/GOOrganController.cpp
```diff
@@ -123,6 +123,7 @@ GOOrganController::GOOrganController(
     m_MainWindowData(this, wxT("MainWindow")) {
   GOOrganModel::SetMidiDialogCreator(pMidiDialogCreator);
   GOOrganModel::SetModelModificationListener(this);
+  m_setter = new GOSetter(this);
```
oleg68, line 126: In your case the `GOSetter` are created twice: in the GOOrganController constructor and in the `ReadOrganFile` (line 252).

So if you add initialising m_setter here, please, remove it from the line 252

src/grandorgue/combinations/GOSetter.cpp
```diff
@@ -1213,7 +1213,8 @@ void GOSetter::SetTranspose(int value) {
 }
 
 void GOSetter::UpdateModified(bool modified) {
-  m_buttons[ID_SETTER_SAVE_SETTINGS]->Display(modified);
+  if (m_buttons.size() != 0)
```
oleg68, line 1216:
```suggestion
  if (m_buttons.size())
```

oleg68 (approved):

oleg68 (approved):

larspalo (approved):

PR #1842, opened 2024-03-15
Added capability of saving inner dialog sizes

oleg68:
As discussed in #1830 and #1832, now GransOrgue remembers dialog windoow external sizes, but does not remember column sizes.

This PR 
- moves all size-related classes to the `size` subdirectory
- introduces the abstract class `GOAdditionalSizeKeeper` for storing additional sizes. GOSizeKeeper implements this class.
- introduces methods `ApplyAdditionalSizes` and `CaptureAdditionalSizes` to GODialog and GODialogTab. Their successors can override them for (re)storing inner element sizes.

This PR adds a capability for future usage. No GO behavior should be changed now.


src/grandorgue/gui/GOGUISizeKeeper.cpp
```diff
@@ -21,6 +21,15 @@ const wxString WX_WINDOW_WIDTH = wxT("WindowWidth");
 const wxString WX_WINDOW_HEIGHT = wxT("WindowHeight");
 const wxString WX_DISPLAY_NUMBER = wxT("DisplayNumber");
 const wxString WX_WINDOW_MAXIMIZED = wxT("WindowMaximized");
+const wxString WX_N_ADD_SIZES = wxT("NAddSizes");
```
larspalo, line 24: Most other cmb settings have "NumberOf" as prefix and one has Count suffix (SavedCount= used for dialog sizes). I can't help but feel that the naming used here is strange and unnecessarily shortened. Maybe to be consistent with the other dialog sizes "AdditionalSizeCount" would be better.

src/grandorgue/gui/GOGUISizeKeeper.cpp
```diff
@@ -21,6 +21,15 @@ const wxString WX_WINDOW_WIDTH = wxT("WindowWidth");
...
+const wxString WX_ADD_SIZE_KEY_FMT = wxT("AddSize.%03d.key");
```
larspalo, line 25: In all the following places that "Add" or "ADD" is used, I'd greatly prefer "Additional" or "ADDITIONAL" and where "N" or "n" is used the full "NUMBER" or "number".

oleg68:
@larspalo I changed `Add` to `Additional` and `N` to `Count`.

larspalo (approved):
Thanks.

src/grandorgue/CMakeLists.txt
```diff
@@ -122,7 +122,6 @@ gui/GOGUIPanel.cpp
 gui/GOGUIPanelWidget.cpp
 gui/GOGUIRecorderPanel.cpp
 gui/GOGUISequencerPanel.cpp
-gui/GOGUISizeKeeper.cpp
```
rousseldenis, line 125: I just wondering why moving from gui to size (as this could be ambiguous about what the size is for)

rousseldenis (approved):

src/grandorgue/CMakeLists.txt
oleg68: Because this PR adds more files related with sizes. I haven't found any other short word for describing what size is for. GUI is not more accepted after moving.

PR #1843, opened 2024-03-16
Increased maximum allowed (panel) size to 32000.

### LINKED ISSUE #1840: size limit for images in GO
Yesterday I tried to convert a new sample set (http://www.pipeloops.com//product_info.php?products_id=64) to GO. It uses bitmaps with the size 5760 x 3336, which is obviously too large for GO (afaik the max size is 4000). Since HD pictures are much bigger than the ones in old days, it might be the time to  enlarge the upper limit for image sizes?
It is a bit  annoying to rescale the images and calculate all the position parameters in the ODF again, although Lars' GoODF helps a bit with this job.
Oliver

larspalo:
Resolves #1840

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Increased maximum value of allowed size to 32000 https://github.com/GrandOrgue/grandorgue/issues/1840
 - Added capability of specifying AmplitudeLevel, Gain, PitchTuning, PitchCorrection and TrackerDelay, Percussive, HasIndependentRelease at the WindchestGroup level of ODF
 - Added capability of specifying Percussive and HasIndependentRelease at the Organ level of ODF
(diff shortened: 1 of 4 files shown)


oleg68:
@larspalo It looks all right, but seems `gui/GOGUISizeKeeper.cpp`, line 16 should be adjusted as well.

larspalo:
@oleg68 Right, I never tried to save any config. Also the tested panel/image size of 5760 would not have triggered an issue with that. I'll change that to 32000 too.

oleg68 (approved):

oleg68:
@larspalo please resolve the conflict.

larspalo:
@rousseldenis Could you please approve of this PR, it's just changing a simple allowed panel width number... Not very much to think about!

rousseldenis (approved):

Issue #704, opened 2021-08-18
Software update


oleg68:
### Discussed in https://github.com/oleg68/GrandOrgue/discussions/51

<div type='discussions-op-text'>

<sup>Originally posted by **Jean-Gabriel1982** August  5, 2021</sup>
Would it be possible to implement a software update module, in GrandOrgue, rather than checking each time for possible updates on Github?</div>

rousseldenis:
> It is too difficult to implement a real software updater (it would required the administrator privileges that ordinar users running GO do not have). But we can implement a optional checking for update with displaying a message that the new release exist.
> 
> This should be switchable on/off in the settings

One thing we could do is publish apt packages. So, the automatic updates can come.

oleg68:
@rousseldenis Yes, publishing repositories for different linux distributions (like Martin did on OBS) would help to update. But this solution is applicable only for linux.

There are lots of users with windows and OSx that also want to receive notifications about the new releases.

So I this issue is about checking for the new release for all platforms.

Making OBS repositories is a subject of a separate issue.

rousseldenis:
> @rousseldenis Yes, publishing repositories for different linux distributions (like Martin did on OBS) would help to update. But this solution is applicable only for linux.
> 
> There are lots of users with windows and OSx that also want to receive notifications about the new releases.
> 
> So I this issue is about checking for the new release for all platforms.
> 
> Making OBS repositories is a subject of a separate issue.

Yes, such feature for windows for instance needs something like a static url to check for new versions and popup user if exists.

PR #1844, opened 2024-03-16
Added checking for updates at startup

nanoufo:
popup

### Visible changes
At startup, GO asynchronously checks GitHub releases for updates. If it finds a new version, it displays a modal dialog with the latest release and its description. Clicking on the download button opens the browser with the releases page. 
If no updates are found, a message is added to the Log Messages window.

By default, update checking is enabled. However, if these pop-ups become annoying, they can be quickly disabled by unchecking the 'Check for updates at startup' checkbox. Update checking can also be enabled or disabled in the settings on the Options tab.

Resolves: #704 .

### Implementation
The application uses libcurl for calls to the GitHub API as wxHttp cannot be used due to its lack of TLS support. On Linux, the system libcurl is utilized, while for Windows, a pre-built libcurl with all dependencies is downloaded from the MSYS2 repositories ( is this acceptable?).

### Testing
The application has been tested on both Linux and Windows platforms. Binaries were built on a local machine with a modified GO version. Building for MacOS seems to work in Github Actions, but it has not been tested yet.


src/grandorgue/config/GOConfig.cpp
```diff
@@ -215,6 +215,8 @@ GOConfig::GOConfig(wxString instance)
     MidiRecorderPath(
       this, wxT("General"), wxT("MIDIRecorderPath"), wxEmptyString),
     MidiPlayerPath(this, wxT("General"), wxT("MIDIPlayerPath"), wxEmptyString),
+    CheckForUpdatesAtStartup(
```
oleg68, line 218: @rousseldenis @larspalo do you agree with enabling checking for updates by default? If most of organs do not have an internet connnection, is it better set to false?

src/grandorgue/dialogs/settings/GOSettingsOptions.cpp
```diff
@@ -407,6 +407,16 @@ GOSettingsOptions::GOSettingsOptions(GOConfig &settings, wxWindow *parent)
   item6->Add(grid, 0, wxEXPAND | wxALL, 5);
   item9->Add(item6, 0, wxEXPAND | wxALL, 5);
 
+  item6 = new wxStaticBoxSizer(wxVERTICAL, this, _("&Updates"));
+  item9->Add(item6, 0, wxEXPAND | wxALL, 5);
+  item6->Add(
+    m_CheckForUpdatesAtStartup = new wxCheckBox(
+      this, ID_COMPRESS_CACHE, _("check for updates at startup")),
```
oleg68, line 414: ID_COMPRESS_CACHE is a bad identifier here. Either introduce a new one or use wxID_ANY if it is not necessary.

src/grandorgue/updater/GONewReleaseDialog.cpp
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2024 GrandOrgue contributors (see AUTHORS)
```
oleg68, line 2: What will the copyright updater precommit hook do in next year? Will it replace with `2025` or with `2024-2025`?

@rousseldenis @larspalo what is our policy for copyright headers in new files? Must have they content 2009 or not?

larspalo:
Personally, I don't like that the log window pops up with the notification that no new release could be found. If nothing could be found - it should be silent. This current behaviour is a main concern and very much un-acceptable for me.

Another, perhaps smaller concern, but still quite important, in my opinion, is that it's the user who should trigger the check for updates, for instance by selecting it in a menu or something similar (usually this would be Help->Check for updates). This current popup is much to invasive for my taste. A small notification line at the bottom in the about window would be perfect - just something like "New version is available: 3.14.0". Then if the user selects the Help->Check for updates the big popup could fire...

In short, I think that:

1. No log window popup at startup at all - especially not the current that only infom that no actual update exists!
2. Rather a menu option like Help->Check to manually trigger check for updates, get the big popup with info
3. Notification of eventual positive automatic checking results (that updates exist) should be discreet and non-invasive

src/grandorgue/updater/GONewReleaseDialog.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#ifndef GONEWRELEASEDIALOG_H
+#define GONEWRELEASEDIALOG_H
+
+#include "config/GOConfig.h"
+#include "updater/GOUpdateChecker.h"
+#include <wx/wx.h>
+
+class GONewReleaseDialog : public wxDialog {
```
oleg68, line 14: Please use the dialogs/common/GOSimpledialog for the base class instead of wxDialog.

See an example in dialogs/GOOrganSettingsDialog as an example of how to add a custom button.

src/grandorgue/updater/GONewReleaseDialog.h
```diff
@@ -0,0 +1,26 @@
+/*
```
oleg68, line 1: Please move this file to `dialogs`

oleg68 (changes requested):

larspalo (changes requested):
Personally, I don't like that the log window pops up with the notification that no new release could be found. If nothing could be found - it should be silent. This current behaviour is a main concern and very much un-acceptable for me.

Another, perhaps smaller concern, but still quite important, in my opinion, is that it's the user who should trigger the check for updates, for instance by selecting it in a menu or something similar (usually this would be Help->Check for updates). This current popup is much to invasive for my taste. A small notification line at the bottom in the about window would be perfect - just something like "New version is available: 3.14.0". Then if the user selects the Help->Check for updates the big popup could fire...

In short, I think that:

1. No log window popup at startup at all - especially not the current that only infom that no actual update exists!
2. Rather a menu option like Help->Check to manually trigger check for updates, get the big popup with info
3. Notification of eventual positive automatic checking results (that updates exist) should be discreet and non-invasive

src/grandorgue/config/GOConfig.cpp
larspalo: @oleg68 Perhaps, but I'd rather not have any notification at all if the checking would fail for any reason.

src/grandorgue/updater/GONewReleaseDialog.cpp
larspalo: @oleg68 Using a range like 2009-2024 is perhaps best describing the whole software release (the copyright notice according to GNU-GPL is tied to finished releases, not when the code was written). https://www.gnu.org/licenses/gpl-howto.html

src/grandorgue/CMakeLists.txt
```diff
@@ -225,7 +227,7 @@ endif()
 add_library(golib STATIC ${grandorgue_src})
 set(go_libs ${wxWidgets_LIBRARIES} ${YAML_CPP_LIBRARIES} ${RT_LIBRARIES} ${PORTAUDIO_LIBRARIES} ${FFTW_LIBRARIES} ${ZITACONVOLVER_LIBRARIES})
 set(go_libdir ${wxWidgets_LIBRARY_DIRS} ${RT_LIBDIR} ${PORTAUDIO_LIBDIR} ${FFTW_LIBDIR})
-target_link_libraries(golib GrandOrgueImages GrandOrgueCore ${go_libs})
+target_link_libraries(golib GrandOrgueImages GrandOrgueCore ${go_libs} CURL::libcurl)
```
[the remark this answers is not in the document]
nanoufo: When using the target name instead of  `${CURL_LIBRARIES}`, include directories [are added automatically](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements).

> Usage requirements are propagated by reading the INTERFACE_ variants of target properties from dependencies and appending the values to the non-INTERFACE_ variants of the operand. For example, the [INTERFACE_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_INCLUDE_DIRECTORIES) of dependencies is read and appended to the [INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INCLUDE_DIRECTORIES.html#prop_tgt:INCLUDE_DIRECTORIES) of the operand.

src/grandorgue/config/GOConfig.cpp
rousseldenis: Yes, agree with @larspalo too on notifications.

src/grandorgue/updater/GONewReleaseDialog.cpp
rousseldenis: Yes, the gnu website ended by saying we can decide and explain what to do. IMHO, new files could mention 2024 only and then modifications will change to 2024-2025 for instance (that's the behavior I use on other open source projects), but we can decide to use the same range for the whole files.

rousseldenis:
I would say notifications should be done silently by adding a text or icon to the main bar for instance:

image

And I tend to opt for a manual check too in the Help section

src/grandorgue/GOApp.cpp
```diff
@@ -195,6 +201,11 @@ int GOApp::OnExit() {
 }
 
 void GOApp::CleanUp() {
+  // Make sure update checker is completed
+  if (m_updateCheckerThread.joinable()) {
+    m_updateCheckerThread.join();
```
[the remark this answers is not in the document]
nanoufo: > What happens if there is no internet connection available? Can GrandOrgue shutdown in this case?

 An exception will be thrown after the timeout of 5 seconds expires. The exception will be caught and logged. See GOUpdateChecker.cpp.
After that GO will close.

src/grandorgue/updater/settings.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef UPDATER_SETTINGS_H
+#define UPDATER_SETTINGS_H
+
+// Don't use long timeouts here since we wait for completion on application
+// exit. Also users may become annoyed if 'New release!' dialog appears while
+// they are playing.
+const long TIMEOUT_MS = 5000;
+const char *const USER_AGENT_HEADER = "User-Agent: GrandOrgue";
+const char *const LATEST_RELEASE_API_URL
```
oleg68, line 15: These two urls should not be hardcodes in the c++ source code files. Make them adjustable with cmake and src/core/go_defs.h

oleg68 (changes requested):

oleg68:
I suggest.

- Not to enable checking for update by default
- Once a user enabled it, the checking should be done automatically every time GrandOrgue is started
- Implementing a special menu action requires more UI elements: showing a progress window with the result. I would not implement it now.
- I agree with @larspalo that no log window should appear if there are no updates available
- If there is a newer version, a modal (or may be modalless) window should be opened. I the user does not like it, ha/she won't enable the update checking.
- I would not add more elements ti the main organ toolbar because we have already lots of them and it there is no copablity to remove some of them. At first, we should make the main toolbar customizable.

@rousseldenis @larspalo do you agree with this approach?

src/grandorgue/updater/GOUpdateChecker.h
```diff
@@ -0,0 +1,60 @@
...
+
+struct GOUpdateCheckerResult {
+  GOReleaseMetadata latestRelease;
+  bool updateAvailable;
+};
+
+class UpdateCheckerCompletedEvent;
+wxDECLARE_EVENT(UPDATE_CHECKER_COMPLETED, UpdateCheckerCompletedEvent);
+
+class UpdateCheckerCompletedEvent : public wxEvent {
```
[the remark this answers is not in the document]
nanoufo: Do these rules apply to implementation-private classes that are not exposed in headers?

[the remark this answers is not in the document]
oleg68: > Do these rules apply to implementation-private classes that are not exposed in headers?

No, they don't. You may put several classes in a single .cpp file.

larspalo:
@oleg68 Your suggestions would give an acceptable behaviour even though a separate menu item and a more manual procedure of checking with only discreet automatic hints would be a better general approach, I think. However, I'd like to add to the requirements that even if the automatic update checking is enabled - a failed check, for instance if no internet is available should be handled gracefully.

nanoufo:
As of now,

1. The update checker posts debug messages that may be seen in Log Messages window in GO debug builds.
2. Update checking may be performed at startup. If it fails, it does so silently. If a new update is found, the user will be notified by the addition of a new menu. The menu contains a 'Download' item that opens the releases page in the browser. Other items simply open a large popup window.
   screenshot
4. A menu item labeled 'Check for Updates' can be found in the 'Help' menu. Clicking on it initiates the update checking process. While the checking is in progress, the mouse cursor displays a 'busy' icon. If a new update is detected, a large popup window is displayed. Otherwise, a wxMessageBox is shown to describe the results.

oleg68:
I tested the new version.

The new menu item is acceptable.

But when the menu item shows update v3.12.0-0.local -> 3.13.3-1, the `Show changelog` displays only the changelog of the lastt release only: 3.13.3-1.

It would be better to show changelog of all versions from 3.12.1 -> 3.13.3-1.

oleg68 (changes requested):

src/grandorgue/updater/GOUpdateChecker.h
```diff
@@ -0,0 +1,86 @@
...
+      return *this;
+    }
+  };
+
+  class CompletionEvent : public wxEvent {
+  private:
+    Result m_result;
+
+  public:
+    CompletionEvent(wxEventType eventType, Result result)
```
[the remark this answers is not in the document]
nanoufo: The result is copied to a field, making at least one copy inevitable. Using 'Result' instead of 'const Result&' in the signature suggests that a copy is made. To prevent additional copies, 'std::move' is utilized.

larspalo:
I like the current version, there are some small things that Oleg already pointed out (some of which are more important than others)... But, as a whole I like the approach and can even live with a default to true for the automatic check at startup as it probably will make it easier for new GO users to begin with.

src/grandorgue/GOFrame.cpp
```diff
@@ -1055,6 +1071,11 @@ void GOFrame::OnMenuClose(wxCommandEvent &event) {
 }
 
 bool GOFrame::CloseProgram(bool isForce) {
+  // Make sure update checking is completed
+  if (m_UpdateCheckerThread) {
+    m_UpdateCheckerThread->Wait();
```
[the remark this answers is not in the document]
nanoufo: Made update checking interruptable

oleg68 (changes requested):

src/grandorgue/GOFrame.cpp
```diff
@@ -1055,6 +1071,11 @@ void GOFrame::OnMenuClose(wxCommandEvent &event) {
 }
 
 bool GOFrame::CloseProgram(bool isForce) {
+  if (m_UpdateCheckerThread) {
+    // Stop update checking so that we can safely destruct the thread
+    m_UpdateCheckerThread->Stop();
```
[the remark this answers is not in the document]
nanoufo: No, because the same UI thread is used for CloseProgram and for nullification.

build-scripts/for-win64/prepare-debian-based.sh
```diff
@@ -17,7 +17,14 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
...
   wget unzip cmake g++ pkg-config g++-mingw-w64-x86-64 nsis \
   docbook-xsl xsltproc gettext po4a imagemagick zip libz-mingw-w64-dev \
-  wine32 winbind
+  wine32 winbind pipx
+
+# Install tool to download pre-compiled libraries from msys2 repositories
+pipx install git+https://github.com/nanoufo/msys2-downloader.git@main
+
+# Download curl and dependencies
+msys2-download --extract-root ./msys2-sysroot --env mingw64 curl-winssl
```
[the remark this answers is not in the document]
nanoufo: Done! Now, MSYS2 packages are converted to .deb packages and installed. To check if curl-winssl is installed, it checks whether the curl-winssl-msys2-mingw64 package is installed.

src/grandorgue/GOApp.cpp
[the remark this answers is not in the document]
nanoufo: Now, update checking is interrupted on program exit.

oleg68:
@nanoufo Seems now everything works as expected.

The only things to do are
- to add a changelog entry
- to update help image of Settings/Options
- to describe the new settings in the help
- to describe the new menu item in the help

help/grandorgue.xml
```diff
@@ -6261,6 +6270,13 @@ closed with changed metronome settings.
         </note>
       </sect2>
     </sect1>
+    <sect1 id="update-checking">
+      <title>Updates</title>
+      <sect2>
+        <title>Check for updates at startup</title>
+        <para>Check the checkbox here if you would like GrandOrgue to notify you when a new version of GrandOrgue is available.</para>
```
oleg68, line 6277:
```suggestion
        <para>Check the checkbox here if you would like GrandOrgue to notify you when a new version of GrandOrgue is available. If there is one, GrandOrgue will create a new menu <link linkend="update-available">Update available!</link>.</para>
```
Next, create a description of the Update available! Menu menu with all entries after the description of the Help Menu.

oleg68 (changes requested):

build-scripts/for-win64/prepare-debian-based.sh
```diff
@@ -17,11 +17,26 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
 sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
   wget unzip cmake g++ pkg-config g++-mingw-w64-x86-64 nsis \
   docbook-xsl xsltproc gettext po4a imagemagick zip libz-mingw-w64-dev \
-  wine32 winbind
+  wine32 winbind pipx
```
oleg68, line 20:
```suggestion
  wine32 winbind pipx git
```

otherwise ``pipx install git+`` doesn not work

oleg68 (changes requested):

oleg68:
@nanoufo Could you enable a help button (it is supported in GODialog automatically) in the Changelog window? You also have to add a description of this window in the help.

build-scripts/for-win64/prepare-debian-based.sh
nanoufo: pipx is installed in [the latest commit](https://github.com/GrandOrgue/grandorgue/blob/d0ad3014f1da880b057cbb129a682ca266c76746/build-scripts/for-win64/prepare-debian-based.sh).

oleg68: But it cannot install anything from git+...  unless  git is also installed.

nanoufo:
> @nanoufo Could you enable a help button (it is supported in GODialog automatically) in the Changelog window? You also have to add a description of this window in the help.

I don't think that a simple window like this needs a help page. Writing instructions for __every__ UI element seems like a waste of time.

oleg68 (approved):

oleg68:
@nanoufo I'm approving this PR but you should to resolve the conflict prior merging this pr.

oleg68:
@rousseldenis @larspalo could you take a look to the latest version of the updater? May we merge it?

larspalo (approved):

PR #1845, opened 2024-03-17
Switched to the c++17 standard

oleg68:
Last time I have been aught to reimplement some standard library functions that are already present in a newer c++ standards. It would be better to use their standard implementation instead.

So I switched GrandOrgue to the c++17. It is fully supported in GCC 9 shiped with ubuntu 20, where GitHub builds occur.

It is still able to build locally with an older c++ passing `-CMAKE_CXX_STANDARD 11` to cmake unless the c++17 features are actually used in the code.


larspalo (approved):

rousseldenis (approved):
Seems fair

PR #1847, opened 2024-03-19
Build for Windows on Windows using MSYS2

nanoufo:
By simplifying the building process, the need for maintaining extra repositories for cross-compiled dependencies is eliminated.
Currently, cross-building on Linux is not functioning properly (because now Wine is not utilized for running cv2pdb), but I can restore it if necessary.

Brief comments on file changes:
1. A `shell` variable has been introduced to matrix builds. The Windows build employs the MSYS2 MINGW64 shell. Starting the shell is a bit tricky as a login shell needs to be started in order to modify the PATH, and we need to preserve the working directory.
2. Quotes have been added around ${{github.workspace}} to prevent bash from interpreting backslashes in the path as escape characters.
3. VC Tools are expected to be installed (essential for cv2pdb), and they come pre-installed in Github Runners.
4. `cmake_path(APPEND ...)` is employed for concatenating paths in CMake install files to tackle the issue where the naive approach results in an invalid path `/D:/grandorgue` when joining an empty path and `D:/grandorgue`. 
5. Outdated MinGW headers were previously included during the PortAudio build process, resulting in build failure. Now, the build process incorporates headers from the current MinGW installation.
6. `C:\Windows\System32\convert.exe` could be mistaken for ImageMagick's `convert` utility. As a solution, we exclude the `System32` directory when searching for `convert.exe`.

Notes:
1. Inkscape is now installed because, for some reason, on GitHub runners, ImageMagick uses it for converting .svg to raster images, while on my computer it uses librsvg. This leads to longer download times for dependencies. Total build time could be reduced by using caching from the [msys2/setup-msys2](https://github.com/msys2/setup-msys2) action. However, this change would require us to deviate from the current 'prepare-x.sh' and 'build-on-y.sh' scheme.
2. All dependencies are available in MSYS2 repositories, and we use them from there, except for PortAudio and RtAudio, which we build with additional USE flags.


oleg68:
1. Using the msys2 repository instead of lots of cross-compilation github projects is a good approach.
2. Capability of building GrandOrgue on windows is also a good approach.
3. Github builds should be reproducible on a local machine. So the build logic should be in two scripts `prepare-windows.cmd` and `build-on-windows.cmd` instead of the github workflow. Please provide that scripts.
4. Personally for me cross-compilation on linux is a preferred way than building on windows, because I can use a podman container instead of windows wirtual machine. So I think github builds should remain on linux.
5. If you converted `prepare-debian-based.sh` to using msys instead of the own cross-compilation projects (in a separate PR), it would be great!

rousseldenis:
@nanoufo @oleg68 What's the status of this ?

oleg68:
Now we already use msys2 for windows build. It is a very little step left to do: to create cmd files for building on windows (move the commands from the workflow prepare-windows.cmd and build-on-windows.cmd). Because @nanoufo does not want to continue, may be someone else will make this?

oleg68:
Replaced with #2571

PR #1849, opened 2024-03-20
Changed displaying of the right part of paths https://github.com/GrandOrgue/grandorgue/issues/1663

oleg68:
Resolves: #1663 

This is the last PR related to #1663.

This PR 
- changed displaying paths in the Organ Selection Dialog
- added capability of saving column sizes in the Organ Selection Dialog


larspalo:
@oleg68 It seems like the church name will overflow into the package name column if it's empty:

GridOverflow

If the package name column has content, the church name will not overflow:

GridNotOverflowing

I think it would be better if the church name didn't overflow at all.

Also you can see the package paths again being right aligned - which is not so nice, but I guess that it would be fixed with the previous PR that has not yet been merged with this one, right?

oleg68:
@larspalo @rousseldenis could you look at this PR?

larspalo (approved):

Issue #1785, opened 2024-01-27
Default MIDI settings not applied where there is a default .cmb file


ahall41:
Where there is a default combination file, containing (say) combination or tuning settings only, which is then loaded on startup, the default MIDI settings (keyboard, swell, GC, etc), should also be applied.

ahall41:
Some suggestions:
1. As well as importing a default .cmb settings, also import default .yaml
2. If the .cmb file does no contain a midi (or key stroke) setting, apply the default after import (including import settings)
3. OR Apply defaults after importing a default cmb (but not after a saved .cmb)
I think it can be assumed that, by and large, the defaults are what the user wants to use?

oleg68:
> 1. As well as importing a default .cmb settings, also import a default .yaml with the same name as the .organ

There is no any reason of importing a default .yaml because the combinations can be stored in the default .cmb,

PR #1865, opened 2024-04-06
Fixed ignoring initial midi setup when loading an organ with a preset without midi events configured https://github.com/GrandOrgue/grandorgue/issues/1785

oleg68:
Resolves: #1785

Earlier: initial midi settings were not used when loading a preset .cmb file.
Now: if some element doesn't have any midi events configured in the preset, it uses the midi events from the initial midi config.

When the user saves the preset, these midi events are also saved. So if the user changes the initial midi config, it does not affect the events have already been saved to the preset.


larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1804, opened 2024-02-17
Problems saving release tail length


ahall41:
To demonstrate the issue

- Open an organ (e.g. Demo). Set the release tail length to (say) 1000 and save your settings.
- Reload the organ - the tail length is set to 1000
- Set release tail length to max, and save. Reload the organ - tail length loads as Max (so far so good)
- Set the release tail length to (say) 2000. Reload without saving any settings - tail length loads as 2000.

I have several organs. Some have a long reverb which needs curtailing, and some don't. As a result I currently have to set them to 3000 rather than max (those set to max pick up whatever is set for the previous organ),

There is a similar issue in Organ Settings. Set a parent object to 3000, then set one of its children to Max. It seems to set OK, but when you go back it reverts to 3000.

Issue #1368, opened 2023-02-03
Feature request: Store convolution reverb settings in CMB.


kerkovits:
There are some very dry samplesets created for GO. This is good when playing in a large hall, but at home, I have to add convolution reverb to these samples. On the other hand, most samplesets are wet, it is not reasonable to use convolution reverb for that at all. Even the 'dryness' of dry samples are different, so the best gain setting varies with the ODF. At the moment, these settings are saved globally, so I must go to the settings window each time I load a new organ.

Would it be possible to store these settings in the CMB of the organ? This would make possible to use a different reverb setting for each organ.

ahall41:
And isn't it about time this feature can be supported? It seems to work quite well, but you definitely need to change (i.e. remove) it when you load a different organ.

PR #1866, opened 2024-04-06
Fixed saving Max release tail to the organ preset https://github.com/GrandOrgue/grandorgue/issues/1804

oleg68:
Resolves: #1804

Earlier the release tail length was a GrandOrgue config option. If an organ had the `Max` value in it's preset, the GrandOrgue config option was used. It made unabl to set `Max` for particular organs.

Now the release tail length only presents in organ settings, not in the GrandOrgue config. It always reset when another organ is loaded.


larspalo (approved):
It's quite possible that someone will be surprised by this change in behaviour, but I actually do think this is better.

kerkovits:
This is closely related to https://github.com/GrandOrgue/grandorgue/issues/1368. Would it be possible to consider that issue in the future?

oleg68:

rousseldenis (approved):

PR #1880, opened 2024-04-14
Fixed crash on opening a very old config https://github.com/GrandOrgue/grandorgue/discussions/1869

oleg68:
This PR
- fixes crash if the `GrandOrgueConfig` file was made with an old GrandOrgue version
- filles package path of organs registered with an old  GrandOrgue version


oleg68:
@larspalo @rousseldenis could you take a look at this PR? I would include it in 3.14.1 as well.

rousseldenis (approved):

larspalo (approved):

Issue #1762, opened 2023-12-30
Multiple Issues with MIDI setups


pipe-organ-enthusiast:
1. When configuring the pedal board and manuals by using "Detect complex MIDI setup", everything used to work correctly. Now when I do that and try to play, the notes remain depressed. I have to then go back into the midi settings and change the minimum velocity to 1 to fix it. Please fix this bug so that when the "Detect complex MIDI setup" is used, the proper values are given for the necessary parameters.

2. When using "Detect complex MIDI setup" for the stops on ANY sampleset, after toggling them to my Novation Launchpads, the event comes up as "9x Note" when it should come up as "9x Note On Toggle" as it did in previous versions of GrandOrgue.
Please FIX THIS!

3. I raised this as an issue before and it was brushed off. I should be able to specify different NOTE-ON and NOTE-OFF values for my stops so the colours are displayed on my Novation Launchpads so I know what pads contain stops, couplers, combinations, etc. I should also be able to have NOTE-OFF values of 0 programmed into the "ON" button in Master Controls to OVERRIDE the NOTE-OFF values in my stops, couplers, combinations, etc, so that when I switch samplesets or shut down GrandOrgue, that my launchpads shouldn't be lighted. This used to work way back in version 3.11.2-1 and quit working in version 3.12.0-1. When I last raised this issue, I was told that MIDI events are sent at random when G.O. was shutdown. This in fact is incorrect, as prior to 3.12.0-1, my Novation Launchpads would "turn-off" with Note-Off values of 0 in the "ON" button and NON-ZERO  values in the Note-Off for stops, couplers, combinations, etc. PLEASE FIX THIS!

oleg68:
@pipe-organ-enthusiast please enable "Log midi events" in grandorgue and post here the exact midi events that your lounchpad sends in each case you tried with "detect complex midi setup" with your comments what is the case exactly.

pipe-organ-enthusiast:
[...]

PR #1884, opened 2024-04-15
Fixed behavior of "Detect complex MIDI setup" with Note events in different cases https://github.com/GrandOrgue/grandorgue/issues/1762

oleg68:
Resolves: #1762

This is the second (and the last) PR related to #1762

It fixes behavior of "Detect complex MIDI setup"
- for a manual now the lowest value is detected as 1 instead of 0, that was preventing sound to stopp when the kewy was released
- for a button - introduces the new events: ``9x Note Fixed On`` and ``9x Note Fixed Off``, that allow to use different notes for ON and OFF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed behavior of "Detect complex MIDI setup" with Note events in different cases https://github.com/GrandOrgue/grandorgue/issues/1762
 - Fixed the order of sending midi events from an On indicator. Now they are sent after sending all events from other controls https://github.com/GrandOrgue/grandorgue/issues/1762
 # 3.14.1 (2024-04-17)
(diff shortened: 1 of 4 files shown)


larspalo:
Just remember that a valid note off message is also the Note On message with velocity 0. That case needs to be handled too in the note off category.

oleg68:
> Just remember that a valid note off message is also the Note On message with velocity 0. That case needs to be handled too in the note off category.

It is already supported. `9x Note Fixed Off` means not exactly `Note Off`, but any note event (eighter `Note On` or `Note Off`) that switches the button OFF.

larspalo:
> It is already supported. 9x Note Fixed Off means not exactly Note Off, but any note event (eighter Note On or Note Off) that switches the button OFF.

Good!

oleg68:

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1787, opened 2024-01-27
Divisional Coupler still not working


Psi58:
Thanks @oleg68 for #1737 which evidently fixed the issue demonstrated by the test organ provided by @larspalo in #1725 (https://github.com/GrandOrgue/grandorgue/discussions/1690#discussioncomment-7613679).

Unfortunately this fix does not seem to have resolved my real-world case, using 3.13.3-1 on W10. This has been tested using 2 ways of making edits to add the DivCoupler - in a text editor and also using GOODF.

With the Divisional Coupler active, divisionals from the originating manual respond as expected,  but the coupled divisionals do not respond, and an Unhandled Exception dialog box is shown. Clicking ‘ignore’ clears the error dialogue and allows me to continue; clicking ‘retry’ immediately terminates the program. Nothing is logged in the Event Viewer.

Attached is a comparison of the relevant sections of the ODF from (1) Lars’ test organ, (2) the working organ to which I’m trying to add a Divisional Coupler, and (3) the edited version of (2) adding the Divisional Coupler. I have aimed to make minimal changes to the test organ ODF.

If my ODF contains an error, I’d be grateful if someone could explain it. If not, I suppose this needs to be raised as an issue again.
![DivCouplerError](https://github.com/GrandOrgue/grandorgue/assets/130241113/3bf520ae-2b65-435b-91f2-822fe0ed132a)

larspalo:
Too little is seen in your odf information to say anything for sure.

However, I've done a test myself. Using the built in setter divisionals together with the divisional couplers will unfortunately crash GO. Using the old style of odf divisionals seems to work in my tests. I attach an archive with two .organ files that demonstrate it. (The .organ file with the "OldStyle" divisionals is also named like that)

[FurtherTestingDivisionalCoupler.zip](https://github.com/GrandOrgue/grandorgue/files/14074276/FurtherTestingDivisionalCoupler.zip)

Psi58:
> Too little is seen in your odf information to say anything for sure.
> 
> However, I've done a test myself. Using the built in setter divisionals together with the divisional couplers will unfortunately crash GO. Using the old style of odf divisionals seems to work in my tests. I attach an archive with two .organ files that demonstrate it. (The .organ file with the "OldStyle" divisionals is also named like that)
> 
> [FurtherTestingDivisionalCoupler.zip](https://github.com/GrandOrgue/grandorgue/files/14074276/FurtherTestingDivisionalCoupler.zip)

[...]

PR #1889, opened 2024-04-19
Clarified usage of DivisionalIndex

oleg68:
This is the first PR related to #1787

GrandOrgue supports two types of divisionals combinations:
- a manual ("old style") divisional. GODivisionalCombination::m_IsSetter == false
- a setter element ("banked") divisional. GODivisionalCombination::m_IsSetter == true

This PR
- renames GODivisionalCombination::m_DivisionalNumber to m_DivisionalIndex, because if it is a setter element, it is an index including the bank number
- changes it's type to unsigned
- moves it's initialising from Init() and Load() to the constructor. It is essential that it is set together with m_IsSetter
- sets GODivisionalCombination::m_IsSetter to true in all GODivisionalSetter combinations.

### DIFF
--- a/src/grandorgue/combinations/GODivisionalSetter.cpp
+++ b/src/grandorgue/combinations/GODivisionalSetter.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2023 GrandOrgue contributors (see AUTHORS)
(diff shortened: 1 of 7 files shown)


oleg68:
I thought that this approach is wrong. The right way is not to store divisionalindex in the combination, but set in on the button or the setter level.

PR #1902, opened 2024-04-30
Adding a tone balance voicing option

### LINKED ISSUE #1901: Tone balance voicing option
As mentioned in discussion #1596 among others, it's desired to have an ability to change the tone quality (harmonic balance) of pipe samples.

larspalo:
Resolves #1901.

As discussed in #1596 and #1882 there is a desire to have a voicing option that affects the tone balance between bass and treble.

This PR adds such a feature in the organ settings dialog. The user can set a value for tonal balance in the range of -100 to 100 where 0 is a default value and means that no filtering will take place. The set value will be converted to a frequency in the range of 20 to 16000 that will be the corner frequency for either a first order lowpass or a highpass filter. The relative strength of the lower and higher frequencies can thus be affected and as an effect of that the tone character can be changed.


src/grandorgue/model/pipe-config/GOPipeConfig.cpp
```diff
@@ -182,3 +194,11 @@ void GOPipeConfig::SetPitchMember(float cents, float &member) {
     cents = 1800;
   SetSmallMember(cents, member, &GOPipeUpdateCallback::UpdateTuning);
 }
+
+void GOPipeConfig::SetToneBalanceMember(int value, int &member) {
+  if (value < -100)
+    value = -100;
+  if (value > 100)
+    value = 100;
+  SetSmallMember(value, member, &GOPipeUpdateCallback::UpdateToneBalance);
```
oleg68, line 203:
```suggestion
  SetSmallMember(std::clamp(value, -100, 100), member, &GOPipeUpdateCallback::UpdateToneBalance);
```

src/grandorgue/model/pipe-config/GOPipeConfig.h
```diff
@@ -37,6 +37,7 @@ class GOPipeConfig {
   float m_PitchCorrection;
   float m_ManualTuning;
   float m_AutoTuningCorrection;
+  int m_ToneBalanceValue;
```
oleg68, line 40:
```suggestion
  int8_t m_ToneBalanceValue;
```

Snd please move it after m_ReleaseTail

src/grandorgue/model/pipe-config/GOPipeConfig.h
```diff
@@ -80,6 +81,7 @@ class GOPipeConfig {
   }
 
   void SetPitchMember(float cents, float &member);
+  void SetToneBalanceMember(int value, int &member);
```
oleg68, line 84: This function is not necessary because we have only one ToneBalance member: `m_ToneBalanceValue`

src/grandorgue/model/pipe-config/GOPipeConfigNode.cpp
```diff
@@ -74,6 +74,13 @@ uint16_t GOPipeConfigNode::GetEffectiveReleaseTail() const {
   return releaseTail;
 }
 
+int GOPipeConfigNode::GetEffectiveToneBalanceValue() const {
```
oleg68, line 77:
```suggestion
int8_t GOPipeConfigNode::GetEffectiveToneBalanceValue() const {
```

src/grandorgue/model/pipe-config/GOPipeConfig.h
```diff
@@ -128,6 +130,11 @@ class GOPipeConfig {
     SetPitchMember(cents, m_AutoTuningCorrection);
   }
 
+  int GetToneBalanceValue() const { return m_ToneBalanceValue; }
+  void SetToneBalanceValue(int value) {
+    SetToneBalanceMember(value, m_ToneBalanceValue);
+  }
```
oleg68, line 136:
```suggestion
  int8_t GetToneBalanceValue() const { return m_ToneBalanceValue; }
  void SetToneBalanceValue(int8_t value) {
    SetSmallMember(std::clamp(value, -100, 100), m_ToneBalanceValue, &GOPipeUpdateCallback::UpdateToneBalance);
  }
```

src/grandorgue/model/pipe-config/GOPipeConfigNode.h
```diff
@@ -112,6 +112,8 @@ class GOPipeConfigNode : private GOSaveableObject {
 
   uint16_t GetEffectiveReleaseTail() const;
 
+  int GetEffectiveToneBalanceValue() const;
```
oleg68, line 115:
```suggestion
  int8_t GetEffectiveToneBalanceValue() const;
```

oleg68 (changes requested):

src/grandorgue/sound/GOSoundToneBalance.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef GOSOUNDTONEBALANCE_H_
+#define GOSOUNDTONEBALANCE_H_
+
+#include "GOSoundFilter.h"
+#include <cmath>
+
+class GOSoundToneBalance {
+public:
+  void Init(int slider) {
```
oleg68, line 16: This code is quite long and it is not needed to be inline. So move it to the `.cpp` file, please

src/grandorgue/sound/GOSoundToneBalance.h
```diff
@@ -0,0 +1,44 @@
...
+ * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
+ */
+
+#ifndef GOSOUNDTONEBALANCE_H_
+#define GOSOUNDTONEBALANCE_H_
+
+#include "GOSoundFilter.h"
+#include <cmath>
+
+class GOSoundToneBalance {
```
oleg68, line 14:
```suggestion
class GOSoundToneBalanceFilter {
```

src/grandorgue/sound/GOSoundSampler.h
```diff
@@ -22,6 +23,7 @@ struct GOSoundSampler {
   unsigned m_AudioGroupId;
   audio_section_stream stream;
   GOSoundFader fader;
+  GOSoundFilter::FilterState toneState;
```
oleg68, line 26:
```suggestion
  GOSoundFilter::FilterState toneBalanceFilterState;
```

src/grandorgue/sound/GOSoundProvider.h
```diff
@@ -82,6 +85,10 @@ class GOSoundProvider : public GOStatisticCallback {
 
   float GetTuning() const;
   void SetTuning(float cent);
+  int GetToneBalanceValue() const;
+  void SetToneBalanceValue(int value);
+  void SetToneBalanceFilterSamplerate(unsigned samplerate);
+  const GOSoundToneBalance *GetToneBalance() const { return &m_ToneBalance; }
```
oleg68, line 91:
```suggestion
  const GOSoundToneBalance *GetToneBalanceFilter() const { return &m_ToneBalance; }
```

src/grandorgue/sound/GOSoundProvider.h
```diff
@@ -41,6 +42,8 @@ class GOSoundProvider : public GOStatisticCallback {
   float m_MidiPitchFract;
   float m_Gain;
   float m_Tuning;
+  int m_ToneBalanceValue;
+  GOSoundToneBalance m_ToneBalance;
```
oleg68, line 46:
```suggestion
  GOSoundToneBalanceFilter m_ToneBalanceFilter;
```

src/grandorgue/sound/GOSoundFilter.h
```diff
@@ -0,0 +1,65 @@
...
+  // Calculated filter coefficients
+  double m_B0;
+  double m_B1;
+  double m_A1;
+
+public:
+  GOSoundFilter();
+  void Init(GO_FILTER_TYPE type, double frequency, double gain = 0);
+  bool IsToApply() const { return m_isToApply; }
+  void SetIsToApply(bool apply) { m_isToApply = apply; }
```
oleg68, line 47:
```suggestion
  void ClearIsToApply(bool apply) { m_type = TYPE_NONE; }
```

src/grandorgue/sound/GOSoundFilter.h
```diff
@@ -0,0 +1,65 @@
...
+#include <cmath>
+
+enum class GO_FILTER_TYPE { LPF, HPF, LOW_SHELF, HIGH_SHELF };
+
+class GOSoundFilter {
+public:
+  class FilterState {
+  public:
+    FilterState() { Init(nullptr); }
+    void Init(const GOSoundFilter *filter) {
```
oleg68, line 20: Init() is called rarerly than ProcessBuffer(), so the implementation is not needed to be inline and should be moved to .cpp.

src/grandorgue/sound/GOSoundFilter.h
```diff
@@ -0,0 +1,65 @@
...
+      m_filter = filter;
+      for (int i = 0; i < 2; i++)
+        m_state[i] = 0;
+    }
+    bool IsToApply() { return m_filter->IsToApply(); }
+    void ProcessBuffer(unsigned n_blocks, float *buffer);
+
+  private:
+    float m_state[2];
+    const GOSoundFilter *m_filter;
```
oleg68, line 30:
```suggestion
    const GOSoundFilter *p_filter;
```

src/grandorgue/sound/GOSoundFilter.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+  class FilterState {
+  public:
+    FilterState() { Init(nullptr); }
+    void Init(const GOSoundFilter *filter) {
+      m_filter = filter;
+      for (int i = 0; i < 2; i++)
+        m_state[i] = 0;
+    }
+    bool IsToApply() { return m_filter->IsToApply(); }
```
oleg68, line 25:
```suggestion
    bool IsToApply() const { return p_filter && p_filter->IsToApply(); }
```

oleg68 (changes requested):

src/grandorgue/model/pipe-config/GOPipeConfig.h
larspalo: std::clamp is strictly a C++17 feature without any fallback on for instance Ubuntu 20.04. Therefore I think we should keep the manual range checking until the lowest version we build on fully supports the feature.

src/grandorgue/dialogs/GOOrganSettingsDialog.cpp
```diff
@@ -682,6 +708,7 @@ void GOOrganSettingsDialog::Load(bool isForce) {
     float autoTuningCorrection = m_Last->config->GetAutoTuningCorrection();
     unsigned delay = m_Last->config->GetDelay();
     unsigned releaseLength = m_Last->node->GetEffectiveReleaseTail();
+    int toneBalance = m_Last->config->GetToneBalanceValue();
```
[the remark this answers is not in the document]
oleg68: Both variants bring the same GO behavior. The differnce is only visual if the value is not set at the pipe level: in your approach the user can not see the value that is actually used, but in my approach he/she does.

The same overriding logic exists with the `releaseTail` (one line above). Now the uses can see the actual working value. I suggest to make the same with the `toneBalance`. But if you do not want to do it, I won't insist.

>  the behaviour would be more intuitive

It is more intuitive for a programmer who understands how the program works. But it is not intuitive for the user, that would prefer some WYSIWYG approach.

src/grandorgue/model/pipe-config/GOPipeConfig.h
oleg68: Yes, it a C++17 feature but it is present in Ubuntu 20.04. Since #1845 building GO requires c++17. ``std::clamp`` is already used in GrandOrgue, for example, [here](https://github.com/GrandOrgue/grandorgue/blob/71e02e6db8d0811705faa073c37408e501953aff/src/grandorgue/wxcontrols/GOGrid.cpp#L60)

src/grandorgue/dialogs/GOOrganSettingsDialog.cpp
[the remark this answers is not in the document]
larspalo: > But it is not intuitive for a musician, that would prefer some WYSIWYG approach.

I cannot say for sure, I'm primarily a musician that has learned some programming! However, the same method I propose is used for tuning and gain/amplitude level adjustments too. For now I'd stay with that approach.

src/grandorgue/model/pipe-config/GOPipeConfig.h
larspalo: Very strange that I then couldn't get GO to compile with it when I tested! And I did include <algorithm>.

src/grandorgue/sound/GOSoundFilter.h
```diff
@@ -0,0 +1,65 @@
...
+ * License GPL-2.0 or later
+ * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
+ */
+
+#ifndef GOSOUNDFILTER_H_
+#define GOSOUNDFILTER_H_
+
+#include <cmath>
+
+enum class GO_FILTER_TYPE { LPF, HPF, LOW_SHELF, HIGH_SHELF };
```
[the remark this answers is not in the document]
larspalo: What's your opinion on the usage of the more modern, strongly type safe, `enum class` vs a plain `enum`? Mostly I'm concerned about the `enum class` needing the `static_cast<bool>(m_type)` in the evaluation vs the plain `enum` could be evaluated directly.

src/grandorgue/model/pipe-config/GOPipeConfig.h
oleg68: What error message did you receive when you was trying to compile on ubuntu 20.04?

src/grandorgue/sound/GOSoundFilter.h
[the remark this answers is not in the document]
oleg68: enum class allows to specify the basic type, ie. `uint8_t`. So I'd prefer enum class rather than enum.

src/grandorgue/model/pipe-config/GOPipeConfig.h
larspalo: Sorry, can't exactly remember. I think it was something about no matching function to call, so it might actually have been related to the SetSmallMember function more than the clamp itself. I admit that I didn't try moving it out from the function while testing.

src/grandorgue/sound/GOSoundFilter.h
oleg68: Usually it is difficult to remember who is an owner of the pointed object, whether it is necessary to delete it when the owning object is destroyed.

My suggested mnemonic is to use different prefixes: `m_` for owned objects and `p_` for not owned objects. So I asked you to change the name prefix from `m_` to `p_`.

src/grandorgue/sound/GOSoundFilter.h
oleg68: Why do not you want to add the pointer check for not-null?

src/grandorgue/model/pipe-config/GOPipeConfig.h
oleg68: I made using clamp based on your code. Everething is compiling under ubuntu 20.04: https://github.com/oleg68/GrandOrgue-official/actions/runs/9045413075

I made [a PR to your branch](https://github.com/larspalo/grandorgue/pull/1141). Could you take a look?

larspalo: Ok, so the problem was that the two other parameters didn't default to int8_t by the value given alone, but perhaps int instead, and then the function wouldn't work and it needs a cast to int8_t to work explicitly as a template, right?

src/grandorgue/sound/GOSoundFilter.h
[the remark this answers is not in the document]
larspalo: @oleg68 I was thinking mostly about the necessary `static_cast<bool>(m_type)` and if you think that's problematic for performance or not? When a separate bool was used the comparison would be fast as would a plain `enum` that implicitly would be converted so that a 0 value would equal to a bool value of false and any other to true. The `enum class` that I choose to use won't allow that without an explicit conversion. Your clever idea of using the filter none as a zero value for comparison would work without any cast for a plain `enum,` but needs to be explicit with `enum class`.

If you think this is of no consequence, I'm not bothered either.

src/grandorgue/sound/GOSoundFilter.h
larspalo: No real objection to it. I'll change the declaration and all the places where necessary, if it makes you happy.

src/grandorgue/model/pipe-config/GOPipeConfig.h
oleg68: Yes, it is. `std::clamp(int8_t, int, int)` does not exist.

src/grandorgue/dialogs/GOOrganSettingsDialog.cpp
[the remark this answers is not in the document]
oleg68: > the same method I propose is used for tuning and gain/amplitude level adjustments too

But tuning and gain/amplitude level have an additive behavior in opposite to the overriding behavior of tune balance. Displaying zero is ok for an additive thing, but not for an overriding thing.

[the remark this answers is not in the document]
larspalo: > Displaying zero is ok for an additive thing, but not for an overriding thing.

Perhaps that's a sound reasoning. But then we _must_ re-design the behaviour of the value and effective value calculations. If we enforce a writing of the value to children like you suggest - then setting 0 for a child should _not_ fall back again to the parent value but instead really set the 0 value for that child and thus always override any parent value.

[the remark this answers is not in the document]
larspalo: As I see no way to determine if a child intentionally sets its value to 0 wishing to override parent value, basically the whole get effective value reduces to always just getting the current item value. When any item value is set/changed and applied we must always scan recursively for children and if any are found also apply the value to them to have it work like you want together with my requirement. Or do you have any better idea?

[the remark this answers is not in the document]
oleg68: > do you have any better idea?

My idea is to leave this restriction as is. I'd only change the visualisation: I'd show the inherited value instead of `0`. Indeed, it may confuse a user when he/she sets 0 and apply restores the inherited value. But it is better than to display a value set by the user (zero) but not to use it.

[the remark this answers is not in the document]
larspalo: @oleg68 Sorry, I'm a bit confused with your answer. Can you please clarify? Can we use the existing implementation in this PR? I'd accept that version as it's clearly expressed what happens and it's never confusing with the values.

If you instead want to also display parent value for children, it's impossible to do with only the get effective value function without the confusing behaviour of a child with 0 value. That is something I won't accept. In such case the value must instead also be applied recursively to all children at the same time when the value of a parent is applied. This basically makes the get effective value funcion questionable as the current node value always will be used. The GOOrganSettingsDialog::OnEventApply part for ToneBalance will be slightly more complex but it's not so much worse than the logic for the ReleaseTail.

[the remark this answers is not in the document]
oleg68: I suggest only to replace `GetToneBalanceValue()` with `GetEffectiveToneBalanceValue()` and PR as is.

The confusing behavior of 0 value still exist in both variants: `GetEffectiveToneBalanceValue()` and `GetToneBalanceValue()`, so this replacing does not make worse.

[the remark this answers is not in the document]
larspalo: > so this replacing does not make worse.

Yes, it does, in my opinion. But it can be fixed.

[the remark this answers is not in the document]
oleg68: @larspalo I disagree with the `ApplyToneBalanceRecursively()` because it makes the inverse overriding: if a user sets a value on a pipe level and then sets a value on a rank level, the rank-level toneBalance will override the pipe-level toneBalance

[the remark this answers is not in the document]
larspalo: Exactly the same would happen with your previous suggestion.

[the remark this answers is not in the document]
larspalo: But a check could of course be made to see if a value already was set to something else than 0 and if so not write the parent value to it!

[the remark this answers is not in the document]
oleg68: After you have added a check for zero, the behavior became unstable: a user can override the inherited value with zero, as you intended, but if the user changes the parent value, the zero-setting at the pipe level stops working.

I think "never works" is better than "worked and stops working". Do you agree?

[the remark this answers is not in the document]
larspalo: >I think "never works" is better than "worked and suddenly stops working". Do you agree?

No not really. Because if you go back to the pipe level (or any child) and re-set to zero, which will work, it will be enforced and no filtering taking place. Also, if a pipe level already was set the rank value would not change it. So, basically this is as good a compromise as I can think of for your suggestions.

My first version would never write any value from parent at all but use the parent value if no child was set. It lacked the possibility to re-set the pipe to no filtering if parent used a non zero value, but it was always predictable. Your suggestion only made it even worse because the value from the parent would be written and used at child zero value as well as parent value.

Basically, I think that if this is not good enough for you, then we should completely remove the tone balance voicing from any level but the pipe level to never have any issues of this kind.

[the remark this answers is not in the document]
larspalo: Actually, the technically correct solution for having the tone balance at multiple levels would be to have individual filters for each "level" and cascade them... But this would both increase complexity and put more stress on the polyphony. But it would be the correct way to do it!

oleg68:
@larspalo I'd prefer the first variant - displaying 0 at the pipe level among all other your variants. A predictable solution is better than an unpredictable one.

oleg68:
@larspalo Yes, it would be better.

oleg68 (approved):

rousseldenis (approved):

oleg68:
@larspalo When you merge this PR please select a 'Squash and merge' option forhaving only one commit in the master branch.

PR #1903, opened 2024-05-01
Allowed to init GOConfig without loading defaults

rousseldenis:
As default paths loading is relying on wxApp to be loaded, it does not allow to load a GOConfig object to load a controller without loading the wx App. This is useful for tests but also decouple GO from wx.

This avoids those logs: 

image

### DIFF
--- a/src/grandorgue/config/GOConfig.cpp
+++ b/src/grandorgue/config/GOConfig.cpp
@@ -217,31 +217,7 @@ GOConfig::GOConfig(wxString instance)
     CheckForUpdatesAtStartup(
       this, wxT("General"), wxT("CheckForUpdatesAtStartup"), true),
     m_MidiIn(MIDI_IN),
-    m_MidiOut(MIDI_OUT) {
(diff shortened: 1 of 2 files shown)


src/grandorgue/config/GOConfig.h
```diff
@@ -167,6 +167,7 @@ class GOConfig : public GOSettingStore, public GOOrganList {
   GODialogSizeSet m_DialogSizes;
 
   void Load();
+  void LoadDefaults();
```
oleg68, line 170: Please move it to the private part

src/grandorgue/config/GOConfig.cpp
```diff
@@ -286,6 +262,9 @@ void save_ports_config(
 }
 
 void GOConfig::Load() {
+  if (!m_ConfigFileName) {
```
oleg68, line 265: This checking is not clear.

Do you mean ``if (m_ConfigFileName.IsEmpty())``?

rousseldenis:
This one is ready

oleg68 (approved):

larspalo (approved):

Issue #1813, opened 2024-02-23
Enclosure - AmpMinLevel=0


ahall41:
If you set AmpMinLevel=0, and then completely close the enclosure, nothing will sound on re-opening until another note is played. These enclosures are usually used to regulate the sound balance between near and far perspectives, and is only a problem if you want to change the balance in real time using a midi controller. The work round is to code AmMinLevel=1, but at this setting some sound will bleed through.
I don't think this is a high priority, but thought the phenomenon should be recorded (I recently forgot to code 1 and set it to 0 and someone complained about it!)

eturpault:
You will find some exchanges we had on this topic here : https://github.com/GrandOrgue/grandorgue/discussions/1381#discussioncomment-4939401

ahall41:
I'm not convinced as if you set AmpMinLevel to 0, some sound still bleeds through - which may or may not be desirable. Perhaps the spec should be changed, so that 0 generates a warning? And add a 'health warning' to the Help too? If 0 sets the pipe to 'off' then perhaps it could be some ridiculously low value that is inaudible?

oleg68:
I think AmpMinLevel=0 ahould be considered as a valid value and the current GrandOrgue behavior as an incorrect.

larspalo:
> AmpMinLevel=0 ahould be considered as a valid value

It **is** a valid value, **and** it mutes the sound - as it forces an amplitude of zero on any attack sample - which also incidentally cancels any existing attack sample and returns the sampler to the pool because that's how the audio engine fader works.

oleg68:
> that's how the audio engine fader works

I'm going to change the audio engine behavior.

larspalo:
> I'm going to change the audio engine behavior.

Ok, but it will require quite a lot of care.

PR #1910, opened 2024-05-13
Fixed not sounding of pipe after an enclosure has closed to zero and then opened https://github.com/GrandOrgue/grandorgue/issues/1813

oleg68:
Resolves: #1813

For fixing the issue I completelly reworked the `GOSoundFader` class. Earlier it calculated only the total volue that was depending on the enclosure volume. So it was not feasible to differentiate between finishing the decay and setting enclosure volume to 0. 

Now `GOSoundFader` calculates two volumes: target volume and external volume. `GOSoundFader::IsSilent()` takes into account only fade of the target volume, but not the external volume.

### DIFF
--- a/src/grandorgue/sound/GOSoundFader.h
+++ b/src/grandorgue/sound/GOSoundFader.h
@@ -9,10 +9,10 @@
 #define GOSOUNDFADER_H_
 
 /**
- * This class is responsible of smooth changing a volume of samples.
+ * This class is responsible for smoothly changing a volume of samples.
  *
  * There are different types of volumes
- * - Target volume. It is defined by total capacity of the asmple size and may
+ * - Target volume. It is defined by total capacity of the sample size and may
  *   be adjusted in odf and settings.
  * - External volume. It may be changed by an enclosure during playback.
  * - VelocityVolume. It is defined when a key is pressed and may be changed if
@@ -20,13 +20,13 @@
  *   It is considered as part of an
  * - Total volume. It is a production of of all volumes above.
  *
- * There are three staging of playing one sample related to the target volume:
+ * There are three stages of playing one sample related to the target volume:
  * - Increasing: from 0 to m_TargetVolume during m_IncreasingFrames. It is used
  *   only for crossfade between two samples.
  * - Constant playing: the volume is not changed. Usually it is used for the
  *   attack sample (because the increasing is already sampled).
- * - Decreasing: from m_СurrentVolume to 0. It is used for bothe crossfade from
- *   a loop to a release sample and when playing the release sample when
+ * - Decreasing: from m_СurrentVolume to 0. It is used for both the crossfade
+ *   from a loop to a release sample and when playing the release sample when
  *   ReleseTail limitation is used
  *
  * totalVol = targetVolume * externalVolume
@@ -52,7 +52,7 @@ private:
 
 public:
   /**
-   * Setup the fadef for constant volume of for increasing from 0 to
+   * Setup the fader for constant volume or for increasing from 0 to
    * targetVolume
    * @param targetVolume target volume to reach
    * @param velocityVolume the initial velocity volume. It may be changed later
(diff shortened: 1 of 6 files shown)


larspalo:
@oleg68 It seems to be working fine. There are a few un-important spelling mistakes in the comments though. However, more important, since long all the enclosures on the built in (floating and volume) panel would have the default AmpMinLevel of 1 in the Init of /src/grandorgue/model/GOEnclosure.cpp file to take into account the previous behaviour (of killing the sound at an amplitude of 0). Since this PR changes that behaviour it might be appropriate to also change the amp min level of the built in volume controls in the Init to reflect this change?

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
 #ifndef GOSOUNDFADER_H_
 #define GOSOUNDFADER_H_
 
+/**
+ * This class is responsible of smooth changing a volume of samples.
+ *
+ * There are different types of volumes
+ * - Target volume. It is defined by total capacity of the asmple size and may
```
larspalo, line 15: * - Target volume. It is defined by total capacity of the sample size and may

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
...
+ *   be adjusted in odf and settings.
+ * - External volume. It may be changed by an enclosure during playback.
+ * - VelocityVolume. It is defined when a key is pressed and may be changed if
+ *   another coupled key with the same pipe is pressed with another velocity.
+ *   It is considered as part of an
+ * - Total volume. It is a production of of all volumes above.
+ *
+ * There are three staging of playing one sample related to the target volume:
```
larspalo, line 23: * There are three stages of playing one sample related to the target volume:

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
...
+ * - Increasing: from 0 to m_TargetVolume during m_IncreasingFrames. It is used
+ *   only for crossfade between two samples.
+ * - Constant playing: the volume is not changed. Usually it is used for the
+ *   attack sample (because the increasing is already sampled).
+ * - Decreasing: from m_СurrentVolume to 0. It is used for bothe crossfade from
```
larspalo, line 28: * - Decreasing: from m_СurrentVolume to 0. It is used for both the crossfade from

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
...
-    NewAttacking(target_gain, sample_rate * ms / 1000);
-  }
+  // Last volume points are the volumes at the end of previous Process()
+  float m_LastTargetVolumePoint;
+  float m_LastExternalVolumePoint;
 
-  void NewConstant(float gain);
+public:
+  /**
+   * Setup the fadef for constant volume of for increasing from 0 to
```
larspalo, line 55: * Setup the fades for constant volume or for increasing from 0 to

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
...
 
-  inline void StartDecay(unsigned ms, unsigned sample_rate) {
-    StartDecay(sample_rate * ms / 1000);
+  /**
+   * Start decreasing the volume from the current value
+   * (m_LastTargetVolumePoint) to 0
+   * @param nFrames number of frames for full decay
+   */
+  inline void StartDecreasingVolume(unsigned nFrames) {
+    // may be m_TargetVolume has not yet been reached, but the velocity of
```
larspalo, line 72: // maybe m_TargetVolume has not yet been reached, but the velocity of

src/grandorgue/sound/GOSoundFader.h
```diff
@@ -8,40 +8,78 @@
 #ifndef GOSOUNDFADER_H_
 #define GOSOUNDFADER_H_
 
+/**
+ * This class is responsible of smooth changing a volume of samples.
```
larspalo, line 12: * This class is responsible for smooth changing the volume of samples.

larspalo (changes requested):

oleg68:
@larspalo I changed the min level of internal enclosures to 0

larspalo (approved):
Seems to be working as it should.

oleg68:

rousseldenis (approved):

Issue #710, opened 2021-09-17
Polyphase interpolation not working with lossless compression


larspalo:
Currently GO falls back to linear interpolation when lossless compression is enabled.

Line 409 in GOSoundAudioSection.cpp:
`/* TODO: Add support for polyphase compressed decoders. Fallback to
		 * linear interpolation for now. */`

PR #1919, opened 2024-06-06
Reworked the resampler code https://github.com/GrandOrgue/grandorgue/issues/710

oleg68:
This is a next PR related to #710 

Earlier there were a lot of similar code that did resampling with different conditions.

I implementing a single template class GOSoundResample::ScalarProductionResampler that does resampling in the function ResampleBlock. There are two subclasses LinearResampler and PolyphaseResampler that use the same algorithm of scalar production but has different coefficients and "points": the number of continous source samples required for calculating one target sample.

All conditions come as a template parameters to ResampleBlock: the source of the continous samples "FloatingSampleVector" with different implementations. For the performance reasons I don't use virtual functions but I use a compile-time linking with template parameters.

It allowed me to eliminate lot of duplicate code and use incarnations of GOSoundResample::ScalarProductionResampler::ResampleBlock in all cases. Most of them are in GOSoundStream:DecodeBlock and GOSoundStream:getDecodeBlockFunction.

There are several other minor changes:
- Moved the current interpolation algorithm from GOSoundResample to GOSoundEngine
- All coefficients are calculated in the GOSoundResample constructor. This class does not more contain any changing state
- Added lots of comments
- Small simplification of GOSoundStream logic (eliminated "margins" and extra members)

It is only a preparing refactoring. No GO behavior should be chanded.

But now implementing a polyphase-compressed is quite easy: we only need to create a proper FloatingSampleVector implementation and describe the new combinations in `GOSoundStream:getDecodeBlockFunction`

### DIFF
--- a/src/grandorgue/sound/GOSoundAudioSection.cpp
+++ b/src/grandorgue/sound/GOSoundAudioSection.cpp
@@ -27,6 +27,8 @@
 #define M_PI 3.14159265358979323846
 #endif
 
+static constexpr unsigned DEFAULT_END_SEG_LENGTH = MAX_READAHEAD * 2;
+
 GOSoundAudioSection::GOSoundAudioSection(GOMemoryPool &pool)
   : m_data(NULL),
     m_ReleaseAligner(NULL),
@@ -105,18 +107,11 @@ bool GOSoundAudioSection::LoadCache(GOCache &cache) {
     return false;
   for (unsigned i = 0; i < temp; i++) {
     EndSegment s;
-    if (!cache.Read(&s.end_offset, sizeof(s.end_offset)))
-      return false;
+
     if (!cache.Read(
           &s.next_start_segment_index, sizeof(s.next_start_segment_index)))
       return false;
     if (!cache.Read(&s.transition_offset, sizeof(s.transition_offset)))
-      return false;
-    if (!cache.Read(&s.end_loop_length, sizeof(s.end_loop_length)))
-      return false;
-    if (!cache.Read(&s.read_end, sizeof(s.read_end)))
-      return false;
-    if (!cache.Read(&s.end_pos, sizeof(s.end_pos)))
       return false;
     if (!cache.Read(&s.end_size, sizeof(s.end_size)))
       return false;
@@ -185,18 +180,11 @@ bool GOSoundAudioSection::SaveCache(GOCacheWriter &cache) const {
     return false;
   for (unsigned i = 0; i < temp; i++) {
     const EndSegment *s = &m_EndSegments[i];
-    if (!cache.Write(&s->end_offset, sizeof(s->end_offset)))
-      return false;
+
     if (!cache.Write(
           &s->next_start_segment_index, sizeof(s->next_start_segment_index)))
       return false;
     if (!cache.Write(&s->transition_offset, sizeof(s->transition_offset)))
-      return false;
-    if (!cache.Write(&s->end_loop_length, sizeof(s->end_loop_length)))
-      return false;
-    if (!cache.Write(&s->read_end, sizeof(s->read_end)))
-      return false;
-    if (!cache.Write(&s->end_pos, sizeof(s->end_pos)))
       return false;
     if (!cache.Write(&s->end_size, sizeof(s->end_size)))
       return false;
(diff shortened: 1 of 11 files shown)


oleg68:
An implementation with polyphase and compression can be tried at https://github.com/oleg68/GrandOrgue-official/actions/runs/9438819810

larspalo:
> An implementation with polyphase and compression can be tried at https://github.com/oleg68/GrandOrgue-official/actions/runs/9438819810

While testing this I have crashes/hangs as soon as I try to load a sample set and play anything (without lossless compression enabled, but with 24 bit 48 kHz) with the following log messages:

Sun 09 Jun 2024 21:34:24 CEST: 21:34:24: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55605fabfc88
Sun 09 Jun 2024 21:34:24 CEST: 21:34:24: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55605fabfc88
Sun 09 Jun 2024 21:34:24 CEST: 21:34:24: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55605fabfc88

And then GO would stop responding (hang/crash).

oleg68:
@larspalo Seems I managed to fix this issue.

The result build without polyphase on compressed samplesets: https://github.com/oleg68/GrandOrgue-official/actions/runs/9527551377

The build with polyphase on compressed samplesets: https://github.com/oleg68/GrandOrgue-official/actions/runs/9527558158

larspalo:
I've played a little with the latest builds for both this PR and the polyphase/lossless compression branch. At first everything worked fine and I could open some quite large sample sets with lossless compression and change the interpolation methods. But then, for some reason, as I loaded BygdsiljumChurch that is a wavpacked .orgue package the same error as above (or a very similar one) occurred:

```
Sat 15 Jun 2024 22:36:04 CEST: 22:36:04: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55c5a0075638
Sat 15 Jun 2024 22:36:04 CEST: 22:36:04: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55c5a0075638
Sat 15 Jun 2024 22:36:04 CEST: 22:36:04: Warning: GOCondition::WaitOrStop: timeout while GOSoundGroupTask::Finish waited for condition 0x55c5a0075638
```

Then I tried loading another .orgue package (NorrfjardenChurch) and the same thing happened (just a short blower sound, then log warnings and hang). The same with BureaChurch package. I then opened Burea Church as non package version but with wavpacked samples and the same thing happened again. Same story with KalvtraskChurch original, but not with the extended version - both of them have samples in plain wav format though. I don't know if it's related, but a common denonominator for all of the sample sets that crash/hang is that they have blower noise that starts engaged.

Anyway, I could also reproduce it with orgue_esmuc, GreenPositiv_GrandOrgue_1_2 and Kassel-Niederzwehren (and quite possibly many others) so it's not only my sample sets that are affected.

oleg68:
@larspalo When I was playing my composite organ I managed to reproduce this effect with exactly one note of one stop. This sample set is in plain wav format. So I think it is related with certain samples. I'll discover it more.

oleg68:
It happens with very long samples (> 10 seconds). The reason is unsigned int overflowing in https://github.com/GrandOrgue/grandorgue/pull/1919/files#diff-ff956ee130c777545820fc98e15b72eb22c70d0a15262d4277d700ccf3ae1120R98-R104

```
    inline unsigned AvailableTargetSamples(unsigned endIndex) {
      return m_index < endIndex ? ((endIndex - m_index) * UPSAMPLE_FACTOR
                                   + m_FractionIncrement - 1 // for rounding up
                                   - m_fraction)
          / m_FractionIncrement
                                : 0;
    }
```

oleg68:
@larspalo I fixed the infinite loop. The builds:
- Without polyphase compressed: https://github.com/oleg68/GrandOrgue-official/actions/runs/9536304072
- With polyphase compressed: https://github.com/oleg68/GrandOrgue-official/actions/runs/9536310627

oleg68:

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1265, opened 2022-11-18
[feature] Add regex matching with the audio device name


larspalo:
Since some time, at least dating back to commit 3a861221a252c4ec60aa93e1ce2c031c99cc1744 but could be older, the device id of USB audio devices seem to randomly change again at boots. This requires re-setting the audio device to get working sound again since wrong id is stored since last boot. A matching solution like that of the MIDI devices perhaps would be in order?

oleg68:
@larspalo 

1 . Do you have several devices with the same `Physical name regex pattern` in their midi setup?
2. Are they plugged in an external usb hub or in the usb sockets on the motherboard?

larspalo:
@oleg68 1. No, not normally anymore 2. Directly, or rather into a single usb interface by MIDI cables

However, I do need to do some additional careful testing to rule out that this was happening not only because I tend to use many different GO versions, that may not all be in sync with their versions, and that may have confused the settings (naming of the audio devices/ports). Anyway, the issue was with the audio device (also usb) that changed the port/id name, not really with the MIDI (but indeed the audio device also is MIDI capable and listed as an available MIDI device).

larspalo:
@oleg68 Just to be clear, this is mainly an audio issue when it actually happens, I'm still trying to decide if it's really an issue or not...

This is the main issue:
![Skärmbild från 2022-11-22 21-13-08](https://user-images.githubusercontent.com/6173941/203421187-170177a4-5666-48c7-9a02-6b547ea7a97e.png)
That is: audio device is changed and not matching anymore
The new working settings would be:
![Skärmbild från 2022-11-22 21-13-37](https://user-images.githubusercontent.com/6173941/203421367-c05d024f-42fc-474c-80a8-1e5e217234c9.png)
And the MIDI settings would be:
![Skärmbild från 2022-11-22 21-13-54](https://user-images.githubusercontent.com/6173941/203421411-69131fe4-d487-43cc-9737-82b6264237c2.png)
At another reboot I could have a new working audio device port after a similar error as the first screenshot:
![Skärmbild från 2022-11-22 21-42-44](https://user-images.githubusercontent.com/6173941/203421741-b1cc2915-8791-4772-b9d7-0910bac97973.png)
And the new MIDI settings would be:
![Skärmbild från 2022-11-22 21-42-59](https://user-images.githubusercontent.com/6173941/203421885-45586db2-dba1-44b8-b6f5-304d5df01948.png)

[...]

PR #1929, opened 2024-06-24
Extracted class GODeviceNamePattern from GOMidiDeviceConfig

oleg68:
This is the first PR related to #1265

It extracts the matching device name code from `GOMidiDeviceConfig` to a new class `GODeviceNamePattern`.

### DIFF
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -56,6 +56,7 @@ combinations/model/GOGeneralCombination.cpp
 combinations/GODivisionalSetter.cpp
 combinations/GOSetter.cpp
 config/GOConfig.cpp
+config/GODeviceNamePattern.cpp
 config/GOMidiDeviceConfig.cpp
 config/GOMidiDeviceConfigList.cpp
 config/GOPortsConfig.cpp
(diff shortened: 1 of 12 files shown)


larspalo:
@oleg68 While testing this PR I noticed that I get consistent crashes when enabling the Lossless compression (without compression it seems to be ok). My settings usually was Linear interpolation, 24 bit 48 kHz with the lossless compression when the crash happens. Doesn't matter what sample set I use. It's loaded and soon after playing a few notes on it, GO will crash.

oleg68:
@larspalo I couldn't reproduce this problem. Might be it was not related with this PR?

I rebased https://github.com/oleg68/GrandOrgue-official/tree/refactor/devicenamepattern over the reworked code working with lossless compression and Linear interpolation. Could you try the new build from https://github.com/oleg68/GrandOrgue-official/actions/runs/9766254204 ?

If crashes still happen, could you provide me any additional information stacktrace, crush dump and so on?

larspalo (approved):
Now I cannot get any crash to happen.

oleg68:

oleg68:

rousseldenis (approved):

PR #1931, opened 2024-06-29
Fixed not compatibilyty between polyphase interpolation and loseless compression https://github.com/GrandOrgue/grandorgue/issues/710

oleg68:
Resolves: #710

This PR adds an implementation of polyphase interpolation for compressed samples. 
- It adds the `OSoundStream::StreamCacheReadAheadWindow` class for providing a vector of 8 continous samples necessary for polyphase interpolation from a compressed stream. 
- Now GOSoundStream::getDecodeBlockFunction supports a combination of compressed + polyphase
- Removes warning from GOSettingsOptions::TransferDataFromWindow when compressed + polyphase are both selected

It also makes a few of cosmetic changes:
- Moves the MAX_READAHEAD constant from GOSoundDefs.h to GOSoundAudioSection.cpp
- Replaces the getVectorLength() function with the VECTOR_LENGTH constant
- Moves StreamPtrWindow and StreamCacheWindow classes inside the GOSoundStream class
- Removes the check margin code

### DIFF
--- a/src/grandorgue/sound/GOSoundStream.cpp
+++ b/src/grandorgue/sound/GOSoundStream.cpp
@@ -13,16 +13,16 @@
 /* Block reading functions */
 
 template <class SampleT, uint8_t nChannels>
-class StreamPtrWindow
+class GOSoundStream::StreamPtrWindow
   : public GOSoundResample::PtrSampleVector<SampleT, int, nChannels> {
 public:
   inline StreamPtrWindow(GOSoundStream &stream)
     : GOSoundResample::PtrSampleVector<SampleT, int, nChannels>(
-      (SampleT *)stream.GetPtr()) {}
+      (SampleT *)stream.ptr) {}
 };
 
 template <bool format16, uint8_t nChannels>
-class StreamCacheWindow
+class GOSoundStream::StreamCacheWindow
   : public GOSoundResample::FloatingSampleVector<nChannels> {
 private:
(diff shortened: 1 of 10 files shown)


larspalo (approved):

oleg68:

oleg68:

src/grandorgue/sound/GOSoundStream.cpp
```diff
@@ -76,33 +130,65 @@ GOSoundStream::DecodeBlockFunction GOSoundStream::getDecodeBlockFunction(
...
+            2>>;
+      }
+    } else {
+      if (channels == 1) {
+        if (bits_per_sample >= 20)
+          return &GOSoundStream::DecodeBlock<
+            GOSoundResample::LinearResampler,
+            StreamCacheWindow<true, 1>>;
+
+        assert(bits_per_sample >= 12);
```
rousseldenis, line 174: What does assert() in case of failure (I've read a program Abort) ? Maybe something more user-friendly should be good no ?

oleg68: It checks the condition in debug mode and generates a signal to debugger if it is not satisfied.

It does nothing in a normal running mode.

Usually if checks the correctness of the program so it shouldn't fail.

This line is a copy of line 159 that was before this PR.

rousseldenis (approved):

PR #1944, opened 2024-07-15
Updated Swedish translation.

larspalo:
Just an update of the Swedish .po file for translation.

### DIFF
--- a/po/sv.po
+++ b/po/sv.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: GrandOrgue\n"
 "Report-Msgid-Bugs-To: ourorgan-developers@lists.sourceforge.net\n"
-"POT-Creation-Date: 2024-07-15 16:41+0200\n"
-"PO-Revision-Date: 2024-07-15 18:39+0200\n"
+"POT-Creation-Date: 2024-07-29 15:04+0200\n"
+"PO-Revision-Date: 2024-07-29 21:12+0200\n"
 "Last-Translator: Lars Palo <larspalo@yahoo.se>\n"
 "Language-Team: Swedish\n"
 "Language: sv\n"
@@ -122,7 +122,7 @@ msgstr "%d ms"
 
 #, c-format
 msgid "%d samples per buffer, %d Hz\n"
-msgstr "%d samplingar per buffer, %d Hz\n"
+msgstr "%d sampel per buffer, %d Hz\n"
 
 #, c-format
 msgid "%d: %s"
@@ -202,7 +202,7 @@ msgid "&Channel:"
 msgstr "&Kanal:"
 
 msgid "&Check for updates"
-msgstr "Sök efter uppdateringar"
+msgstr "&Sök efter uppdateringar"
 
 msgid "&Clear"
 msgstr "Rensa"
@@ -244,10 +244,10 @@ msgid "&Device:"
 msgstr "&Enhet:"
 
 msgid "&Down"
-msgstr "Ne&d"
+msgstr "&Ned"
 
 msgid "&Download"
-msgstr "Ladda ne&d"
+msgstr "&Ladda ned"
 
 msgid "&Enhancements"
 msgstr "&Förbättringar"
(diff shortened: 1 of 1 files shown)


oleg68 (approved):

oleg68:
@larspalo I'm approving this PR, but unfortunatelly I cannot check the translation.

larspalo:
@jason-barba I've updated the Swedish translation of GrandOrgue that you perhaps could check both here (the actual sv.po file) and in the intermediate build at https://github.com/larspalo/grandorgue/actions/runs/9943350137. If you have any thoughts, ideas or objections, feel free to share them.

jason-barba:
Will do. Might take some days ...

jason-barba:
@larspalo I am now back at home so I can take a look on the Swedish translation.
Q: Do I have to download the build to perform your request?

larspalo:
@jason-barba To view only the changed strings you can just click on the "Files Changed" tab at the top (where you have "Conversation" to the far left). If you want to view the strings in context of a running version of GrandOrgue you'll need to download a suitable build from https://github.com/larspalo/grandorgue/actions/runs/9943350137 and run it.

jason-barba:
What is the purpose/function om the ampersand in för example Ne&d?

larspalo:
> What is the purpose/function om the ampersand in för example Ne&d?

It's a menu mnemonic.

jason-barba:
Ah! So the d in Ned should comply with din Down.

jason-barba:
BTW how do you want my comments?

larspalo:
@jason-barba You could just leave them in a message here.

jason-barba:
@larspalo You will find my comments inline in the attached PDF. 
[Updated Swedish translation. by larspalo · Pull Request #1944 · GrandOrgue_grandorgue.pdf](https://github.com/user-attachments/files/16411150/Updated.Swedish.translation.by.larspalo.Pull.Request.1944.GrandOrgue_grandorgue.pdf)
Hope they will be of some use ;-)

I have not tried out the build, som there might be some missunderstandings from my side.

I find it troublesome to translate the short and distinct English words/sentences to Swedish because 1. we often use the English word in the Swedish context (Swenglish) and 2. a direct translation word for word may not yield a correct meaning in Swedish, sometimes very much on the contrary. I learned a lot when I was co-authoring a Swedish textbook in data communication.

larspalo:
> I find it troublesome to translate the short and distinct English words/sentences to Swedish because 1. we often use the English word in the Swedish context (Swenglish) and 2. a direct translation word for word may not yield a correct meaning in Swedish, sometimes very much on the contrary.

Indeed, that's exactly why I didn't translate a lot of the MIDI messages. Some of the technical terms doesn't make sense to translate as we use the english word as a "borrowed" word anyway.

larspalo:
@jason-barba The word "loop" is one of the terms I'm hesitant to change, both because it's technical term but also an accepted term in Swedish.

larspalo:
> I can agree, but still dispute

Sure. In the case of the word "loop" it's however a Swedish word according to SAOL and since anyone with at least the slightest knowledge of an audio sample will understand what a loop is in that context I think I'll leave all such words as they are.

I also see some of your comments are about presently un-used strings. I won't change any of them right now.

I'll change all "sampling" related words to "sampel" except for resampling - "Omsampling".

jason-barba:
Good! Glad to be able to contribute to the project :-)

larspalo:
@jason-barba Thanks for your feedback. I've done some pretty extensive updating of the Swedish translation trying to be consistent and taking your suggestions into account. I suggest that you please try an intermediate build from https://github.com/larspalo/grandorgue/actions/runs/10150498087 (some of them can actually be tested without any actual installation, just download for instance https://github.com/larspalo/grandorgue/actions/runs/10150498087/artifacts/1752444517 and unpack it somewhere and run the binary in the /bin directory where you've unpacked it) so that you may see the strings (and choice of mnemonics) in context.

jason-barba:
Just a quick look, but looks good. One thing thou, shouldn't "Arkiv/Settings .." be "Arkiv/Inställningar..."?

larspalo:
@jason-barba (File->)Settings is one of the strings that cannot be translated.

oleg68:
@jason-barba @larspalo have you finished or you still continue working?

larspalo:
@oleg68 It's ready.

oleg68:

oleg68:
@rousseldenis Parce que in n'ya plus de travaux ici, je pense que il est miex de fusionner ce PR. Pourriez-vous approve ca? Merci.

rousseldenis (approved):

PR #1948, opened 2024-07-18
Implemented Load/Save in GODeviceNamePattern

oleg68:
This is a next PR retated to #1265

It moves the loading/saving config code from GOMidiDeviceConfigList to GOMidiDeviceConfig and GODeviceNameConfig.

It will be reused for audio devices later.

### DIFF
--- a/src/grandorgue/config/GODeviceNamePattern.cpp
+++ b/src/grandorgue/config/GODeviceNamePattern.cpp
@@ -9,6 +9,9 @@
 
 #include <wx/regex.h>
 
+#include "config/GOConfigReader.h"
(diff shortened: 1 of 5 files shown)


oleg68:
After some testing I noted that this PR makes an incompatible GrandOrgueConfig changes with earlier versions. This is an undesired behavior and should be fixed before merging.

oleg68:
I fixed the configuration incompatibility. So you may check this PR

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1888, opened 2024-04-19
RtAudio submodule using a commit that is orphan (not on any branch)


hfiguiere:
RtAudio submodule using a commit that is orphan (not on any branch)

![image](https://github.com/GrandOrgue/grandorgue/assets/114441/eaeb81e0-5b40-4553-9927-d60fb19dbfea)

This cause the flatpak build to fail.

```
fatal: Fetched in submodule path 'submodules/RtAudio', but it did not contain 6e962c0422b5953c82466ecf07765c471a7bfacb. Direct fetching of that commit failed.
```

hfiguiere:
If this commit is on a fork, then the submodule should point to that fork.

kerkovits:
The patch provided affects only Windows, so for Linux, you should build against the official release.

hfiguiere:
There is no source package, and the github generated one exclude submodules. So it's pulled form the git tag, and that's where the submodule bug is.

kerkovits:
There is no source package?

I successfully build GO for Linux using
http://www.music.mcgill.ca/~gary/rtaudio/release/rtaudio-6.0.1.tar.gz
https://github.com/GrandOrgue/grandorgue/archive/refs/tags/3.14.1-1.tar.gz

hfiguiere:
for GrandOrgue. This is GrandOrgue issue tracker, isn't it?

kerkovits:
Or alternatively, https://github.com/GrandOrgue/grandorgue/releases/download/3.14.1-1/grandorgue-wx32-3.14.1-1.src.rpm contains all sources in the submodule directory.

PR #1962, opened 2024-08-05
Fixed rtaudio url in the submodule description https://github.com/GrandOrgue/grandorgue/issues/1888

oleg68:
Resolves: #1888

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "submodules/RtAudio"]
 	path = submodules/RtAudio
-	url = https://github.com/thestk/rtaudio.git
+	url = https://github.com/GrandOrgue/RtAudio.git
+        branch = bugfix/6.0.1-asio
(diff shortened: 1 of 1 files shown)


larspalo:
@oleg68 Just a question, how does this change affect the existing (locally) checked out code/submodule that comes from another/earlier source?

rousseldenis (approved):
Indeed

larspalo:
> how does this change affect the existing (locally) checked out code/submodule

I guess not at all since it seems that they point to the same url/commit now.

https://github.com/thestk/rtaudio/tree/6e962c0422b5953c82466ecf07765c471a7bfacb

and

6e962c0422b5953c82466ecf07765c471a7bfacb

commit id in https://github.com/GrandOrgue/RtAudio/tree/bugfix/6.0.1-asio which then points back to the above url...

larspalo (approved):

PR #1971, opened 2024-08-08
[IMP] GOOrganController: allows to be loaded without an application

rousseldenis:
This will improve the independence of GOOrganController from wx library:

- Remove the inheritance from GOTimer (that inherits from wxTimer). https://github.com/GrandOrgue/grandorgue/commit/3b4a332b1901d2d3164fdd60ae94e3e6eb86fab5
- Remove the loading of base images if wxApp is not initialized. https://github.com/GrandOrgue/grandorgue/pull/1971/commits/5717432d99d0c83cc6c14a37f0c2a80d3ba67f8b

Based on https://github.com/GrandOrgue/grandorgue/pull/1268 to check tests.

Review the last two commits.

No GO behavior should have been changed.

In tests, previous version was triggering warnings (by wx library):

image

In this version :

image

### DIFF
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -100,7 +100,7 @@ GOOrganController::GOOrganController(
     m_AudioRecorder(NULL),
     m_MidiPlayer(NULL),
     m_MidiRecorder(NULL),
-    m_TimerManager(NULL),
+    m_timer(NULL),
     p_OnStateButton(nullptr),
     m_volume(0),
     m_b_customized(false),
@@ -127,7 +127,7 @@ GOOrganController::GOOrganController(
     m_MainWindowData(this, wxT("MainWindow")) {
   if (isAppInitialized) {
     // Load here objects that needs App (wx) to be loaded
-    m_TimerManager = new GOTimer();
+    m_timer = new GOTimer();
     m_bitmaps = new GOBitmapCache(this);
   }
   GOOrganModel::SetMidiDialogCreator(pMidiDialogCreator);
@@ -145,8 +145,10 @@ GOOrganController::~GOOrganController() {
   m_tremulants.clear();
   m_ranks.clear();
   m_VirtualCouplers.Cleanup();
-  m_TimerManager->Cleanup();
-  m_bitmaps->Cleanup();
+  if (m_timer)
+    delete m_timer;
+  if (m_bitmaps)
+    delete m_bitmaps;
   GOOrganModel::Cleanup();
   GOOrganModel::SetModelModificationListener(nullptr);
   GOOrganModel::SetMidiDialogCreator(nullptr);
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -121,13 +121,14 @@ GOOrganController::GOOrganController(
     m_MidiSamplesetMatch(),
     m_SampleSetId1(0),
     m_SampleSetId2(0),
-    m_bitmaps(this),
+    m_bitmaps(nullptr),
     m_PitchLabel(this),
     m_TemperamentLabel(this),
     m_MainWindowData(this, wxT("MainWindow")) {
   if (isAppInitialized) {
     // Load here objects that needs App (wx) to be loaded
     m_TimerManager = new GOTimer();
+    m_bitmaps = new GOBitmapCache(this);
   }
   GOOrganModel::SetMidiDialogCreator(pMidiDialogCreator);
   GOOrganModel::SetModelModificationListener(this);
@@ -1006,8 +1007,6 @@ wxString GOOrganController::GetCombinationsDir() const {
 GOMemoryPool &GOOrganController::GetMemoryPool() { return m_pool; }
 
 GOConfig &GOOrganController::GetSettings() { return m_config; }
-
-GOBitmapCache &GOOrganController::GetBitmapCache() { return m_bitmaps; }
 
 GOMidi *GOOrganController::GetMidi() { return m_midi; }
 
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -145,6 +145,8 @@ GOOrganController::~GOOrganController() {
   m_tremulants.clear();
   m_ranks.clear();
   m_VirtualCouplers.Cleanup();
+  m_TimerManager->Cleanup();
+  m_bitmaps->Cleanup();
   GOOrganModel::Cleanup();
   GOOrganModel::SetModelModificationListener(nullptr);
   GOOrganModel::SetMidiDialogCreator(nullptr);
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -82,7 +82,9 @@ static const wxString WX_ORGAN = wxT("Organ");
 static const wxString WX_GRANDORGUE_VERSION = wxT("GrandOrgueVersion");
 
 GOOrganController::GOOrganController(
-  GOConfig &config, GOMidiDialogCreator *pMidiDialogCreator)
+  GOConfig &config,
+  GOMidiDialogCreator *pMidiDialogCreator,
+  bool isAppInitialized)
   : GOEventDistributor(this),
     GOOrganModel(config),
     m_config(config),
@@ -98,6 +100,7 @@ GOOrganController::GOOrganController(
     m_AudioRecorder(NULL),
     m_MidiPlayer(NULL),
     m_MidiRecorder(NULL),
+    m_TimerManager(NULL),
     p_OnStateButton(nullptr),
     m_volume(0),
     m_b_customized(false),
@@ -122,6 +125,10 @@ GOOrganController::GOOrganController(
     m_PitchLabel(this),
     m_TemperamentLabel(this),
     m_MainWindowData(this, wxT("MainWindow")) {
+  if (isAppInitialized) {
+    // Load here objects that needs App (wx) to be loaded
+    m_TimerManager = new GOTimer();
+  }
   GOOrganModel::SetMidiDialogCreator(pMidiDialogCreator);
   GOOrganModel::SetModelModificationListener(this);
   m_setter = new GOSetter(this);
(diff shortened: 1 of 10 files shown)


larspalo (changes requested):
You've not isolated only the timer change but also added a myriad of other changes. Please fix this before I'll approve.

rousseldenis:
> You've not isolated only the timer change but also added a myriad of other changes. Please fix this before I'll approve.

@larspalo Read my PR description. You need to review only the last commit.

It's done as is intentionally to show the tests results (in logs) before the change and after.

Don't worry, if the tests PR is not merged before this, I will remove the tests commits.

larspalo:
> Don't worry, if the tests PR is not merged before this, I will remove the tests commits.

Let us be very clear about this. I'm not going to approve of including the tests (as they currently are done) as long as they are automatically run on the Github builds as a condition of the builds or PRs here. It's just not going to happen.

I will approve of having the tests built as a separate app that's possible to run locally when anyone specifically wants it, though. And in that regard I do see some usefulness in a test suite. If you really have problems understanding why, I can explain further, but I think it's obvious enough already...

rousseldenis:
> > Don't worry, if the tests PR is not merged before this, I will remove the tests commits.
> 
> Let us be very clear about this. I'm not going to approve of including the tests (as they currently are done) as long as they are automatically run on the Github builds as a condition of the builds or PRs here. It's just not going to happen.
> 
> I will approve of having the tests built as a separate app that's possible to run locally when anyone specifically wants it, though. And in that regard I do see some usefulness in a test suite. If you really have problems understanding why, I can explain further, but I think it's obvious enough already...

@larspalo I've updated the PR to show you the actual changes I want to merge.

src/grandorgue/GOOrganController.cpp
```diff
@@ -82,7 +82,9 @@ static const wxString WX_ORGAN = wxT("Organ");
 static const wxString WX_GRANDORGUE_VERSION = wxT("GrandOrgueVersion");
 
 GOOrganController::GOOrganController(
-  GOConfig &config, GOMidiDialogCreator *pMidiDialogCreator)
+  GOConfig &config,
+  GOMidiDialogCreator *pMidiDialogCreator,
+  bool IsAppInitialized)
```
larspalo, line 87:
```suggestion
  bool isAppInitialized)
```

src/grandorgue/GOOrganController.h
```diff
@@ -126,7 +126,9 @@ class GOOrganController : public GOEventDistributor,
 
 public:
   GOOrganController(
-    GOConfig &config, GOMidiDialogCreator *pMidiDialogCreator = nullptr);
+    GOConfig &config,
+    GOMidiDialogCreator *pMidiDialogCreator = nullptr,
+    bool IsAppInitialized = false);
```
larspalo, line 131:
```suggestion
    bool isAppInitialized = false);
```

src/grandorgue/GOOrganController.cpp
```diff
@@ -118,10 +121,15 @@ GOOrganController::GOOrganController(
     m_MidiSamplesetMatch(),
     m_SampleSetId1(0),
     m_SampleSetId2(0),
-    m_bitmaps(this),
+    m_bitmaps(nullptr),
     m_PitchLabel(this),
     m_TemperamentLabel(this),
     m_MainWindowData(this, wxT("MainWindow")) {
+  if (IsAppInitialized) {
```
larspalo, line 128:
```suggestion
  if (isAppInitialized) {
```

larspalo (changes requested):
We usually have parameter names starting with lowercase. But beyond that, my practical tests show that this seems to be working fine.

rousseldenis:
IMHO, we can improve too the code by paying attention to not use too much wx library in core models (e.g.: wxstring -> char instead).

larspalo:
> (e.g.: wxstring -> char instead)

The first note paragraph of the [wxString documentation here](https://docs.wxwidgets.org/3.0/classwx_string.html) is:

```
Note
    While the use of wxString is unavoidable in wxWidgets program, you are encouraged to use the standard string classes std::string or std::wstring in your applications and convert them to and from wxString only when interacting with wxWidgets.
```

So in principle, I can absolutely agree with your idea, though not as char*, but as std::string. That's of course a subject for other PRs though. I'll test this PR a little more later but so far it has performed nicely.

rousseldenis:
> > (e.g.: wxstring -> char instead)
> 
> The first note paragraph of the [wxString documentation here](https://docs.wxwidgets.org/3.0/classwx_string.html) is:
> 
> ```
> Note
>     While the use of wxString is unavoidable in wxWidgets program, you are encouraged to use the standard string classes std::string or std::wstring in your applications and convert them to and from wxString only when interacting with wxWidgets.
> ```
> 
> So in principle, I can absolutely agree with your idea, though not as char*, but as std::string. That's of course a subject for other PRs though. I'll test this PR a little more later but so far it has performed nicely.

Yes, you're right. I wanted to say any standard string representation that does not depends on wx. Letting those on classes for dialogs only. And of course I had in mind the little helper code that does the conversion.

src/grandorgue/GOOrganController.cpp
```diff
@@ -1160,3 +1168,8 @@ GOLabelControl *GOOrganController::GetTemperamentLabel() {
 GOMainWindowData *GOOrganController::GetMainWindowData() {
   return &m_MainWindowData;
 }
+
+/**
+ * Return the Timer Manager for Metronome, Midi recorder, ...
+ */
+GOTimer *GOOrganController::GetTimerManager() { return m_TimerManager; }
```
larspalo, line 1175: Usually Oleg wants such small and simple methods to have their implementation in the header file.

src/grandorgue/GOOrganController.cpp
```diff
@@ -1006,7 +1014,7 @@ GOMemoryPool &GOOrganController::GetMemoryPool() { return m_pool; }
 
 GOConfig &GOOrganController::GetSettings() { return m_config; }
 
-GOBitmapCache &GOOrganController::GetBitmapCache() { return m_bitmaps; }
+GOBitmapCache &GOOrganController::GetBitmapCache() { return *m_bitmaps; }
```
larspalo, line 1017: As you anyway change this method, it could perhaps also be moved to the header file as the implementation is just one line?

rousseldenis:
@larspalo @oleg68 This is ready

src/grandorgue/GOOrganController.cpp
```diff
@@ -118,10 +121,15 @@ GOOrganController::GOOrganController(
...
     m_SampleSetId1(0),
     m_SampleSetId2(0),
-    m_bitmaps(this),
+    m_bitmaps(nullptr),
     m_PitchLabel(this),
     m_TemperamentLabel(this),
     m_MainWindowData(this, wxT("MainWindow")) {
+  if (isAppInitialized) {
+    // Load here objects that needs App (wx) to be loaded
+    m_TimerManager = new GOTimer();
```
oleg68, line 130: If you create these objects here you should also delete these objects in ~GOOrganController

larspalo (approved):
I have tested the build for Linux and not found any issues caused by it.

src/grandorgue/GOOrganController.h
```diff
@@ -184,7 +186,7 @@ class GOOrganController : public GOEventDistributor,
   void AddPanel(GOGUIPanel *panel);
   GOMemoryPool &GetMemoryPool();
   GOConfig &GetSettings();
-  GOBitmapCache &GetBitmapCache();
+  GOBitmapCache &GetBitmapCache() { return *m_bitmaps; }
```
oleg68, line 189:
```suggestion
  GOBitmapCache &GetBitmapCache() const { return *m_bitmaps; }
```

src/grandorgue/GOOrganController.h
```diff
@@ -232,6 +234,11 @@ class GOOrganController : public GOEventDistributor,
   GOMidi *GetMidi();
 
   GOGUIMouseState &GetMouseState() { return m_MouseState; }
+
+  /**
+   * Return the Timer Manager for Metronome, Midi recorder, ...
+   */
+  GOTimer *GetTimerManager() { return m_TimerManager; }
```
oleg68, line 241:
```suggestion
  GOTimer *GetTimerManager() const { return m_TimerManager; }
```

src/grandorgue/GOOrganController.h
```diff
@@ -73,6 +72,7 @@ class GOOrganController : public GOEventDistributor,
   GOAudioRecorder *m_AudioRecorder;
   GOMidiPlayer *m_MidiPlayer;
   GOMidiRecorder *m_MidiRecorder;
+  GOTimer *m_TimerManager;
```
oleg68, line 75:
```suggestion
  GOTimer *m_TimerManager;
```

Either rename `m_TimerManager` to `m_timer` or rename `GOTimer` to `GOTimerManager`

oleg68 (changes requested):

src/grandorgue/GOOrganController.cpp
rousseldenis: Of course. Too used to python where it is not necessary :sweat_smile:

src/grandorgue/GOOrganController.cpp
```diff
@@ -137,6 +145,8 @@ GOOrganController::~GOOrganController() {
   m_tremulants.clear();
   m_ranks.clear();
   m_VirtualCouplers.Cleanup();
+  m_timer->Cleanup();
+  m_bitmaps->Cleanup();
```
oleg68, line 149:
```suggestion
  if (m_timer)
    delete m_timer;
  if (m_bitmaps)
    delete m_bitmaps;
```

Cleanup() should be called from the destructors instead of there

src/grandorgue/GOOrganController.h
```diff
@@ -232,6 +234,11 @@ class GOOrganController : public GOEventDistributor,
   GOMidi *GetMidi();
 
   GOGUIMouseState &GetMouseState() { return m_MouseState; }
+
+  /**
+   * Return the Timer Manager for Metronome, Midi recorder, ...
+   */
+  GOTimer *GetTimerManager() const { return m_timer; }
```
oleg68, line 241:
```suggestion
  GOTimer *GetTimer() const { return m_timer; }
```

oleg68 (changes requested):

src/grandorgue/GOOrganController.h
rousseldenis: Great to have test framework soon 😅

src/grandorgue/GOOrganController.cpp
rousseldenis: That was the implementation a little above. That's why I mimicked.

oleg68 (approved):

PR #1978, opened 2024-08-12
[ADD] Icons management

### LINKED ISSUE #1977: [RFC] Better icons
Quite an easy (not so finally) enhancement for a better user experience and making GO interface a little bit more modern.

I've tested a little bit the Material Design icons provider for wxWidgets and the result seems cool:

image

I try to manage a way to benefit from the library in the project without increasing executable size (maybe the only solution is to make static char as in current implementation).

rousseldenis:
This is a naive implementation of how icons can use the excellent wxMaterialDesignArtProvider.

https://github.com/perazz/wxMaterialDesignArtProvider


src/grandorgue/GOFrame.cpp
```diff
@@ -274,16 +276,19 @@ GOFrame::GOFrame(
 
   m_ToolBar = CreateToolBar(wxNO_BORDER | wxTB_HORIZONTAL | wxTB_FLAT);
   m_ToolBar->SetToolBitmapSize(wxSize(16, 16));
+  GOIconManager *m_icon = new GOIconManager();
+  std::string set = "set";
   m_ToolBar->AddTool(
     ID_AUDIO_MEMSET,
     _("&Memory Set\tShift"),
-    GetImage_set(),
+    m_icon->GetIcon(set),
```
oleg68, line 284:
```suggestion
    m_icon->GetIcon("set"),
```

src/grandorgue/GOFrame.cpp
```diff
@@ -274,16 +276,19 @@ GOFrame::GOFrame(
 
   m_ToolBar = CreateToolBar(wxNO_BORDER | wxTB_HORIZONTAL | wxTB_FLAT);
   m_ToolBar->SetToolBitmapSize(wxSize(16, 16));
+  GOIconManager *m_icon = new GOIconManager();
```
oleg68, line 279:
```suggestion
  GOIconManager iconManager;
```

src/grandorgue/GOFrame.cpp
```diff
@@ -274,16 +276,19 @@ GOFrame::GOFrame(
...
+  std::string set = "set";
```
oleg68, line 280: Could you remove this line? As well as the other std::string variables?

src/grandorgue/icons/GOIconManager.h
```diff
@@ -0,0 +1,61 @@
...
+class GOIconManager {
+
+public:
+  GOIconManager();
+  ~GOIconManager();
+
+  /*
+   * This returns the icon in char
+   */
+  wxBitmap GetIcon(std::string icon_name) {
```
oleg68, line 30: Please make this function static

oleg68 (changes requested):

oleg68:
@larspalo Dou you agree with the new icons?

larspalo:
> @larspalo Dou you agree with the new icons?

@oleg68 @rousseldenis Unfortunately, I think that some of the suggested icons might a bit better and some are much worse. I've not yet taken the time to investigate if there are better options within the art library that's used for the ones I really don't like from what @rousseldenis has picked.

Also I note that it seems that the wx 3.0 builds are broken with this PR which is un-acceptable in itself.

rousseldenis:
> > @larspalo Dou you agree with the new icons?
> 
> @oleg68 @rousseldenis Unfortunately, I think that some of the suggested icons might a bit better and some are much worse. I've not yet taken the time to investigate if there are better options within the art library that's used for the ones I really don't like from what @rousseldenis has picked.
> 
> Also I note that it seems that the wx 3.0 builds are broken with this PR which is un-acceptable in itself.

Yes, as said, this is a draft proposal to enhance ux in a coherent way of icons representations.

Indeed, I need to see what is incompatible with wx 3.0 (I think I've read it was). But, the library makes the executable growing to 120 Mo which is quite unacceptable.

So, at the end, I should replace the library import by static chars icons.

Nevertheless, @larspalo, this is a proposal, we can adjust the ones we want at the end. We can also pick another style or another library.

larspalo:
> this is a proposal, we can adjust the ones we want at the end. We can also pick another style or another library.

Sure. I'm not at all against a refreshing of the UI. But, in my opinion, something isn't necessarily better just because it's "new". When things are getting closer to a real suggestion/PR I'll look into it further.

Issue #1986, opened 2024-08-22
Regression: GO 3.15 hangs when trying to play Binauralpipes samples


kerkovits:
# Steps to reproduce

1. Download one of the sample sets from https://binauralpipes.com/instrumente
2. Open GO 3.15 (I use the wx32 build on Linux, if it matters)
3. Play any note (you do not have to switch any stops) either by MIDI-message or simply by clicking on a key of any keyboard / pedal.

# What happens?

Infinite loop, process must be killed. No sound produced. Previous versions of GO work well.

larspalo:
I can confirm this issue. GO crash with the following backtrace:

```
Thread 23 "GrandOrgue" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fffbd7fa700 (LWP 48907)]
0x00005555557d6168 in GOSoundResample::ResamplingPosition::AvailableTargetSamples (this=0x55555655c6f4, endIndex=16420)
    at /home/lars/Sources/github/grandorgue/src/grandorgue/sound/GOSoundResample.h:105
105	               / m_FractionIncrement)
(gdb) bt
#0  0x00005555557d6168 in GOSoundResample::ResamplingPosition::AvailableTargetSamples(unsigned int) (this=0x55555655c6f4, endIndex=16420)
    at /home/lars/Sources/github/grandorgue/src/grandorgue/sound/GOSoundResample.h:105
#1  0x00005555557d8990 in GOSoundStream::ReadBlock(float*, unsigned int)
    (this=0x55555655c6a8, buffer=0x7fffbd7f7610, n_blocks=512)
    at /home/lars/Sources/github/grandorgue/src/grandorgue/sound/GOSoundStream.cpp:353
#2  0x00005555557bf93d in GOSoundEngine::ProcessSampler(float*, GOSoundSampler*, unsigned int, float)
    (this=0x7ffff1cc6300, output_buffer=0x7fffbd7f86e0, sampler=0x55555655c680, n_frames=512, volume=0.177827939)
    at /home/lars/Sources/github/grandorgue/src/grandorgue/sound/GOSoundEngine.cpp:236
#3  0x000055555581f413 in GOSoundGroupTask::ProcessList(GOSoundSamplerList&, bool, float*)
    (this=0x555556d1b2d0, list=..., toDropOld=true, output_buffer=0x7fffbd7f86e0)
[...]

PR #1987, opened 2024-08-23
Fixed crash on releasing a key of any Binauralpipes organ https://github.com/GrandOrgue/grandorgue/issues/1986

oleg68:
Resolves: #1986

The reason was initiaising GOSoundResample::ResamplingPosition::m_FractionIncrement with 0 that caused dividing by zero.

### DIFF
--- a/src/grandorgue/sound/GOSoundStream.cpp
+++ b/src/grandorgue/sound/GOSoundStream.cpp
@@ -319,7 +319,8 @@ void GOSoundStream::InitAlignedStream(
   /* Translate increment in case of differing sample rates */
   resample = existing_stream->resample;
   m_ResamplingPos.Init(
-    pSection->GetSampleRate() / existing_stream->audio_section->GetSampleRate(),
+    (float)pSection->GetSampleRate()
+      / existing_stream->audio_section->GetSampleRate(),
     startIndex,
     &existing_stream->m_ResamplingPos);
   decode_call = getDecodeBlockFunction(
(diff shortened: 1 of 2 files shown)


src/grandorgue/sound/GOSoundStream.cpp
```diff
@@ -319,7 +319,8 @@ void GOSoundStream::InitAlignedStream(
   /* Translate increment in case of differing sample rates */
   resample = existing_stream->resample;
   m_ResamplingPos.Init(
-    pSection->GetSampleRate() / existing_stream->audio_section->GetSampleRate(),
+    (float)pSection->GetSampleRate()
```
rousseldenis, line 322: @oleg68 I'm wondering how a change on left side of division will solve a problem of zero dividing?

rousseldenis: @larspalo thanks for the explanation!

oleg68: @larspalo The root cause was the attack sample of key noise had 48000, but the release sample had 44100. So when GO was switching to the release, existing_stream->audio_section->GetSampleRate() returned 48000 and pSection->GetSampleRate() returned 44100.

Maybe some other samples have 44100 for both attack and releases. It depends on certain samples. So if you doubt in the code correctness, please explain the details with the samples example.

larspalo: > The root cause was the attack sample of key noise had 48000, but the release sample had 44100.

Yes, the silent Stille.wav file used for the attack part of action releases (done in old style like before we introduced the HasIndependentRelease feature) is indeed at 48000 while the actual "release" sample is at 44100.

> So if you doubt in the code correctness, please explain the details with the samples example.

I would like to investigate the whole re-sampling handling further as it's very interesting, but right now I simply don't have time for it. As already said by someone else, the same type of re-sampling that we already have but with the "reverse" intention to add frequency modulation could be used for expanding the tremulant model and adding a wind model.

larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #1907, opened 2024-05-06
Problem with Changing Paths in Settings


orgelspielerkmd:
The Path for Combinations in the Settings is not allowing me to change it.  I change it and click OK and I open up the settings again and it has returned to the default value.  I do not have my installation on the C:/ drive and that is causing problems.  Every other path changes, but Combinations does not.  I have the latest version 3.14.2-1

Regards,
Evrim ("Ay-vreem")

PR #1991, opened 2024-08-27
Fixed saving of Settings->Paths https://github.com/GrandOrgue/grandorgue/issues/1907

oleg68:
Resolves: #1907

The reason why the path settings may not be saved was that the directory did not exist.

This PR adds making the directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
+- Fixed saving of Settings->Paths https://github.com/GrandOrgue/grandorgue/issues/1907
 - Fixed crash on releasing a key of any Binauralpipes organ https://github.com/GrandOrgue/grandorgue/issues/1986
-- Fixed absance of warnings of unused Pipe999ReleaseCrossfadeLength https://github.com/GrandOrgue/grandorgue/issues/1904
+- Fixed absence of warnings of unused Pipe999ReleaseCrossfadeLength https://github.com/GrandOrgue/grandorgue/issues/1904
 - Fixed loading organ to not abort if reading InfoFilename failed and also fixed showing link in properties dialog if InfoFilename exists
 # 3.15.0 (2024-08-06)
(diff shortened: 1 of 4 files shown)


larspalo:
I have nothing per se against the functionality of this PR, but why are you changing the naming pattern for (one of) the global functions like this?

oleg68:
> I have nothing per se against the functionality of this PR, but why are you changing the naming pattern for (one of) the global functions like this?

Current naming conventions:

- Regular member functions: CamelCase starting with a capital letter
- Static member functions: camelCase starting with a small letter
- global or local static functions: lowcase with underscores.

larspalo:
> global or local static functions: lowcase with underscores.

This is a new one. No previous case for it in the code. Normally that style was only used for local variables. But it really doesn't matter at all in any practical way... It just looks strange to me to have a function named in that manner.

larspalo:
Also, such a style is used for naming a few of the files...

oleg68:
> > global or local static functions: lowcase with underscores.
> 
> This is a new one. No previous case for it in the code. Normally that style was only used for local variables. But it really doesn't matter at all in any practical way... It just looks strange to me to have a function named in that manner.

But I use this style in lots of other files. For example, https://github.com/GrandOrgue/grandorgue/blob/d087875cf03f8803049ac9f0e83852cc813a614f/src/grandorgue/config/GOConfig.cpp#L247

oleg68:
> However, as there are other functions in the same file that follow another naming pattern I think it would have been a better approach to just fix the issue without any re-naming in this PR, and in another separate PR change the naming of all the functions in the same file so that they share a coherent naming pattern (at least within that same file).

I've done it.

This PR only fixes the bug, and another PR #1998 renames the global functions in this file.

larspalo (approved):

rousseldenis (approved):

PR #1999, opened 2024-09-03
Fixed override warnings in macOS build

willeke1234:
Partially resolves #1984

Fixes warnings in the headers like
>warning: 'PreparePlayback' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]

`-Wall` is switched off to remove warnings like
>warning: class 'GOSetterState' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]

Edit: fixed typo, it's -Wall instead of -wall


oleg68:
@willeke1234 I'd prefer to correct the code for avoiding warnings instead of disabling the warnings.

`-wall` is sutable only for building external projects as PortAudio, RtAudio etc because we are limited to change their code.

willeke1234:
@oleg68 I tried to discuss this, should I have filed an issue instead?

rousseldenis:
@willeke1234 Maybe exploring each warning should be great as @oleg68 said, this is better to solve warnings than muting them.

I don't have explored the warnings list, but indeed an issue should be great to list them. And at least, doing a commit for each one (or each group of) should be great, that allows us to merge faster (doing few PR's if some warnings are more reluctant to be solved).

oleg68 (approved):

rousseldenis (approved):
LGTM

larspalo (approved):

Issue #1935, opened 2024-07-05
Multiple Issues With Crescendo Override Off


larspalo:
When I turn Crescendo Override Off (aka. Add mode) I get the following results

- Divisionals/Generals (odf or built in) continue to function correctly but their state won't change anymore thus not indicating which combination was recently pressed
- If I in a programmed crescendo would manually add stops and then back to a lower crescendo step the manually added stops are removed

I would expect that, when in Add mode, the on/off states of other combinations still would work when crescendo override is off and that the manually added stops would be respected no matter if I'm going up or down the crescendo. (Currently the manually added stops/couplers/switches etc. would be respected if going up but when going back down passing the number where they were added they would go away)

In the crescendo add mode I'd actually expect only the on/off states of programmed/stored items to be affected at all, but this can of course be achieved in other ways (for instance by setting scope/scoped).

Steps to reproduce first issue:

- Open/Load any sample set
- Program a few divisionals and/or generals and verify that they work correctly
- Turn off Crescendo Override
- Press any previously programmed divisional/general

Steps to reproduce second issue:

- Open/Load any sample set
- Program a Crescendo (with or without Override)
- Turn off Override when finished if it wasn't off to begin with
- Back crescendo to lowest (now any manually added stops work as expected)
- Go a few steps up the crescendo
- Add a few stops manually
- Back crescendo and the newly manually added stops are removed

hnb2907:
Hi folks,

This is behaviour described by @larspalo is also something I've noticed, and have been planning to discuss.

I don't have experience of how a crescendo pedal on a Classical organ would work.
However, on a Theatre organ, the behaviour is slightly different to GO.

1. The crescendo doesn't change the status of actual stop switches on the console.  In GO, this is represented by the stops on the GUI; here on GO they do change state with the crescendo pedal.  A Theatre organ crescendo never changes the status of the actual stop tabs.
2. On a Threatre organ, inside the electrical "action", the status of each stop tab is internally AND'ed with the stops set at any given crescendo pedal step
3. Increasing the crescendo step number is additive.  As the crescendo step number increases, more stops are added to those already added at previous crescendo steps.  Maybe this explains it [http://www.nightbloomingjazzmen.com/NBJ_Organ/Schantz_Expanded_Crescendo.jpg](url)
[...]

PR #2012, opened 2024-09-24
Added an internal OR switch to every drawstop

oleg68:
This is the second PR related to #1935

It adds an internal `OR` switch to every drawstop. The switch has several named entries. The default entry has an wxEmptyString name.

When any of drawstop switches named entry is activated, the drawstop is also activated. When all of the entries are deactivated the drawstop is also deactivated.

Now this capability is not used and it will be used in a future PR for the Add mode of Crescendo.

So no GO behavior should be used now.


larspalo (changes requested):
The action build show that this PR causes at least switch drawstops to not function correctly. When the drawstop gets active no sound is produced, but when it's deactivated again sound output happens even though the drawstop is off! When the drawstop is visibly active no sound is emitted. In other words after first activation/de-activation they are reversed!

A few of the affected sample sets are Bureå Church, Bygdsiljum Church and Norrfjärden Church.

Older sample sets that directly use/display the stop seems to function normally though.

On a side note, I have no problem with introducing this feature to fix the original issue with the crescendo - that is, when everything actually is working! However, I'm not convinced that exposing this internal OR in any way (ODF) is a good idea - which I understood from another discussion that you eventually intend to do.

oleg68:
@larspalo I fixed the behavior of AND switches.

larspalo (approved):
Seems to be working now.

oleg68:

rousseldenis (approved):

PR #2014, opened 2024-09-28
Renamed GOSetter::m_swell to m_CrescendoCtrl

oleg68:
This is a next PR related to the crescendo.

It just renames `m_swell` to `m_CrescendoEnclosure`.

### DIFF
--- a/src/grandorgue/combinations/GOSetter.cpp
+++ b/src/grandorgue/combinations/GOSetter.cpp
@@ -305,7 +305,7 @@ GOSetter::GOSetter(GOOrganController *organController)
     m_CrescendoDisplay(organController),
     m_TransposeDisplay(organController),
     m_NameDisplay(organController),
-    m_swell(*organController) {
+    m_CrescendoEnclosure(*organController) {
   CreateButtons(*m_OrganController);
 
   m_buttons[ID_SETTER_PREV]->SetPreconfigIndex(0);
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):

larspalo:
@oleg68 Do you realize that a crescendo pedal has nothing at all with an enclosure to do?

oleg68:
> @oleg68 Do you realize that a crescendo pedal has nothing at all with an enclosure to do?

In the real life - yes. But GrandOrgue uses the same `GOEnclosure` class for both swell/enclosures and crescendo.

Earlier `GOSetter` had the `m_swell` member for handling crescendo pedal. It was not clear. So I rename it to `m_CrescendoEnclosure`. `Crescendo` - because it is used for handling crescendo, `Enclosure` - because it has the `GOEnclosure *` datatype. If you disagree, I may rename it to `m_CrescendoPedalCtl`.

larspalo:
> I may rename it to m_CrescendoPedalCtl

I like that much better, yes.

The visual (GUI) expression pedal rendering of swell/enclosure (enclosed swell box) and the crescendo (register crescendo) might very well be of the same type even if the crescendo in real life also might be done as a Rollschweller type. The backend functionality and model is vastly different for them though. Perhaps it could be a future area of improvement to more clearly differentiate between them...

oleg68:
> I like that much better, yes.

Renamed to m_CrescendoCtrl

larspalo (approved):

PR #2016, opened 2024-10-05
Added GODrawStop tests

oleg68:
The first unsuccessful attempt of #2012 motivated me to write a test of GODrawStop.

This PR also adds some GODrawStop methods for constructing a switch network in tests.

### DIFF
--- a/src/grandorgue/model/GODrawStop.cpp
+++ b/src/grandorgue/model/GODrawStop.cpp
@@ -39,6 +39,45 @@ void GODrawstop::RegisterControlled(GODrawstop *sw) {
   m_ControlledDrawstops.push_back(sw);
 }
 
+void GODrawstop::UnRegisterControlled(GODrawstop *sw) {
+  auto end = m_ControlledDrawstops.end();
+  auto pos = std::find(m_ControlledDrawstops.begin(), end, sw);
+
+  m_ControlledDrawstops.erase(pos);
+}
+
+void GODrawstop::ClearControllingDrawstops() {
+  for (auto pControlling : m_ControllingDrawstops)
+    pControlling->UnRegisterControlled(this);
+  m_ControllingDrawstops.clear();
+}
+
+void GODrawstop::AddControllingDrawstop(
+  GODrawstop *pDrawStop, unsigned switchN, const wxString &group) {
+  auto end = m_ControllingDrawstops.end();
+
+  if (m_Type == FUNCTION_INPUT)
+    throw wxString::Format(_("Switch %d already assigned to %s"), group);
+  if (std::find(m_ControllingDrawstops.begin(), end, pDrawStop) != end)
+    throw wxString::Format(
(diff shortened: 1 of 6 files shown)


src/tests/testing/model/GOTestDrawStop.cpp
```diff
@@ -0,0 +1,80 @@
...
+  GOAssert(!sw12->IsActive(), "false AND true");
+  GOAssert(sw13->IsActive(), "false OR true");
+  GOAssert(!sw23->IsActive(), "true XOR true");
+
+  sw1->SetButtonState(true);
+  GOAssert(sw12->IsActive(), "true AND true");
+  GOAssert(sw13->IsActive(), "true OR true");
+  GOAssert(!sw23->IsActive(), "true XOR true");
+
+  /*
```
larspalo, line 67: I generally don't like commented out code (which either should be removed completely or made relevant and thus active), but since this is concerning a non-essential test part I won't argue about it more now.

larspalo (approved):

src/tests/testing/model/GOTestDrawStop.cpp
oleg68: I removed comments

src/grandorgue/model/GODrawStop.cpp
```diff
@@ -39,6 +39,45 @@ void GODrawstop::RegisterControlled(GODrawstop *sw) {
   m_ControlledDrawstops.push_back(sw);
 }
 
+void GODrawstop::UnRegisterControlled(GODrawstop *sw) {
```
rousseldenis, line 42: This is not only about tests, correct?

oleg68: It is used for tests

rousseldenis: I usually don't recommend to add code in core to be used exclusively by tests. That's a weird design.

oleg68: But initially the core was designed to be set up only from the ODF file, not from the testing code. So I had to add some set up/clean up code for using in tests.

I think that is a right pattern: to replace the Load() code with calls to certain (new) setting up methods and to use the same setting up methods in the tests.

oleg68:
@rousseldenis could you approve this PR? I'd like to add more tests for the new functionality (Drawstop internal state), but now I cannot do it because it relays on this PR.

src/grandorgue/model/GODrawStop.cpp
rousseldenis: Ok, but I think it's because of wrong implementation of MVC as they should be independent and so instantiated separately.

rousseldenis: @oleg68 adding comments to these should be great!

rousseldenis (approved):

PR #2023, opened 2024-10-10
Removed GODrawStop::IsActive

oleg68:
I mentioned that ``GODrawStop::IsActive()`` duplicated ``GOButtonControl::IsEngaged()``

So this PR removes ``GODrawStop::IsActive()`` and replaces all it's usages with ``GOButtonControl::IsEngaged()``

### DIFF
--- a/src/grandorgue/model/GOCoupler.cpp
+++ b/src/grandorgue/model/GOCoupler.cpp
@@ -64,7 +64,7 @@ void GOCoupler::PreparePlayback() {
     m_Keyshift = m_DestinationKeyshift + src->GetFirstLogicalKeyMIDINoteNumber()
       - dest->GetFirstLogicalKeyMIDINoteNumber();
   }
-  if (m_UnisonOff && IsActive())
+  if (m_UnisonOff && IsEngaged())
     src->SetUnisonOff(true);
 }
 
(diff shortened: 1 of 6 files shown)


rousseldenis:
Mmmh, I would need some clarification as, from my comprehension, both have different meaning.

I saw the introduction of `IsEngaged()`there: https://github.com/GrandOrgue/grandorgue/commit/49375794fd07cfd092780259d7ae763fa833e19f#diff-da3dd564c78c8f5f60155bc77a2176d5ef7b16df922c9bdca6bd5d28cac1e0b4R50

And the use of `IsActive()` : https://github.com/GrandOrgue/grandorgue/commit/ef40b2011459e20b568173283f774b39236bf5ff

It seems Martin did the inverse of you do here.

@larspalo

oleg68:
@rousseldenis Yes, Martin introduced different state flags for GOButtonControl (`Engaged`) and for GODrawStop (Active). Might be earlier they were used separately, but now the only place that changes these flags for GODrawStop is GODrawstop::SetInternalState():

```
      Display(resState);
      // must be before calling m_ControlledDrawstops[i]->Update();
      m_ActiveState = resState;
```

Display() changes Engaged::

```
void GOButtonControl::Display(bool onoff) {
  if (m_Engaged == onoff)
    return;
  m_sender.SetDisplay(onoff);
  m_Engaged = onoff;
  r_OrganModel.SendControlChanged(this);
}
```

So now for drawstops m_ActiveState always equals to m_Engaged , and there are no more reasons to use them both.

A previous PR introduced another internal states GODrawStop::m_InternalStates rather m_ActiveState #2012. I use the new ones in another PR #2021.

larspalo (approved):

rousseldenis (approved):

Issue #2019, opened 2024-10-09
Crash when using -g switch


willeke1234:
If I run GO with the -g switch (very useful in developing GUI) then GO crashes when I press a key and then press some drawstops. Steps to reproduce:
-Start GO with -g
-Load the demo organ
-Click on a key on the lower manual
-Click on Cornett III
-Click on Flute douce 4'

This is not a bug that regular users will encounter.

macOS Sequoia 15.0
GrandOrgue v3.15.1-1

[Crash.txt.zip](https://github.com/user-attachments/files/17304885/Crash.txt.zip)

oleg68:
The root case is GOReferencePipe becuse it's Initialize() has never been called.

PR #2038, opened 2024-11-08
Fixed crash of reference pipes with the "--gui-only" option https://github.com/GrandOrgue/grandorgue/issues/2019

oleg68:
Resolves: #2019

The reason of the crash was that GOReferencePipe internal structures were set up during the ``Initialise()`` call that was skipped with `--gui-only`.

This PR
- Introduces the new class GOReferencingObject that may refer to another objects that might be loaded from the ODF in any order.
- It's subclass should implement the `OnResolvingReferences` method where it should actually finds another objects from the model because they all have already loaded from the ODF.
- GOOrganModel::Load calls ResolveReferences of all GOReferencingObject instances as the last step of ODF loading
- GOReferencePipe is now a subclass of GOReferencingObject. GOReferencePipe::OnResolvingReferences now sets up it's internal structures instead of GOReferencePipe::Initialize().
- GOReferencePipe is not more a subclass of GOCachedObject because it needs to read neither any files nor the cache

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed crash of reference pipes with the "--justgui" option https://github.com/GrandOrgue/grandorgue/issues/2019
 - Fixed crash on attempt of loading an organ if it's files did not exist https://github.com/GrandOrgue/grandorgue/issues/1990
 - Fixed the sequencer "Save file" button not lightening after inserting or deleting a combination https://github.com/GrandOrgue/grandorgue/issues/2024
 - Fixed appearence, sizing and the scrollbar issues with the Stops window https://github.com/GrandOrgue/grandorgue/issues/1961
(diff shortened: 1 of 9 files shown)


larspalo:
@oleg68 I can still reliably crash GO started with -g switch when I:

- open/load any sample set with a tremulant
- activate tremulant
- then de-activate the tremulant to cause the crash

I've confirmed that the steps above work for among others Burea Church (2023), Jeux d'Orgues 2, Norrfjärden, Pitea MHS and Orgue de Pibrac so the issue seems to be related to synthetic tremulants, as for instance Bygdsiljum with a wave tremulant doesn't cause a crash.

oleg68:
@larspalo I fixed crash with a synth tremulant

larspalo (approved):

oleg68:

oleg68:

rousseldenis (approved):

Issue #2001, opened 2024-09-05
Compile Warnings


willeke1234:
The clang compiler on macOS reports the following warnings. I don't want to fix them myself because I lack the knowledge of GrandOrgue, modern C++ and scripts.

8 times:
```
WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"
```

187 times:
```
clang: warning: argument unused during compilation: '-mmmx' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-msse' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-msse2' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-msse3' [-Wunused-command-line-argument]
```

6 times:
```
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/combinations/model/GODivisionalCombination.h:18:1: warning: class 'GOSetterState' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
class GOSetterState;
^
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/combinations/model/GOSetterState.h:11:8: note: previous use is here
struct GOSetterState {
       ^
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/combinations/model/GODivisionalCombination.h:18:1: note: did you mean struct here?
class GOSetterState;
^~~~~
struct
```

2 times:
```
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/combinations/model/GOSetterState.h:11:1: warning: 'GOSetterState' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
struct GOSetterState {
^
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/combinations/GODivisionalSetter.h:28:1: note: did you mean struct here?
class GOSetterState;
^~~~~
struct
```

16 times:
```
/Users/runner/work/grandorgue/grandorgue/src/grandorgue/sound/GOSoundSamplerPool.h:15:1: warning: class 'GOSoundSampler' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
class GOSoundSampler;
^
[...]

PR #2052, opened 2024-11-28
Elimenated clang compilation warnings "argument unused during compilation" https://github.com/GrandOrgue/grandorgue/issues/2001

oleg68:
This is the first PR related to #2001

It moves adding the cpu-specific compilation options to a separate file ``cmake/AddCpuOptions.cmake`` and adds a checking if the target CPU supports them.

No GO behaviour should be changed.


rousseldenis (approved):
Indeed, better

larspalo (approved):

PR #2070, opened 2024-12-13
Reorganised prepare scripts for tests

oleg68:
This PR

- moves installing cpptrace to a separate script ``prepare-cpptrace.bash``
- removes the special script ``prepare-debian-for-tests.sh``
- added the steps necessary for tests to the normal installation scripts ``prepare-debian-based.sh`` and ``prepare-fedora.sh``


larspalo:
I'm not fond of adding this downloading and building of a third party software as a requirement as the tests are (and should always be) optional.

rousseldenis:
> I'm not fond of adding this downloading and building of a third party software as a requirement as the tests are (and should always be) optional.

Yes, at least that should be optional when wanting to run tests. That was the philosophy before this.

oleg68:
@larspalo @rousseldenis Now by the testing software is not installed unless the `tests` parameter is passed

larspalo (approved):

PR #2076, opened 2024-12-27
Prepared to a new release

oleg68:


rousseldenis:
@oleg68 I see you changed minor version. What do you plan to put in it?

rousseldenis:
That comment as I thought about the change on Mac supported version. IMHO, those kind of change should not be included in patch versions. At minimum a minor version, better major version, in order to give a clean signal to user

larspalo (approved):

oleg68:
> @oleg68 I see you changed minor version. What do you plan to put in it?

You can see this plan at https://github.com/GrandOrgue/grandorgue/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.16.0

oleg68:
> That comment as I thought about the change on Mac supported version. IMHO, those kind of change should not be included in patch versions. At minimum a minor version, better major version, in order to give a clean signal to user

Yes, you are right.

Unfortunally, building for old macos versions on github takes lots of effort (mostly due the lack of support from homebrew), so we will break it often in the future.

@willeke1234 may be you know a good way of installing old dependencies? There is a docker/podman container technology for linux that allows to freeze the old images and to use them for building. But is it possible for macos?

oleg68:
@rousseldenis I share your concerns on breaking the macos 12 support in 3.15.4. But do you have any reasons not to make 3.16.0? If no, could you approve this PR?

rousseldenis (approved):

rousseldenis:
> @rousseldenis I share your concerns on breaking the macos 12 support in 3.15.4. But do you have any reasons not to make 3.16.0? If no, could you approve this PR?

No, just wanted to highlight some releasing flow care about major/minor/patch

willeke1234:
>@willeke1234 may be you know a good way of installing old dependencies? There is a docker/podman container technology for linux that allows to freeze the old images and to use them for building. But is it possible for macos?

I have no idea. Better ask @MStraeten or @Benitoite.

MStraeten:
if you’re using homebrew then you can’t build for an older macOS than the build system. The GitHub runners can’t be changed that easily.
If you want to build for an older macOS you might use a local macports environment (there you can set a development target - you should do it also in /opt/local/etc/macports.conf)

these steps I used building an arm64 version:
first install dependencies (since I have a build environment for darktable and lilypond, these are just the additional dependencies - maybe there are further dependencies, so you need to check the output of the cmake command)

```
sudo port install cmake gettext docbook-xsl-nons pkgconfig fftw-3 wxWidgets-3.2 wavpack yaml-cpp imagemagick jack fftw3-single
sudo port select --set wxWidgets wxWidgets-3.2
```

then switch to the build directory wherever you extracted the grandorgue source package
`cd src/grandorgue/build `

```
cmake .. \
-DWAVPACK_LIBRARIES="-L/opt/local/lib -lwavpack" \
-DDOCBOOK_DIR=/opt/local/share/xsl/docbook-xsl-nons \
-DCMAKE_OSX_DEPLOYMENT_TARGET=11.1 \
-DCMAKE_BUILD_TYPE=Release \
-G "Unix Makefiles"

if that step indicates missing dependencies, you can install those via `sudo port install …`
if you’re justbbuilding for your os, you can ditch the line -DCMAKE_OSX_DEPLOYMENT_TARGET=11.1 \
make
```

oleg68:
@MStraeten 

> first install dependencies

This is the hardest step on github runner, because homebrew does not work for old macos versions.

My question was about is there a way to install dependencies from homebrew for macos 13 now, when it works an to save the result to an "image", that we can use in the future, even when 13 is dessuported.

Docker is such technology for linux. But I don't know, is there any analogues for MacOs?

MStraeten:
there aren‘t different dependencies - using a recent XCode16 with recent updates sdk allows to build for macOS back to 10.14. macports is able to support this, homebrew isn‘t.
I don’t know, if macOS can be run properly in a docker container. But it’s possible to run an older macOS in an UTM virtual machine on a recent M1-4 Mac. Using Rosetta it’s possible to build also for x86_64 architecture on a silicon Mac (at least with macports - I’m not using homebrew, so no experience with that)

MStraeten:
was able to build 3.16.0-0 tag properly with deployment target 11.1 for arm64 and 10.14 for x86_64 using macports

PR #2096, opened 2025-01-14
Renamed the file GODrawStop to GODrawstop

oleg68:
Earlier the file name was `GODrawStop`, but the class was `GODrawstop`

This PR makes these names equal.


larspalo:
@oleg68 Since "Drawstop" is one word, wouldn't it be better to rename things the other way around?

oleg68:
@larspalo You are right. So I renamed the file

larspalo (approved):

rousseldenis (approved):

PR #2103, opened 2025-01-21
Extracted a common midi related code to the midi/object classes

oleg68:
Earier GOLabelControl, GOButtonControl, GOEnclosure, GOManual, GORank had their m_sender, m_midi, m_shortcut, m_division members and a code of their initialising/loading/saving/resetting that was similar.

This PR extracted this code to the new class hierarchy

- GOMidiObject
    - GOSendingMidiObject
        - GOReceivingSendingMidiObject
            - GOMidiObjectWithShortcut
            - GOMidiObjectWithDivision

The virtual methods LoadMidiObject/SaveMidiObject incapsulate the loading and saving code.

Also this PR contains some renaming.

This is just refactoring. No GO behavior should be changed.


larspalo:
@oleg68 While testing this PR I get a lot of log warnings like:

`Fri 24 Jan 2025 20:44:26 CET: 20:44:26: Warning: Empty at section 'Setter000Divisional000' entry 'Name'`

I also note that the divisional names are gone (it was 1, 2, 3 etc) from the MIDI Objects list and now only contain blank entries.

The above happen with any sample set no matter if it was saved or not (earlier).

Furthermore, the initial MIDI connections of (at least) manuals and enclosures no longer seem to work as they did before. That is, they don't connect automatically according to the settings for the initial MIDI when opening an organ not saved.

larspalo:
@oleg68 I can confirm that the first part (with the log warnings) is fixed.

However, the initial MIDI is still not working. Saved organs work fine, but non-saved organs doesn't get connected properly. I can go into the File->Settings Initial Midi tab and see that the exact same keyboard (port) is entered just the same as when I use detection for a keyboard.

Here are screenshots from both the initial MIDI tab and a manual connection that confirm that they are the same. The manual connection will work whereas the initial MIDI configuration will not (automatically as they used to do).

InitialMidi

ManualConnection

oleg68:
@larspalo Seems I fixed it in the last commit.

larspalo:
@oleg68 The MIDIInputNumber for manuals seems to be working now as they auto connect fine.

Unfortunately I get a new warning:

`Sun 26 Jan 2025 19:09:05 CET: 19:09:05: Warning: Unused ODF entry 'Enclosure001/MIDIInputNumber'
`

with any sample set that has defined it. And the auto connection to the swell pedal(s) isn't working for initial MIDI.

oleg68:
> with any sample set that has defined it.

@larspalo Do you have an example of it?

larspalo:
> Do you have an example of it?

Bygdsiljum Church, Piteå MHS, Jeux d'Orgues2, Orgue de Pibrac and many more (including some HW sample sets converted with OdfEdit).

oleg68:
Thank you, Lars.

I fixed the issue.

larspalo (approved):
As far as I've been able to test this, it now performs as expected.

oleg68:

oleg68:
@rousseldenis Je ne peux pas fair les autres changes sans ce PR. Pourriez-vous l'approuver?

rousseldenis (approved):

PR #2119, opened 2025-02-15
Renamed ChurchName to OrganName and moved from GOOrganController to GOOrganModel

oleg68:
It is just a small refactoring. No GO behavior should be changed.


larspalo (approved):
Seems to be working as it should in my tests. Some confusion might arise with this naming change, but I can agree that organ name is perhaps a better description as sometimes more than one organ can exist in a single church.

oleg68:

Issue #1199, opened 2022-09-15
(Enhancement] Export and import of MIDI settings in editable format


willeke1234:
It is a lot of work to add in and out MID events to all buttons and other objects. Hence the request to export and import the MIDI settings in an editable, commonly used format.

See also:
MIDI setup change #596
how to edit cmb files #622
Custom initial MIDI events #1081
Organ configuration and respective files #565

willeke1234:
See also:
create a settings file that can be imported into new organs #1951

oleg68:
@willeke1234 I implemented this capability in [my intermediate build](https://github.com/oleg68/GrandOrgue-official/actions/runs/13347998806). It is available as buttons in the MIDI objects dialog. Could you test it?

willeke1234:
There are some issues:
- When I try to export then I get
![Image](https://github.com/user-attachments/assets/e5c92806-6ec4-46dc-a942-7757a1dfbfc5)
If I choose "Use .ini" then the file is named "Friesach.midi.ini". If I choose "Use Both" then the file is named "Friesach.midi.ini.ini.midi.ini".
- When I try to import then I can't select a .ini file. I changed the name to "Friesach.ini", didn't work.
- The .ini file format is not what I would call an editable format. Technically it is editable, but in practice it is not.

oleg68:
> * When I try to export then I get

Seems it is a mac-specific issue. Do you know anything about forbidding `.ini` in your OS?

>   If I choose "Use .ini" then the file is named "Friesach.midi.ini". 

It is a correct filename. The right extension is `.midi.ini` instead of `.ini`.

If I choose "Use Both" then the file is named "Friesach.midi.ini.ini.midi.ini".
> * When I try to import then I can't select a .ini file. I changed the name to "Friesach.ini", didn't work.

Does `Friesach.midi.ini` work?

> * The .ini file format is not what I would call an editable format. Technically it is editable, but in practice it is not.

I think it is acceptable if only non empty MIDI settings are there. The size is not so huge.

willeke1234:
> forbidding .ini in your OS

I don't know why and how I should do this.

>The right extension is .midi.ini instead of .ini.
>Does Friesach.midi.ini work?

I don't think macOS can handle dots in extensions and sees "Friesach.midi" with extension "ini".

If this file format should be edited by users then it needs a specification and a manual.

oleg68:
[...]

PR #2120, opened 2025-02-15
Eliminated writting default value of midi setup to the cmb file https://github.com/GrandOrgue/grandorgue/issues/1199

oleg68:
A next PR relating to #1199

Most of MIDI objects do not have any MIDI events assigned to them. Earlier when a user saved organ settings to a preset (*.cmb) file, the default values of MIDI object properties also were written to the file.

Now the default values are not written to .cmb, so only objects with any midi assignments appear in the saved files.

It changes nothing in loading the preset files, but it will be used for exporting MIDI settings to separate files in the future.


oleg68:

rousseldenis (approved):

PR #2121, opened 2025-02-15
Added support of saving not compressed ini files https://github.com/GrandOrgue/grandorgue/issues/1199

oleg68:
A next PR related to #1199.

Earlier GOConfigFileWriter can save any settings only in compressed format. 

This PR adds the capability of saving as uncompressed text file.

Now this feature is not used for config and ,cmb files, but in the future it will be used for exporting MIDI settings.


larspalo (approved):

oleg68:

rousseldenis (approved):

Issue #782, opened 2021-10-16
Add setting of Enclosure AmpMinimumLevel to the GUI


jason-barba:
The Enclosure minimum amplitude level should be settable from the GUI. This is IMHO a parameter that is more related to the local hardware (like amplifier, speakers etc) and the environment of the VPO than to the sample set.

//Jens

jason-barba:
Reference #769

mutti-wilson:
I agree. This is something I adjust on sample sets to fit the room or just personal taste. Since it is only a volume adjustment and not changing EQ or other parameters, modifying it has no effect on the overall sound quality.

mutti-wilson:
Check out this that could possibly change this all up.

https://github.com/GrandOrgue/grandorgue/issues/717

oleg68:
There are two kinds of organ settings GUI now:

- Organ settings dialog. It mostly designed for changing pipe properties in the hierarchy (pipe - rank - windchest - organ)
- Midi settings. It is opened with a right click and is designed for settings MIDI properties.

Thinking on implementing setting an enclosure minimum level and another properties in the future, we may add it either to the Organ Settings dialog or to the MIDI settings dialog.

I would prefer to add it to the MIDI settings dialog to an additional tab. My reasons:
- MIDI settings dialog already has tabs, so adding a new tab would not change the GUI.
- There is no any enclosure hierarchy to add to the Organ Settings. Now enclosures are not shown in this dialog.
- Right-clicking is more intuitive for the user than opening a separate dialog

But there are also some contras:
- Minimum amplitude is not related to MIDI directly. So we should rename the dialog from "MIDI settings" to "Settings"
- There won't be a visual relationship between the enclosure and the windchest.

@larspalo @rousseldenis what are your opinions on where we should add this setting?

oleg68:
> it's an ODF specific operation that the user shouldn't be allowed to change - it's the responsibility for the sample set designer (organ builder...).

What is the difference between AmpMinLevel and Ampllitude/Gain, PitchTuning of pipes? Why the second should be customisable and the first should not.

larspalo:
> Why the second should be customisable and the first should not?

Because for "mistakes" done on the pipes by the sample set producer there's no real alternative! For enclosures there's already built in versions supplied by GO itself to use instead as alternatives. It's the GO built in versions that should be possible to customize the desired low limit of for the user.

[...]

PR #2123, opened 2025-02-17
Added capability of customising minimum amplitude level of enclosures added internally by GrandOrgue https://github.com/GrandOrgue/grandorgue/issues/782

oleg68:
#Resolves: #782

This PR adds an Enclosures tab to the Organ Settings dialog.

It shows the Minimal amplitude level of all enclosures and allows the user to change them for all internal enclosures.


larspalo:
@oleg68 On Ubuntu 20.04 the build of this PR will crash immediately when I try to open the Audio/Midi->Organ Settings dialog. It doesn't matter what sample set I try to use or what settings I might have changed, GO reliably crashes as soon as I open the Organ Settings dialog.

oleg68:
@larspalo I fixed the crash

oleg68:

larspalo:
@oleg68 I'm just back from a vacation trip. I'll start reading the PRs.

oleg68:
Excelent!

larspalo (commented):
Personally I'd like to have the "Master" enclosure displayed only at one single position in the tree (at the [Organ] level equivalent root). I wonder if it also would be better to simplify the enclosure tree so that selecting a ("windchest") node would immediately activate/fill the Minimum Amplitude Level for that built in enclosure and all ODF defined ones could just never be displayed... It would be a much cleaner version and if you think the tab name Enclosures would be confusing then it could be re-named as Built-in Enclosures or Panel Enclosures.

At the very least I think it's quite redundant to have first the windchest as a node and below it the (built-in) enclosure with exactly the same name.

So, if you don't want to re-work this completely like I suggest and still want the ODF defined enclosures displayed, I still suggest that selecting the "windchest node" of the tree would automatically set/be the built-in enclosure.

Another thing is that while it's a very nice touch to show the affected windchests for any enclosure I think it would be good if they were not possible to select since it's not really relevant for anything anyway and might actually lead to misunderstandings about what the minimum amplitude level below actually will set (the item in the tree and not the one in the listbox). Since wxListBox doesn't support an non-selectable style by default other tricks would be needed like disabling the control, handling/removing selection, or perhaps more easily by switching to using another control completely.

oleg68:
@larspalo Thank you for your review.

I like that the new Enclosures tab shows the full structure of windchests and enclosures so the user can see the properties of both odf and built-it ones ant can choose what enclosure to use.

I'll move Master to the organ level and I'll try to disable selecting the windchests.

oleg68:
@larspalo 

I implemented displaying `Master` at the organ level and I disabled selecting affected windchests.

Talking about auto selecting a default inner enclosure when the windchest is selected, I don't want to do it, because it would be very unexpected from the UI point of view: now the user has a full control of selecting and deselecting.

I also fixed your small requests. So I'd like to ask for your approval now.

larspalo:
> Talking about auto selecting a default inner enclosure when the windchest is selected, I don't want to do it, because it would be very unexpected from the UI point of view: now the user has a full control of selecting and deselecting.

I was hinting more at the option to just use the windchest tree item as the same built-in GO version of that enclosure, thus getting rid of the (child) duplication. I'm ok with having the ODF versions displayed under them, but for ease of use and a neater look I'd prefer using the windchest (parent) also as the built-in enclosure. This would also make adjustments easier/quicker as one wouldn't need to expand the tree items to get to the (built-in) enclosure.

Otherwise, I think that this PR is now much better than before. I'd like to await the thoughts of @rousseldenis too though on the windchest/enclosure hierarchy and the obvious duplication of items/children that this causes. If the current layout is fine for him I'll also approve since the feature as such, in my opinion, is worth much more than the implementation details.

oleg68:
> I was hinting more at the option to just use the windchest tree item as the same built-in GO version of that enclosure, thus getting rid of the (child) duplication.

I object this idea because it would bring a value to the windchest level but it would lack it's inheritance to the child enclosures.

I've tried to implement your another idea with the list of enclosures instead the windchest-enclosure tree https://github.com/oleg68/GrandOrgue-official/actions/runs/13819598079 , but I don't like it because it hides the windchest-enclosure hierarchy from the user. 

Another issue with that implementation is wrong behavior on changing selection when there are some unapplied changes. Unlike wxTree, wxListBox does not allow to veto selection changes. It can be resolved with restoring the wxTree again with using only the first level items.

larspalo:
>  windchest level but it would lack it's inheritance to the child enclosures

Dear God! There is no such thing and shouldn't be! There should never ever be an inheritance from windchest to enclosures! In real life enclosures can span multiple windchests or just parts of windchests.

oleg68:
> There should never ever be an inheritance from windchest to enclosures!

Now there are no any settings at the windchest level so there is nothing to inherit. 

But you suggest to use the windchest node not as a windchest but as a default enclosure, so the user may expect that changing the default enclosure settings would affect all child nodes, that are other enclosures.

There is an inheritance on the Pipes tab, and users can expect that if two guis look the same, their behavior also would be the same.

oleg68:
@larspalo Could you take a look at https://github.com/oleg68/GrandOrgue-official/actions/runs/13841023334

There is an alternative GUI implementation of enclosures. It is not ideal for me but it might be an acceptable compromise.

larspalo:
> But you suggest to use the windchest node not as a windchest but as a default enclosure, so the user may expect that changing the default enclosure settings would affect all child nodes, that are other enclosures.

> There is an inheritance on the Pipes tab, and users can expect that if two guis look the same, their behavior also would be the same.

This is a valid argument in my opinion, I can agree with that.

I've had a look at the other build with the list. It's much clearer in my opinion and I like it a lot! The only troublesome aspect of it is that the Master enclosure actually does "enclose" all other enclosures.

Thus the list as it currently is done is not an entirely "accurate" display of the hierarchy of the enclosures. Perhaps just moving the Master to the top and then adding all the ODF defined enclosures before the built-in versions would be enough to "clarify" the relationship.

I do like that it's now very quick and easy to select each enclosure directly (as the tab indicates) to see and possibly set the MinimumAmplitudeLevel and also find out exactly what windchests are affected by each enclosure. The exact relationship between individual enclosures could potentially be much more complex but I think that this version might indeed be a good compromise as you said!

larspalo:
@oleg68 Having tested this a little more I think that it perhaps would be better in the list version to have Master and all the built-in enclosures at top and below them list the ODF defined versions instead of having it the other way around.

oleg68:
> it perhaps would be better in the list version to have Master and all the built-in enclosures at top and below them list the ODF defined

I pushed the list version here and I implemented your suggestion. Please check here: https://github.com/oleg68/GrandOrgue-official/actions/runs/13872602989

larspalo:
> I pushed the list version here and I implemented your suggestion.

It's still implemented with a wxTreeCtrl though, but that's of course fine for me. The help image needs to be updated and the help text slightly adjusted to the new grouping of enclosures.

Otherwise I'm very happy with this version. It's easy to use and very clear without any distractions. The initial size of the tree is a bit off for most of my organs but that's easily changed.

oleg68:
> It's still implemented with a wxTreeCtrl though

I use wxTreeCtrl because wxListBox does not support to forbid change selection when there are some unapplied changes.

oleg68:
> The help image needs to be updated and the help text slightly adjusted to the new grouping of enclosures.

Done.

help/grandorgue.xml
```diff
@@ -5652,6 +5652,52 @@ role="bold">Group</emphasis> attribute are grouped in the same submenu.
...
+	<title>Organ Settings dialog, Enclosures tab</title>
+	<mediaobject>
+	  <imageobject>
+	    <imagedata scalefit="1" width="100%" fileref="images/organ_settings_enclosures.png"/>
+	  </imageobject>
+	</mediaobject>
+      </figure>
+      <para>
+	In its left box, this dialog displays all enclosures: both added 
+	internally by GrandOrgueand ones found in the definition file. Now only
```
larspalo, line 5668:
```suggestion
	internally by GrandOrgue and the ones found in the definition file. Only
```

help/grandorgue.xml
```diff
@@ -5652,6 +5652,52 @@ role="bold">Group</emphasis> attribute are grouped in the same submenu.
...
+	It is possible to select multiple elements at once, to e.g. update one
+	setting of multiple elements to the same value in a single operation.
+      </simpara>
+      <sect2>
+	<title>Affected windchests</title>
+	<para>
+	  The list of windchest the enclosure affects.
+	</para>
+	<para>
+	  The same enclosure appears under all the the windchests in tree.
```
larspalo, line 5686: I don't exactly understand what this line means right now though...

larspalo (approved):
I just suggest a few more help text changes, otherwise I'm happy with this.

help/grandorgue.xml

help/grandorgue.xml
oleg68: Removed

oleg68:

rousseldenis (approved):

Issue #2062, opened 2024-11-30
Send a midi message from a lablel just after the MIDI send event is changed


jefferycrowley:
Hello all,

I have recently been attempting to port Sysex messages to my 3 LCD Screens on my Virtual Organ console.   I have only been partially successful.   

Last night I finally got the [sequencer position] from the MIDI Objects menu to display on what was supposed to be one screen but it ended up on all three screens.   I also attempted to get the Organ Name and Temperament label to appear on the first (1) LCD Screen and also get the current file name and  General Bank on two separate line on the second (2) LCD Screen.   I configured the [sequencer position] to be on the second line of 16 characters of the third screen (3).

The sequencer position label appears now on all three screens (this was after attempting to set-up all three screens) albeit the sequencer position is showing correctly on all three screens, but the organ and temperament, current file and general bank are not showing at all on screens 1 & 2.

I  have set up the Device ID = 1 on both Organ Name = Device:  Any Device>Event:  = SYSEX Hauptwerk 16 Byte String Name>Start-Position:  = 1>  Length:  = 16.   Then the Temperament as Device:  =  Any Device>Event:  = SYSEX Hauptwerk 16 byte Sting Name>Start-Position:  = 17> Length:  = 16.

Nothing appears on the screen except the sequencer position - Do I have to restart the Organ in Grand Orgue to make this work?  

For the second LCD Screen Device ID = 2 for both Current File Name=Device:  = Any Device>Event:  = SYSEX Hauptwerk 16 Byte String Name> Start Position:  =1>Length:  =16.   For General Bank Label= Device:  Any Device>Event:  =SYSEX Hauptwerk 16 Byte String Name> Start Position::  =17> Length:  =16.

For the third Screen I have set [sequencer position] as Device ID:  = 3.  Device:  = Any Device>Event:  = SYSEX Hauptwerk 16 Byte String Name> Start Position:  = 23> Length:  3.   Because I want the Sequencer Position to appear in the center of the second group of 16 available spaces.    

I would really appreciate if you would consider to make the  General/Divisional last pressed AND the Button Pressed to appear as a Label item and not a fixed stop-like image so that we can then implement them as an on-line graphic for the sample-set and that is what I would like to program to appear on LCD Screen #3 - first 16 characters.

[...]

PR #2128, opened 2025-02-22
Added resending midi events when exiting from MidiEventDialog https://github.com/GrandOrgue/grandorgue/issues/2062

oleg68:
Resolves: #2062

This PR adds sending midi events on exiting from the Midi Events Dialog with the `OK` button.


larspalo:
@oleg68 Seems like the MIDI event(s) for active stops/switches are not sent anymore when the organ is loaded (and it was programmed/customized/saved before). De-activating the stop/switch/button and then activating it again will correctly send events again. It's just when first loading the organ it won't be sent like before when some element(s) might already be activated (which it correctly still is on screen in GO).

oleg68:
@larspalo Thank you for your review. I fixed the issue with sending MIDI when the organ is loaded.

larspalo (approved):
Seems to work now as it should. Both initial MIDI and saved states show up correctly synchronized between GO and playing console. And as advertised, setting up a send event for an already active element will now also correctly send the event when closing the dialog.

oleg68:

rousseldenis (approved):

PR #2130, opened 2025-02-27
Simplified the code of GOButtonCreator

oleg68:
This PR makes the code of GOButtonCreator a little bit simpler.

Earlier there was a virtual method GetButtonDefinitionList() that returned button definitions. GOButtonCreator::CreateButtons was calling GetButtonDefinitionList().

Because CreateButtons() was called only from GOButtonCreator subclasses where the button definitions were known, this PR makes passing the button definitions to CreateButtons() directly. It makes GetButtonDefinitionList() unnecessary.

It also renames the `m_element_types` members in the subclasses to BUTTON_DEFS and makes the static variables in their cpp files.

It is only refactoring. No GO behavior should be changed.


oleg68:

rousseldenis (approved):

Issue #1974, opened 2024-08-09
Expanding the file/settings/initial midi tab


davidgritter:
The initial midi tab in grandorgue settings presently allows one to define midi receive event for a limited set of organ elements in a few groups ( manuals, enclosures, sequencer, master controls,  metronome.)  I believe the items in the sequencer group refer to buttons in th combination setter panel.

With increasing use of small displays, custom organ consoles, and headless operation, it would be very useful to expand the number of organ elements that can have an initial midi configuration defined here.  Thus I would suggest that having settings for all the standard panels (those not defined for a specific organ, but present for all organs) would be appreciated.  Specifically, the following additional panels: 

- Crescendo Pedal  (including for the pedal itself)
- Divisionals
- Generals
- Combination setter (partially implemented already)
- Master controls (partially implemented already)
- Recorder/Player
- The coupler panels would be nice but not necessary, and may be difficult since these panels seem to change, depending on the organ loaded

Suggested changes:

- For all elements include the abiltiy to define send events as well as receive events. For example, the Save button iluminates when configuration changes have been made, but not yet saved. Having a send event would allow a button on a remote display to also become illuminated in this case.  Similarly when the combination SET button is active, a remote display or LED might want to know.
- Include all active labels, so that they can send hauptwerk sysex messages to remote displays, allowing display of the loaded organ, current transpose, temperament, memory levels, crescendo level, etc.
- Consider making the initial midi tab list hierarchical  by panel, similar to the audio/midi/organ settings panel, due to the larger number of elements that will be in the list.
- Consider reworking the Master controls panel  to make temperament selection easier.  From the audio/midi/temperament drop down menu in grandorgue, temperament selection is hierarchical.  This could be implemented on the master controls panel by adding category up/category down buttons and a new active label to show the current category. ( this, suggestion, of course, has little to do with the initial midi tab, but with the large number of temperaments, selecting one headlessly is very difficult. Perhaps it might be better to implement some means of assigning midi messages to each available temperament?)

willeke1234:
Maybe it would be more useful to implement #1199 Export and import of MIDI settings in editable format.

bonninr:
I'd like to add a petition for allowing the current loading organ elapsed/remaining time as sysexes to screend.

oleg68:
@larspalo please describe your ideas of improvements of the Initial MIDI settings

oleg68:
@larspalo I'm going to implement your idea in another way.

[...]

Issue #1740, opened 2023-12-06
Feature Request: Add panel to MIDI Objects dialogue


ahall41:
When using the MIDI Objects dialogue to set a control, there are numerous items with the same name (e.g. `Button`). Is it possible to add the panel, or some other identifier, to ascertain exactly what each Button (etc.) is doing?

oleg68:
@ahall41 This dialog contains a table with two columns: Type and Element.

`Button` is the control type. The actual name of the button is in the second column. So I don't see what can I enchance with this panel.

ahall41:
For divisionals you can have several sets of 1-10, (as well as 1-10 in the generals). And +1, +10, -1, -10 for Master Controls/Metronome etc. In the .cmb file they all have unique identifiers - maybe this could be propagated? Its only a suggestion, after someone commented in a post that they couldn't find a particular control.

PR #2133, opened 2025-03-01
Simplified passing MIDIInputNumber

oleg68:
This is a next MIDI-related PR.

Earlier there was a member m_MIDIInputNumber in GOEnclosure and GOManual and a very complex and unstable call graph for obtaining it in GOMidiReceiver::Load with additional virtual functions.

Now GOMidiReceivingSendingObject has m_MidiInputNumber and passes it to GOMidiReceiver::Load without extra virtual functions.

As a side-effect, now it is possible to specify MIDIInputNumber for Stop, and Switch objects in the ODF as well for Enclosure and Manual.


oleg68:

larspalo:
> As a side-effect, now it is possible to specify MIDIInputNumber for Stop, and Switch objects in the ODF

Seems to not complain about specifying MIDIInputNumber for Tremulants and Couplers either.

What's the intended use case for setting MIDIInputNumber property for Stop/Switch/Coupler/Tremulant/Button in the ODF?

Also, MIDIInputNumber set to zero was previously a default which meant no connection/association at all. It seems to be used at least for the setter button prev as a connection value now.

From my initial tests the Keyboard, Enclosure and (setter) Buttons InitialMIDI seems to work just like before and nothing is done for the ODF set MIDIInputNumber of any Stop/Switch/Coupler/Tremulant (rather obvious, since there's nothing to (auto)connect them to in the InitialMIDI tab either).

larspalo:
> There is a feature request for customizing the initial midi settings list https://github.com/GrandOrgue/grandorgue/issues/1974.

Yes, but this is for only built-in functionalty of the different panels.

> I think it would be nice to add some "standard" InitialMidiNumbers for stops. For example, 1, 2, 3, etc for pedal stops,
101, 102, 103 etc for the stops of the first manual etc.

That would "hard" limit the stop/switch/tremulant/divisional connection to 99 for each manual/division while there's no limit in the ODF itself for basically any object except from number 1 to 999. So, basically, that's useless unless you're assuming some old limited "convention" that's no longer valid.

The connection from ODF object 999 number to initial MIDI number was removed long time ago (actually it was used from a property named StopControlMIDIKeyNumber). There's currently no "standard" way of defining what [Stop999], [Coupler999], [Divisional999], [Tremulant999], [Switch999] or [General999] of the ODF should be referencing. It would all depend on what organ is modelled and the taste of the sample set producer.

For me to approve of anything even remotely like this, there must be a better abstraction layer between the user connections and the ODF defined objects from the sample set producer.

oleg68:
1. I never proposed to use the Stop/Switch/tremulant number as a connection number.
2. Why MIDIInputNumber is not a proper abstraction level?
3. This PR is not about a new feature. It is just about simplifiication the code because the old code brought several errors after small changes. 
4. I'm not going to document the capability MIDIInputNumber for button-based objects now because I agree with you that it requires more thinking.

larspalo:
> I never proposed to use the Stop/Switch/tremulant number as a connection number.

But it's with this PR now a "valid" entry in the ODF.

> Why MIDIInputNumber is not a proper abstraction level?

Why? Because there's currently no standard connection that a sample set creator can define a MIDIInputNumber for a certain stop/switch that the user can connect at will to his specific console. Which is by the way the real use case for the MIDIInputNumber to begin with.

If we'd be satisfied with a simple solution like connecting [Stop001] to input number 1 or [Switch001] to number 1 that could be done even without the MIDIInputNumber, but the trick with using the MIDIInputNumber is how to connect anything that a sample set creator might specify in the ODF to what an user can sensibly expect to auto connect to his specific console. In other words, how to sensibly specify in the ODF that for instance that this is the Principal 8' (Open Diapason) of the Great that the user can have pre-configured a MIDI message to match what his console is using for that "button"...

> This PR is not about a new feature. It is just about simplifiication the code because the old code brought several errors after small changes.

No it's not really only that anymore, as it introduces new properties, although currently "un-used".

> I'm not going to document the capability MIDIInputNumber for button-based objects now because I agree with you that it requires more thinking.

Great that we agree on that at least!

oleg68:
@larspalo I removed the capability to specify MIDIInputNumber in ODF. The code became a little bit more complex.

Could you approve this PR because it blocks submitting a final PR for #1740

src/grandorgue/midi/objects/GOMidiReceivingSendingObject.h
```diff
@@ -36,11 +39,24 @@ class GOMidiReceivingSendingObject : public GOMidiSendingObject,
...
+  int GetMidiInputNumber() const { return m_MidiInputNumber; }
+  void SetMidiInputNumber(int midiInputNumber) {
+    m_MidiInputNumber = midiInputNumber;
+  }
 
   virtual void SetElementId(int id) override;
 
 protected:
+  // Now it is present only for the symmetry with Load
+  // TODO: add the initilMidiNumber parameter
```
larspalo, line 51:
```suggestion
  // TODO: add the initialMidiNumber parameter
```

larspalo (approved):

oleg68:

PR #2137, opened 2025-03-15
Added const and static qualifiers to some members of GOMidiReceiverBase and GOMidiSender

oleg68:
This PR adds const and static qualifiers to some member functions of GOMidiReceiverBase and GOMidiSender

It also renames that members according to the current GrandOrgue naming conventions.


rousseldenis (approved):

oleg68:
@larspalo Could you approve this PR?

larspalo (approved):

PR #2141, opened 2025-03-27
Renamed class IniFileEnumEntry to GOConfigEnum::Entry and extracted to a separate file https://github.com/GrandOrgue/grandorgue/issues/1199

oleg68:
Relates to #1199

Earlier the converting between enum names and enum values was internal in the `GOConfigReader` class, so it could be used only for reading and writing ODF/CMB files.

This PR introduces a new class `GOConfigEnum` with two functions `getName` and `getValue`, so they can me used, for example, for export and import midi settings in yaml format.

It is just refactoring. No GrandOrgue behavior should be changes.


oleg68:

oleg68:
@larspalo @rousseldenis Are you planning to review this and another PRs?

rousseldenis (approved):
Great!

PR #2147, opened 2025-04-01
Removed support of linux distributions with GLIBC versions before 2.35 (ubuntu<22, debian<12, fedora<36, oracle linux<9)

oleg68:
Since Github drops support of ubuntu 20 runners (https://github.com/actions/runner-images/issues/11101), this PR switches building from ubuntu 20 to ubuntu 22.04 runners.

It causes GrandOrgue won't be more installable on linux distributions with glibc<2.35 https://repology.org/project/glibc/versions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,39 +97,39 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - run_on: ubuntu-20.04
+          - run_on: ubuntu-22.04
             for: linux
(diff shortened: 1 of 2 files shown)


larspalo (approved):
Yes, I think we'll just have to accept that end of life of Ubuntu 20.04 is in may this year... 24.04 will be supported to 2029 though, so that's likely my next step personally.

oleg68:

rousseldenis (approved):

PR #2149, opened 2025-04-05
Moved GOKeyConvert from core to grandorgue and introduced the GOKeyConvert class

oleg68:
The file name `GOKeyConvert` corresponded to the naming convention of class names but did not contain the `GOKeyConvert` class.

It also resided in the `core` subdirectory but iw was used only by grandorgue classes.

This PR
- Moves this file to the grandorgue directory
- Introduces the `GOKeyConvert` class.
- Makes all other structures and functions declared in this file inside the `GOKeyConvert` class.
- Renames them according to the GrandOrgue naming conventions.

It is a small moving/renaming. No GO behavior should be changed.


oleg68:

larspalo (approved):

PR #2177, opened 2025-04-26
Fixed GOTestWindchest

oleg68:
GOTestWindchest was flacky, because it depended on the result of GOWindchest::GetRank(index) when index was over the  number of ranks.

This PR makes the behavior more stable.


rousseldenis:
> GOTestWindchest was flacky, because it depended on the result of GOWindchest::GetRank(index) when index was over the number of ranks.
> 
> This PR makes the behavior more stable.

Great!

I'm happy to see tests will raise some implementation gaps!

rousseldenis (approved):

larspalo (approved):

PR #2146, opened 2025-04-01
Added the "Context" and "Configured" columns to the Midi Objects dialog https://github.com/GrandOrgue/grandorgue/issues/1740

oleg68:
Resolves #1740

This PR
- Adds a `Context` column to the MIDI object dialogs that shows what the midi object belongs to.
- Adds filling the contect information to the most of MIDI objects
- Add a `Configured` column that shows are there any MIDI events assigned to the object

PR #2180, opened 2025-04-30
Implemented an export and import midi settings https://github.com/GrandOrgue/grandorgue/issues/1199

oleg68:
Resolves: #1199

This PR adds the `Export` and `Import` buttons to the MidiObjects dialog.

Now it is possible to export all midi assignments of all midi objects to a yaml file and to import them.

I've already started the discussion with @willeke1234 about the implementation at #1199. Let's continue it here.

The intermediate builds are available at https://github.com/oleg68/GrandOrgue-official/actions with `feature/export-midi`.


larspalo:
@oleg68 Just to be clear, while I certainly won't object to this feature as such, I consider it of much less value than getting a full support of initial MIDI configurations (of the built-in features) inside of GO. Editing text files manually is a rather outdated way of doing things.

willeke1234:
I did a quick test. None of the issues are fixed and GO keeps crashing. This feature is not ready for use.

@larspalo 

> Editing text files manually is a rather outdated way of doing things.

What I had in mind, instead of listen for event and copy the event to a send event for every button on the panels of a sample set:
1. export the MIDI settings
2. open the file in a spreadsheet app (columns: panel, button name, MIDI device, channel, event, low value, high value etc.)
3. add and/or change MIDI events (copy, paste, fill)
4. Save as text file
5. import the MIDI settings

I originally requested this feature so my remote control app could add five events to each button, but I won't be using it.

larspalo:
> What I had in mind, instead of listen for event and copy the event to a send event for every button on the panels of a sample set:

For all the built-in GO panels and functionality it would be much better handled if it was possible to do from the "Initial MIDI". That is done only once and will they apply to all (new/un-saved) sample sets you open.

The procedure you suggest is not even remotely close to what I'd like to be able to do. But again, I won't object to this feature as such...

willeke1234:
@larspalo How would I set the initial MIDI events for Flute 8' in the "Initial MIDI"? What if I want to modify my console and add buttons for the metronome? What if I want an overview of all MIDI events? What you'd like to do is something different and it is not mutually exclusive.

Maybe there should be a priorities discussion thread. It now feels like the functionality I would like to see is no good, because something else needs to take higher priority.

larspalo:
@willeke1234 Of course only the GO built-in functionality and some very basic stuff that would be the same for **every** organ on the physical console makes sense to set by Initial MIDI. But that would still be a huge time saver, at least for me, if it was complete.

> How would I set the initial MIDI events for Flute 8' in the "Initial MIDI"?

You don't.

> What you'd like to do is something different and it is not mutually exclusive.

Of course.

> It now feels like the functionality I would like to see is no good, because something else needs to take higher priority.

My opinion is only mine and doesn't need to be anyone else's, nor does it have higher priority by default over others.

I've repeatedly said that I won't object to this feature as such. I just tried to say that I feel that improvement of the Initial MIDI really would be (more) useful. As Oleg currently is doing most of the work it's certainly his decision to pick whatever he's interested in working on as _the_ priority.

oleg68:
@willeke1234 

> None of the issues are fixed and GO keeps crashing

Let's start with crashing.

Could you test the last release (3.15.4-1) and the last intermediate build (https://github.com/GrandOrgue/grandorgue/actions/runs/14752884470): does the craching exist? If so, please open a separate issue and attach your GrandOrgueConfig there.

willeke1234:
@Oleg I have decided to stick with GO 3.7 and I'm not going to report bugs anymore. I think it's not the same crash as before. Steps to reproduce: Open de MIDI Objects window, Reload.

oleg68:
> Steps to reproduce: Open de MIDI Objects window, Reload.

Thank you.

This crash is not caused by this PR directly. It was introduced earlier with #2146. I submitted a fix as #2181.

This fix is also applied to this PR.

oleg68:
@larspalo 

> I feel that improvement of the Initial MIDI really would be (more) useful

That improvement is planned fro implementing as #1974. I will start working on it after finish working with this PR.

oleg68:
@larspalo Now I'm working on the issues mentioned at https://github.com/GrandOrgue/grandorgue/issues/1199#issuecomment-2817145318 so future commits are comming soon.

oleg68:
@willeke1234 

I made most of changes according your feedback.

> Issues:
> * The names of the buttons are used as keys but the names aren't always unique (Chapelet, Burea Funeral Chapel, Metronome). 

There are buttons with the same name, but within different contexts, So they are not mixed in the yaml file.

> The names of the buttons on the standard panels are translated and so are the keys. This is a problem if a translation is updated or if a yaml file is used with a different language.

Fixed.

> * The file format is not documented in the Help. How do I add a MIDI event to a button?

To copy from another existent event

> * The Configured column in the MIDI objects window isn't updated after import.

Fixed.

> * The imported values aren't checked. I can import illegal values and they are saved in the cmb files. Unknown keys are not reported.

I've added validation to import MIDI settings from yaml.

> * If I edit the yaml file and mess up the yaml format then I get two error messages.

Could you give an example?

> * If I change the MIDI event of a manual and export the MIDI settings then sometimes the previous MIDI event is exported.

Have you pressed `Apply`?

> * MIDI is spelled in capitals.

Fixed.

willeke1234:
>The names of the buttons are used as keys but the names aren't always unique

Chapelet:
Chapelet

Burea Funeral Chapel:
Burea Funeral Chapel

Metronome:
Metronome

>If I edit the yaml file and mess up the yaml format then I get two error messages.

Invalid yaml examples (demo organ):
[Invalid yamls.zip](https://github.com/user-attachments/files/20034474/Invalid.yamls.zip)

>If I change the MIDI event of a manual and export the MIDI settings then sometimes the previous MIDI event is exported.

It happened every now and then and I can't reproduce it. Maybe I left the MIDI event dialog open.

oleg68:
@willeke1234 I changed the columns of the Midi Object Dialog. Now there is a `Path` column that contains the context and the name of the object. The path corresponds the yaml  hierarchy. You can see exactly the same names that appear in the yaml file.

The Chapelet looks like this:

image

And the metronome:

image

The only issue remains is related to invalid yaml. I'm still working on it.

oleg68:
@willeke1234 I fixed all issues you mentioned. Could you test from https://github.com/oleg68/GrandOrgue-official/actions/runs/14946261997 ?

willeke1234:
@oleg68 The names are not required to be unique.

Chapelet (ODF downloaded in 2022, probably from MPS Orgelseite)
Scherm­afbeelding 2025-05-10 om 17 46 36

Burea Funeral Chapel
Scherm­afbeelding 2025-05-10 om 18 04 56

The other issues are fixed.

oleg68:
@willeke1234

1. I converted Chapelet from HW to GO with OdfEdit. All names are unique.
2. I downloaded Burea Funeral Chapel. It also has objects with unique names.

image

oleg68:
@larspalo @rousseldenis I finished forking on this PR. Could you approve it?

willeke1234:
@oleg68 Can a sample set be incompatible? I can't use ODFEdit. I downloaded Burea Funeral Chapel again. It still has duplicate names, did you test both versions? Is the name required to be unique? Is this documented in the Help?

oleg68:
> Can a sample set be incompatible?

I'd say more precisely: a sample set is incompatible not with Grangorgue, but only with one particular GrandOrgue feature.

> I can't use ODFEdit. 

[Chapelet Spanish Collection DEMO.zip](https://github.com/user-attachments/files/20148696/Chapelet.Spanish.Collection.DEMO.zip)

> I downloaded Burea Funeral Chapel again. It still has duplicate names, did you test both versions?

Yes, you are right. I tested the "Extended" version, it does not have duplicate names. The original version still has duplicate names.

> Is the name required to be unique? Is this documented in the Help?

Earlier it was not required. But I think it would be a good enchancement to make it required and to put a warning if some names are not unique.

larspalo:
>  But I think it would be a good enchancement to make it required and to put a warning if some names are not unique.

No! It's nonsense!

larspalo:
@oleg68 Tell me, why does Ranks show up in MIDI objects dialog?

oleg68:
> @oleg68 Tell me, why does Ranks show up in MIDI objects dialog?

Because they may send MIDI events.

larspalo:
> Because they may send MIDI events.

Why wouldn't that be covered by the drawstop?

oleg68:
Ranks send Note events to external sound modules.

larspalo:
> Ranks send Note events to external sound modules.

I'd expect that it would be the manual keys or drawstops that sent the note values to trigger any external sound module. What would the exact use case for sending from a rank really be that cannot be done in another way?

oleg68:
> > Ranks send Note events to external sound modules.
> 
> I'd expect that it would be the manual keys or drawstops that sent the note values to trigger any external sound module. What would the exact use case for sending from a rank really be that cannot be done in another way?

The first use case when an external sound module can play sounds of one rank or more ranks, but cannot switch stops. In this case we use GrandOrgue for switching stops and for calculation, what ranks are active now and sending NoteOn and NoteOff events exactly to that ranks.

The second use case is a hybrid organ: when GrandOrgue plays sound internally of some ranks and sends Note events of another ranks.

Anyway this functionality were existing before this PR and is not changed here. There are Rank objects in the MIDI objects dialog in previous GrandOrgue versions too. So if you want to remove ranks from the dialog, please create a separate issue.

larspalo:
> Anyway this functionality were existing before this PR and is not changed here. There are Rank objects in the MIDI objects dialog in previous GrandOrgue versions too. So if you want to remove ranks from the dialog, please create a separate issue.

Thanks, I don't like the long list in this dialog either so I'll consider that.

Anyway, for the uniqueness topic - it would be pretty easy for you to add the stop number in the path to solve it. In the case of Bureå funeral chapel the original organ actually has drawstop knobs without any indication of what is bass and descant. In a future version of that sample set I'd make just a single rank that the two drawstops would reference different parts of. It would be easy to give the drawstops unique names but it should not ever be a requirement!

oleg68:
> it would be pretty easy for you to add the stop number in the path to solve it

So the paths for stops would be

`manuals.001.stops.002` instead of `manuals.001.stops.Rohrfloete 8'` ?

larspalo:
> So the paths for stops would be manuals.001.stops.002 instead of manuals.001.stops.Rohrfloete 8' ?

Yes, that would ensure uniqueness. Though, I'm not sure if it really helps the user so much.

oleg68:
@larspalo @willeke1234 I added suport of not-unique MIDI object names.

See the intermediate build here: https://github.com/oleg68/GrandOrgue-official/actions/runs/15074999383

willeke1234:
@oleg68 How will users use MIDI export/import, how do you add MIDI events? Could you please explain the yaml hierarchy and put it in the Help? Or do users have to reverse engineer the hierarchy and retype the paths? The paths are not built consistently. And how do you get from a button on a panel to the path?

In short, I have my doubts about how usable MIDI export/import is as it is now.

oleg68:
> @oleg68 How will users use MIDI export/import, how do you add MIDI events? Could you please explain the yaml hierarchy and put it in the Help? Or do users have to reverse engineer the hierarchy and retype the paths? The paths are not built consistently. And how do you get from a button on a panel to the path?

The MIDI Objects dialog shows the actual paths of buttons.

Feel free to create a PR with adding the yaml hierarchy to the Help.

larspalo:
> Or do users have to reverse engineer the hierarchy and retype the paths?

The only thing I've tested is to create one configuration that had everything working as I wanted it for one sample set and then exported it. Next I created a new exported "empty" file for another sample set and copied over the things I wanted from the first file manually to the second in a text editor. After that I imported the modified file. It's not a very user friendly approach, but it seems to be working.

oleg68:

willeke1234:
@oleg68 Writing a manual or help function forces a developer to look at software from the user's perspective. I was hoping you would see the problems.

larspalo:
@oleg68 Sorry that it takes so long time for me to test things, but here are some observations that I think needs attention before we merge this.

1. The Coupler manuals and volume panel items for the enclosures are name sensitive (and actually translated) which means that they are currently not portable by copy and paste operations only (except possibly for the Master and Metronome, but even they could change when language is changed).
2. Many (most) of the elements are actually translated which will make the import fail if language is changed.
3. The false value of Use Note Off (0x8X) in the Send tab is not handled correctly

I think that the identification names (and path) for at least all the built in features should be more "general" (i.e. not translated) or replaced with more neutral identificators that won't change. (Enclosure Windchest 001 etc.) This would also make them more portable to different sample sets without needing modifications.

The Note Off seems to be correctly written to the .yaml file:

```
banked-generals:
  General01:
    name: 1
    receive:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        low_value: 0
        high_value: 1
    send:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        note_off: false
        low_value: 0
        high_value: 127
```
but gives the error message:

`22:25:14: Error: The value of banked-generals.General01.send[0].note_off (false) is not a valid bool`

when I import the file and thus the lights on the device won't be turned off correctly when they should.

oleg68:
@larspalo @willeke1234 

Whould it be easier if I changed the yaml structure to be more "flat".

From:
```
banked-generals:
  General01:
    name: 1
    receive:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        low_value: 0
        high_value: 1
    send:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        note_off: false
        low_value: 0
        high_value: 127```
Sequencer:
  Prev:
    receive:
      - device1: "jack: Midi-Bridge:OrganConsoleAllMidi (capture)"
        event_type: ControlChange
        channel: 3
        key: 64
        low_value: 0
        high_value: 1
        debounce_time: 1000
    shortcut:
      shortcut_key: left
  Next:
    receive:
      - device: "jack: Midi-Bridge:OrganConsoleAllMidi (capture)"
        event_type: ControlChange
        channel: 2
        key: 64
        low_value: 0
        high_value: 1
        debounce_time: 1000
    shortcut:
      shortcut_key: right
```

to

```
objects:
  - path: banked-generals.General01
    name: 1
    receive:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        low_value: 0
        high_value: 1
    send:
      - device: "alsa: APC mini mk2 APC mini mk2 Contr"
        event_type: Note
        channel: 1
        key: 100
        note_off: false
        low_value: 0
        high_value: 127
  - path: Sequencer.Prev
    receive:
      - device: "jack: Midi-Bridge:OrganConsoleAllMidi (capture)"
        event_type: ControlChange
        channel: 3
        key: 64
        low_value: 0
        high_value: 1
        debounce_time: 1000
    shortcut:
      shortcut_key: left
  - path: Sequencer.Next
    receive:
      - device: "jack: Midi-Bridge:OrganConsoleAllMidi (capture)"
        event_type: ControlChange
        channel: 2
        key: 64
        low_value: 0
        high_value: 1
        debounce_time: 1000
    shortcut:
      shortcut_key: right
```

In that case you need to copy-paste an existing object, to change its's path to the value displayed in the Midi Object dialog and to alter some values. You wouldn't need to know the path hierarchy.

larspalo:
> Whould it be easier if I changed the yaml structure to be more "flat".

I think it's perhaps a bit easier and less error prone.

> In that case you need to copy-paste an existing object, to change its's path to the value displayed in the Midi Object dialog and to alter some values.

With regards to the suggested usage and possible "instructions" for this feature, the best approach I can think of for an user would be to:

- Create one fully defined organ for the largest sample set you can imagine for your particular console and export it.
- For any new organ:
  - Export an "empty" .yaml file for the new organ.
  - Copy over the parts you want/need from the fully configured file to the exported one for the new organ.
  - Modify names and paths in the new organ .yaml file that are unique to the sample set as necessary (the MIDI Object dialog can be a help as you said)
  - Import the modified .yaml file and verify that it's working and if so - Save the organ to a preset.

However, I think that it would be extremely helpful if the paths would be as generalized as possible (as I noted in https://github.com/GrandOrgue/grandorgue/pull/2180#issuecomment-2905761575) and that the names wouldn't create errors so easily as it currently does.

oleg68:
@larspalo Yes, I've finished with this PR and I'm waiting for your approval.

larspalo (approved):

rousseldenis (approved):

PR #2186, opened 2025-05-04
Introduced class GOStringSet

oleg68:
This is a very small PR that allows to use a set of wxString outside the `GOConfigReader` class.


rousseldenis (approved):
Good idea.

larspalo (approved):

Issue #2185, opened 2025-05-02
Crashes


Bumblebee001:
I am using the latest version of GO.

I am noticing that after a couple of reloads necessary whilst working on creating a new sample-set, GO crashes and closes. To prevent this, I am loading the odf from Open Recent.

oleg68:
> I am using the latest version of GO.

Which one: 3.15.4-1 or an intermediate build?

> I am noticing that after a couple of reloads necessary whilst working on creating a new sample-set, GO crashes and closes. To prevent this, I am loading the odf from Open Recent.

1. So File->Reload causes crash and File->Open Recent does crash, do they?
2. What operating system do you run?
3. What GrandOrgue windows are open when you do `Reload`?

Bumblebee001:
-     I am using 3.15.4-1
-     Reload does not crash for the first possibly second time but then it does. Open recent has never crashed so far even after having the organ already open.
-     Windows 64x
-     The main tool bar and the console (used to appear as one single "unit" / window before you took over but for a while have been appearing as 2 separate windows which is rather annoying I must admit; the Toolbar is always at top right half outside the screen and the console screen never maximised).

![Image](https://github.com/user-attachments/assets/8ccc2db3-0018-49ae-a9ad-4b42ce5dfdd2)

Mark

Bumblebee001:
Is there a record kept by GO?

Bumblebee001:
I am sending a crashdump of one of the smaller files, dated today.

[GrandOrgue.exe.293668.dmp](https://github.com/user-attachments/files/20033654/GrandOrgue.exe.293668.dmp)

Bumblebee001:
What am I supposed to install?

oleg68:
> What am I supposed to install?

Nothing to install. Just download https://github.com/oleg68/GrandOrgue-official/actions/runs/14891010699/artifacts/3081055641 , unpack to some directory and run GrandOrgue.exe from there.

Bumblebee001:
![Image](https://github.com/user-attachments/assets/c2540cd2-d039-45bf-92e1-79a01897267b)

oleg68:
@Bumblebee001  `ios_base_library_initv` should be in the `libstdc++-6.dll` library.

This error looks like you have extracted only GrandOrgue.exe from the .zip archive and haven't extracted other files.

You should extract **all** files from the .zip archive to some directory and run GrandOrgue.exe from there.

Bumblebee001:
[...]

PR #2192, opened 2025-05-10
Fixed crash on reloading an organ https://github.com/GrandOrgue/grandorgue/issues/2185

oleg68:
Resolves: #2185

The reason of the crash was using a GOTimer instance that had already deleted from another objects that had not yet been deleted.

This PR changes the deletion order.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2197, opened 2025-05-17
Introduced class GOMidiPlayingObject

oleg68:
For enchancing the initial MIDI tab of the organ settings, it has to operate with an GOMidiObject  instead of an GOMidiReceiver.

Earlier GOMidiObject was depending on the GOOrganModel instance of the current open organ, but the initial MIDI tab must be available even when there is no any current organ open.

So this PR extracts a part of GOMidiObject into a separate class GOMidiPlayingObject.


rousseldenis (approved):
Great!

oleg68:
@larspalo could you approve this PR?

larspalo (approved):

PR #2207, opened 2025-06-03
Removed references from GOMidiReceiver to GOConfig

oleg68:
This PR removes the looped dependency between GOMidiReceiver and GOConfig.

Now only GOConfig depends on GOMidiReceiver.

It is just refactoring. No GO Behavior should be changed.

### DIFF
--- a/src/grandorgue/config/GOConfig.cpp
+++ b/src/grandorgue/config/GOConfig.cpp
@@ -251,7 +251,7 @@ GOConfig::GOConfig(wxString instance)
 }
 
 GOOrgan *GOConfig::CloneOrgan(const GOOrgan &newOrgan) const {
-  return new GORegisteredOrgan(*this, newOrgan);
+  return new GORegisteredOrgan(newOrgan);
 }
 
 bool GOConfig::IsValidOrgan(const GOOrgan *pOrgan) const {
@@ -415,7 +415,7 @@ void GOConfig::Load() {
       cfg, MIDI_PORTS, GOMidiPortFactory::getInstance(), m_MidiPortsConfig);
 
     for (unsigned i = 0; i < GetEventCount(); i++)
-      m_MIDIEvents[i]->Load(cfg, GetEventSection(i), m_MidiMap);
+      m_MIDIEvents[i]->Load(ODFCheck(), cfg, GetEventSection(i), m_MidiMap);
 
     long cpus = wxThread::GetCPUCount();
     if (cpus == -1)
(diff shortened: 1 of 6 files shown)


larspalo:
@oleg68 While testing this PR (and #2206 just before this but at that time the send events was still working without me finding any issues) I noticed that the programmed/saved send events stopped working due to the MIDI output devices suddenly was disabled and thus removed from the events. I noted that the MIDI device name is changed (with a dash and number from earlier without). So, likely something is changed in how the devices are named.

Enabling the device again (with the new name) and explicitly adding it back to the event got it working. Also, if send events was saved without a device and just sent to whatever that listened they also started working after enabling the MIDI device again.

larspalo:
@oleg68 Unfortunately I cannot reliably reproduce the issue which makes it so much harder to track down. It happens rather randomly and not at all every time. Also the naming obviously can differ as you'll see below... I have had lots of crashes with GO lately, though it's likely to not be (only) directly related to this PR, but I'd suspect that they could very well be caused by the MIDI devices, as usually GO then crashes either directly at startup or when the Settings dialog is invoked.

This is result from my local build when everything is working as it should:

Skärmbild från 2025-06-07 21-19-23

And this is from one time that the bug manifests itself with the build from this PR action:

Skärmbild från 2025-06-07 21-21-59

Note that the device suddenly was automatically disabled (ie. not done by me) and even lost it's name completely this time! What happens is then that the APC mini can still be used as input, but the send events of course no longer works so I don't get the buttons to light up as they should.

I think I'll need to download this PR locally and build a debug version that I can run in the debugger to get a better understanding of what's happening.

oleg68:
Does the issue persist with the last GO release?

larspalo:
> Does the issue persist with the last GO release?

No, I've never been able to reproduce the issue with anything other than the action build for this PR and even then very rarely.

I've tested a local debug build for this PR with about 50 starts of which none so far would exhibit the issue but where some 15 starts/loadings/Settings dialog would crash for some reason but all (traced in gdb) seem to mostly be due to different things in the wxWidgets library (gtk) rather than anything directly in the GO source itself.

Thus, I'm almost inclined to say that it might not be this PR at all that caused the strange send event issues I encountered. If you don't have any other idea that I can try, I think that we could just let this PR through and see what happens later on, or what do you think?

oleg68:
I agree with you.
Please attach here the crash stacktrace. I'll try to investigate.

larspalo:
@oleg68 As I said, I don't think the crashes really are specific to this PR. However, even though I've not been able to reproduce the original issue that I experienced with this particular PR I've had plenty of crashes at different situations while testing. All of the attached crash backtraces attached were performed on a local debug build from this PR.

[GoCrashes.zip](https://github.com/user-attachments/files/20646545/GoCrashes.zip)

For now, I'll just approve of this PR as I really cannot see anything in particular that should be causing problems.

larspalo (approved):

PR #2216, opened 2025-06-13
Added the GOConfigMidiObject class

oleg68:
This PR adds the class GOConfigMidiObject and replaces GOMidiReceiver instances in GOConfig with GOConfigMidiObject ones.

Now only the GOReceiver element of GOConfigMidiObject is used so no GrandOrgue behavior should be changed. But in the future all four possible elements will be used.


oleg68:

rousseldenis (approved):

PR #2219, opened 2025-06-25
Introduced the class GOGUIMidiControl

oleg68:
Earlier each GUI control class (GOGUIButton, GOGUIEnclosure, GUGOManual, GOGUILabel) had an own code of handling right-click events and displaying the MIDI event dialog.

This PR introduces a single place GOGUIMidiControl for this code.


oleg68:
@larspalo I fixed the visualisation

PR #2220, opened 2025-06-28
Replaced wxListView with wxGrid in GOSettingsMidiInitial

oleg68:
This is a next PR related to #1974

It 
1. replaces the wx control is used for listing the initial MIDI objects;
2. renames some GOSettingsMidiInitial members;
3. renames the column titles: `Element` -> `Name`, `MIDI Event` -> `Configured`.

I'm going to add more columns to this grid, so I'm not change the screenshot now.


rousseldenis:
@oleg68 I think it should be a little bit enhanced:

- Missing vertical lines to separate columns
- Default horizontal size to size of window
image

larspalo:
@oleg68 @rousseldenis It's likely due to how the wxGrid is displayed by default. I've got no vertical lines on Ubuntu 24.04 with this PR in File->Settings->Initial MIDI or in Audio/MIDI->MIDI Objects (as before) and the size is never working correctly just the same as in any GOGrid that's based of wxGrid (like with the available organs to load). Specific size restoring in any wxGrid has never worked reliably, and it seems to be even worse so under Wayland but there could be other reasons too...

larspalo (approved):

rousseldenis (approved):

PR #2226, opened 2025-07-04
Added passing GOMidiObject to GOMidiEventDialog

oleg68:
Earlier the MIDI dialog constructor was accepted the instances of all four MIDI elements independently.

Now it accepts a single GOMidiObject instance instead.

Because there this dialog may be used for GORegisteredOrgan that does not inherit from GOMidiObject, an auxiliary constructor has been created that accepts a GOMidiReceiver instance.


larspalo (approved):

oleg68:

PR #2238, opened 2025-07-16
Added capability of settings Save, Division and Shortcut elements of Initial MIDI objects https://github.com/GrandOrgue/grandorgue/issues/1974

oleg68:
Now it is possible to edit all elements of InitialMIDI objects with the Properties dialog from the Initial MIDI settings tab.

It also changed the GrandOrgueConfig file format, so after saving it with the new version and restoring it with the old one the Initial MIDI settings would be lost. But after reading an old-version GrandOrgueConfig all initial MIDI settings should be imported successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Added capability of settings Save, Division and Shortcut elements of Initial MIDI objects https://github.com/GrandOrgue/grandorgue/issues/1974
+- Added capability of setting Send, Division and Shortcut events for Initial MIDI objects https://github.com/GrandOrgue/grandorgue/issues/1974
 - Reduced decimal numbers for PitchLabel https://github.com/GrandOrgue/grandorgue/issues/2234
 - Added support of exporting current midi settings to a text file and of importing them from both text and .cmb files https://github.com/GrandOrgue/grandorgue/issues/1199
 - Fixed crash on reloading an organ https://github.com/GrandOrgue/grandorgue/issues/2185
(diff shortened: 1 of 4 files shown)


larspalo:
@oleg68 Sorry, but I I'm not certain I fully understand your instructions above. Please correct me where/if I don't follow.

With this PR we can now set Receive, Send and Shortcut etc. events of all the already existing InitialMIDI elements, whereas earlier we could only set the Receive event, right?

When we test this PR any previously existing InitialMIDI configurations will be detected and "converted" to the new format that later will be saved to the config file when it's overwritten, correct?

If we revert to another (earlier) version of GO after testing with this PR all the InitialMIDI will be lost as the older version won't recognize the new format, right?

If we have a backup of the GrandOrgueConfig from before testing this PR and restore it after having done the testing,  overwriting the one just created by the GO version of this PR, any previous version of GO that we might be using will restore our InitialMIDI configuration as the older format then can be read, right?

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Added capability of settings Save, Division and Shortcut elements of Initial MIDI objects https://github.com/GrandOrgue/grandorgue/issues/1974
```
larspalo, line 1: Perhaps it would be better phrased as:

```suggestion
- Added capability of setting Send, Division and Shortcut events for Initial MIDI objects https://github.com/GrandOrgue/grandorgue/issues/1974
```

larspalo (approved):
Just suggesting an optional edit to the CHANGELOG to better describe actually what this is about.

rousseldenis (approved):

PR #2240, opened 2025-07-18
Changed matching of Initial MIDI buttons

oleg68:
Earlier most of the Setter and Metronome buttons had an unique MidiInputNumber. It was used for matching with initial MIDI settings.

Unlike Manuals and Enclosures the buttons' MidiInputNumber could not be used for assignment in the ODF. Providing  unique numbers for was a complicated and not obvious task.

This PR replaces matching by MidiInputNumber with matching by path. So the buttons do not more require MidiInputNumber.


larspalo:
I can now see that the events seems to be set in the InitialMIDI tab for the buttons but they all have the same data (for me it's 4) and actually work from that button but that's not really working!

larspalo:
Another observation is that also buttons that I didn't define InitialMIDI events for now also got the same device, channel and data as the ones I did set... This means that events are defined for all buttons on the InitialMIDI and they all share the same physical console button!

oleg68:
@larspalo I fixed reading the old GrandOrgueConfig.

I had to restore filling MIDIInitialNumber in GOConfig because it is used for matching initial receivers with old configs.

oleg68:

rousseldenis (approved):

PR #2249, opened 2025-07-26
Reworked the MIDI initial tab of settings https://github.com/GrandOrgue/grandorgue/issues/1974

oleg68:
Resolves: #1974

This is a final PR related to #1974.

It:

- Adds the `To Initial` buttons to the right-click MIDI event editor as well as to the MIDI object dialog. It allows to add the current editer/selected MIDI object(s) to the Initial MIDI configuration.
- They appear in the `User-added` group
- Adds more columns to the Initial MIDI tab.
- Adds the `Delete` button to the tab. It allows to remove selected user-added MIDI objects from the initial MIDI configuration.


rousseldenis (approved):

larspalo:
I've tested this PR a little and mostly it works fine and will be an excellent enhancement to get this into GO.

One issue I've run into is that even though I have the master volume enclosure (volumes/Master) programmed and added into the InitialMIDI from right-click MIDI dialog it's not applied to newly opened organs while all other volume enclosures (volumes/Windchest001 etc.) are. Actually, I can even see that the volumes/Master is configured in InitialMIDI but **not** in the Midi-Objects dialog list so something is obviously not in sync.

Another small UI thing I noted is that there's absolutely no confirmation that the item/event was added to InitialMIDI when pushing the ToInitial button. Luckily, pressing it multiple times won't create multiple versions of it.

oleg68:
> One issue I've run into is that even though I have the master volume enclosure (volumes/Master) programmed and added into the InitialMIDI from right-click MIDI dialog it's not applied to newly opened organs while all other volume enclosures (volumes/Windchest001 etc.) are. Actually, I can even see that the volumes/Master is configured in InitialMIDI but **not** in the Midi-Objects dialog list so something is obviously not in sync.

I couldn't reproduce this. Could you provide a step-by-step instruction?

larspalo:
> I doubt that these confirmations would be good, because they would increase the number of clicks necessary for adding lots of objects to the Initial MIDI configuration.

You're right that we don't need any additional message dialogs. I won't push for it now, but it would be really great if we had some kind of notification (checkbox or whatever kind of alerting widget - could even be a simple static text) that this item already is added to the initial MIDI configuration.

What is important for now is to get the volumes/Master to behave correctly.

larspalo:
> I couldn't reproduce this. Could you provide a step-by-step instruction?

Sure.

- Open any sample set, I use Bureå gravkapell extended with Swedish language in GO
- Open Panels->Coupler manuals and volume
- Listen for event for Master volume and set it to a potentiometer/slider and then "To Initial" before closing the dialog
- Repeat previous step for pedal, manual and metronome.
- Close the organ without saving (it should now be in initial MIDI)
- Open the same or another organ and then Panels->Coupler manuals and volume
- Test the potentiometers/sliders and see that the master volume won't work, actually metronome won't work either with the same sample set, but the other windchests with numbers will work.

So, I'd suspect that the problem is again with translations and the path detection. volumes/Master won't match in any other language than English while the numbered windchests will work. Metronome won't actually work either.

oleg68:
> So, I'd suspect that the problem is again with translations and the path detection. volumes/Master won't match in any other language than English while the numbered windchests will work. Metronome won't actually work either.

Thank you, Lars. I coudn't reproduce this effect with the English language but I managed to reproduce it with the Swedish language. I'll do more investigations.

oleg68:
> > I couldn't reproduce this. Could you provide a step-by-step instruction?
> 
> Sure.
> 
> * Open any sample set, I use Bureå gravkapell extended with Swedish language in GO
> * Open Panels->Coupler manuals and volume
> * Listen for event for Master volume and set it to a potentiometer/slider and then "To Initial" before closing the dialog
> * Repeat previous step for pedal, manual and metronome.
> * Close the organ without saving (it should now be in initial MIDI)
> * Open the same or another organ and then Panels->Coupler manuals and volume
> * Test the potentiometers/sliders and see that the master volume won't work, actually metronome won't work either with the same sample set, but the other windchests with numbers will work.
> 
> So, I'd suspect that the problem is again with translations and the path detection. volumes/Master won't match in any other language than English while the numbered windchests will work. Metronome won't actually work either.

oleg68:
> You're right that we don't need any additional message dialogs. I won't push for it now, but it would be really great if we had some kind of notification (checkbox or whatever kind of alerting widget - could even be a simple static text) that this item already is added to the initial MIDI configuration.

Would be sufficient displaying a message in the Log window?

oleg68:
@larspalo I fxed the inhertance of the Master Volue.

 I also tested the Metronome ON button with the French language. It seems to work. What metronome buttons exactly do not work n your enironments?

larspalo:
> I also tested the Metronome ON button with the French language. It seems to work. What metronome buttons do not work n your enironments exactly?

Not the buttons, the enclosure for the Metronome windchest was not stored just like with the volumes/Master. I guess it's the same with the translation Metronom that is different from Metronome...

oleg68:
> One possible solution would be to simply disable the \"To Initial\" button if the element already exist.

I does not like this idea because sometimes a user needs to change the initial MIDI configuration for an object.

larspalo:
> I does not like this idea because sometimes a user needs to change the initial MIDI configuration for an object.

Well, it would still be possible to do from the InitialMIDI tab with perhaps a few more clicks, but I really don't think the current behavior of this PR is any more user friendly as it makes it far too easy to mess things up without even realizing it.

Again, I don't think it's really that important though, so if you want to leave it as it is for now I won't argue about it.

oleg68:
@larspalo I fxed the initial settings of the Metronome volume enclosure.

> so if you want to leave it as it is for now I won't argue about it

Would be better to log the result of `To Initial`: whether it was added or replaced?

larspalo:
> Would be better to log the result of To Initial: whether it was added or replaced?

I'm doubtful about using the log window for that kind of reason. For me, the log window should be mainly be used for warnings or errors.

If you don't want a simple wxStaticText, what about dynamically changing the "To initial" button text if element already exists in InitialMIDI? It could be "To initial (exists)" or "To initial (added)" just to give the user a hint but not stopping a replacement.

src/grandorgue/midi/objects/GOMidiObject.h
```diff
@@ -55,8 +55,11 @@ class GOMidiObject : public GOSaveableObject, public GOMidiConfigDispatcher {
   GOMidiMap &r_MidiMap;
   ObjectType m_ObjectType;
 
-  wxString m_NameForContext;
+  // a name that may be translated (for built-in objects)
   wxString m_name;
+  // a name tat is never translated. May be empty that causes that m_name is
```
larspalo, line 60:
```suggestion
  // a name that is never translated. May be empty which causes that m_name is
```

src/grandorgue/model/GOWindchest.h
```diff
@@ -26,6 +26,10 @@ class GOWindchest : private GOSoundStateHandler {
 private:
   GOOrganModel &r_OrganModel;
   wxString m_Name;
+  // a name tat is never translated. May be empty that causes that m_name is
```
larspalo, line 29:
```suggestion
  // a name that is never translated. May be empty which causes that m_name is
```

larspalo (approved):
This is absolutely good enough for now. Thank you for your work on this! I've left a few suggestions on a few choices of words that's mainly optional.

Issue #2278, opened 2025-09-10
alsa connections via pipewire are not helpful


William-Purvis:
I've just switched our church organ to use pipewire instead of jack. As part of the change I switched the inputs from various scanners to use the alsa interface, rather than the jack interface. However, the inputs from the scanners, and outputs back to them show up in the dialogues as Midi-in:1 etc.
These are not very helpful, as the scanners explicitly set their output names to be meaningful. The connection map using qpwgraph shows the names but these names do not get through to the GO dialogs. Could this be changed to show the device names, as they were when using Jack.
Alternatively, is there an easy way to find the correspondence?

oleg68:
@William-Purvis Could you post here the screenshots of Settings->Midi Devices?

William-Purvis:
Sorry, 
Bad habit just responding to the emails....

Here's the three files (I hope)

<img width="661" height="1047" alt="Image" src="https://github.com/user-attachments/assets/d996e7eb-714c-4486-ba34-26523568187a" />

<img width="737" height="555" alt="Image" src="https://github.com/user-attachments/assets/4b542947-cb7d-48b2-8a10-139d2cf90333" />

![Image](https://github.com/user-attachments/assets/3de7affd-21e3-4c8b-903c-30d2101996f0)

oleg68:
@William-Purvis You can press on the `Matching...` button on the Settings->MIDI Devices tab and change the Logical Name for all MIDI devices. Then you will see these names in the MIDI event editor dialog.

PR #2304, opened 2025-11-21
Added validation of the device logical names for leading/trailing spaces in the Device Matching Dialog https://github.com/GrandOrgue/grandorgue/issues/2278

oleg68:
Resolves: #2278

Because the problems mantioned in the issue were caused by a trailing space in the device name, I added the validation of the name

1. When the Device Matching Dialog becomes open.
2. When the Logical name is changed.


oleg68:

oleg68:

PR #2311, opened 2025-11-29
Changed types of GOMidiEventPattern members

oleg68:
This PR replaces field types `int` and `unsigned` of MIDI-related structures with `uint16_t` and `in8_t`.

It is just a small refactoring. No GO behaviur should be changed.

### DIFF
--- a/src/grandorgue/midi/events/GOMidiEventPattern.cpp
+++ b/src/grandorgue/midi/events/GOMidiEventPattern.cpp
@@ -41,10 +41,14 @@ void GOMidiEventPattern::DeviceIdFromYaml(
 }
 
 int GOMidiEventPattern::convertValueBetweenRanges(
-  int srcValue, int srcLow, int srcHigh, int dstLow, int dstHigh) {
-  const int dstAbsLow = std::min(dstLow, dstHigh);
-  const int dstAbsHigh = std::max(dstLow, dstHigh);
-  int dstValue = srcValue - srcLow;
+  int_fast8_t srcValue,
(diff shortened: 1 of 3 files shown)


src/grandorgue/midi/events/GOMidiEventPattern.cpp
```diff
@@ -41,10 +41,14 @@ void GOMidiEventPattern::DeviceIdFromYaml(
 }
 
 int GOMidiEventPattern::convertValueBetweenRanges(
-  int srcValue, int srcLow, int srcHigh, int dstLow, int dstHigh) {
-  const int dstAbsLow = std::min(dstLow, dstHigh);
-  const int dstAbsHigh = std::max(dstLow, dstHigh);
-  int dstValue = srcValue - srcLow;
+  int8_t srcValue,
```
rousseldenis: @oleg68 For my undertandings, I saw there is also a 'fast' type `int_fast8_t`. Do you know the difference with `int8_t` and if we should not need the fast implementation instead ? Thanks

oleg68: Thank you for a very good question.

`int_fast8_t` is a preferable datatype for local variables and parameters because it may offer better performance where `int_t` is preferable for data structures.

But in both x86_64 and arm64 architectures (that GrandOrgue is built on GitHub for) the implementation of these types are the same so it might bring any advantages only for some exothic architectures. And _fast types are used rarely in comparison to int ones for now.

rousseldenis (approved):

Issue #2281, opened 2025-09-27
Import MIDI settings doesn't report invalid device names


William-Purvis:
When importing a MIDI device file (.yaml) GO should report any invalid device names. It does report syntax errors, but if the device name has been mis-spelled, or the device names have been changed, it should at least issue a warning message, preferably with a line number. I wasted several days trying to import a .yaml file which included minor typos and the result was various entries with 'Any device', which for a 'send' entry seems to mean no device.

PR #2318, opened 2025-12-11
Renamed GOMidiMap::GetDeviceIdByLogicalName to EnsureLogicalName. Added a read-only version GOMidiMap::GetDeviceIdByLogicalName

oleg68:
A next PR is related to #2281

Earlier GOMidiMap::GetDeviceIdByLogicalName added automatically the device name to the map if it had not existed before.

1. It did not follow the name conventions for getters: GetXXX should not have a side effect
2. It is not always desired because we need do something (ex. put a warning message) if the device has been missed during import.

So this PR 
1. Makes GOMidiMap::GetDeviceIdByLogicalName read only
2. Adds an auto-inserting version of the method: GOMidiMap::EnsureLogicalName
3. Replaces all old calls to GOMidiMap::GetDeviceIdByLogicalNam with GOMidiMap::EnsureLogicalName

### DIFF
--- a/src/grandorgue/midi/GOMidi.cpp
+++ b/src/grandorgue/midi/GOMidi.cpp
@@ -60,7 +60,7 @@ void GOMidi::Open() {
     if (pDevConf && pDevConf->m_IsEnabled)
       ((GOMidiInPort *)pPort)
         ->Open(
-          m_MidiMap.GetDeviceIdByLogicalName(pDevConf->GetLogicalName()),
+          m_MidiMap.EnsureLogicalName(pDevConf->GetLogicalName()),
           pDevConf->m_ChannelShift);
     else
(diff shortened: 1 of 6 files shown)


oleg68:
@larspalo 

```
--2025-12-11 15:48:57--  https://github.com/GrandOrgue/YamlCppAdd/releases/download/0.6.2-5.go/libyaml-cpp-static_0.6.2-5.go_all.deb
Resolving github.com (github.com)... 140.82.113.4
Connecting to github.com (github.com)|140.82.113.4|:443... connected.
HTTP request sent, awaiting response... 503 Service Unavailable
2025-12-11 15:48:59 ERROR 503: Service Unavailable.
```

I think there are some temporary issues with github environment.

PR #2349, opened 2025-12-31
Removed the file GOUtil

oleg68:
The files GOUtil.h and GOUtil.cpp contained functions for parsing and formating double and float values. It fully covered with already existing functions std::stol, std::stod and wxString::Format. So they may be removed safely.

This PR.
1. Removes GOUtil.h and GOUtil.cpp
2. Replaces usage of their content with std::stol, std::stod and wxString::Format.

### DIFF
--- a/src/core/config/GOConfigReader.cpp
+++ b/src/core/config/GOConfigReader.cpp
@@ -1,12 +1,13 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2025 GrandOrgue contributors (see AUTHORS)
+ * Copyright 2009-2026 GrandOrgue contributors (see AUTHORS)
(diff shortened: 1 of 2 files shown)


larspalo:
@oleg68 You can try:

- start GO in a console with this PR
- switching to Swedish language in GO settings
- open any sample set
- close the sample set and exit GO
- open GO again (from command line/console) and watch the warnings appear

At least that's what I've done to receive the warnings about locale dependent floating points that seem to be written to the GO config file (more specifically: in the Audio device section Device001Channel001Group001Left and similar entries, MemoryLimit and ReverbGain). Note that I've done the closing/starting GO a few times though, don't know how many exactly... Also, I've had quite a lot of crashes while testing this PR.

oleg68:

Issue #2258, opened 2025-08-12
Metronome sound change to a "click" instead of a "beep"


efwalcker:
First of all, thank you to the authors of this great software!

I wanted to ask if there is a possibility to change or to have an option of choosing the metronome ticking sound to be a "click" like on a Seiko metronome or a mechanical metronome, instead of the high-pitched "beep"? This might be subjective, but the "beep" is somewhat distracting and even sometimes annoying, and I believe that I am not the only one who finds it problematic. I would very much appreciate that. Thank you!

larspalo:
If I remember correctly I sampled the metronome sounds from my Boss DB-30 or some other digital metronome I had back then. Replacing the samples is a fairly simple task but finding an analog or a digital metronome with clicks that everyone would like could be a bit harder...

Until the samples have been changed, might I remind you that you can somewhat customize the metronome sounds in GO yourself. There are two types of clicks available, one for the first beat of the measure and another for the other beats so if you set the metronome measure beat number to 0 you only get the less high pitched beats (and with 1 you only get the really high pitched ones). Some additional tweaking of the metronome sounds can also be made in the Audio/MIDI->Organ settings dialog just as with any other samples.

larspalo:
Actually, as noted in #376 it's perfectly possible for an user to exchange the beat.wv and first_beat.wv files in the .zip (or tar.gz) packages and have the metronome sounds changed. You can find them in the share/GrandOrgue/sounds/metronome/ directory. The samples in the distributed GO packages are packed with WavPack to reduce size but I'd expect that any ordinary .wav file that is just re-named to match the .wv would work just as well.

larspalo:
Just for testing purposes I generated the attached metronome clicks in Audacity though I'm not personally convinced that they really are better but it indeed works fine to just replace the files in the share/GrandOrgue/sounds/metronome/ directory.

[MetronomeBeats.zip](https://github.com/user-attachments/files/21728967/MetronomeBeats.zip)

oleg68:
I think there should be a user option to select a metronome sound among a set of predefined ones with GUI. Or, at least, a capability of playing a user-added wav-file.

These options should be easy configured with GUI.

I'd like to go by the first option. But the most complex thing is to resolve licensing issues. I doubt that we may just to grab the metronome sound from a youtube video and use it in GrandOrgue without permission from it's author. @rousseldenis could you help to research this thing?

efwalcker:
[...]

PR #2354, opened 2026-01-10
Added capability of changing metronome sound https://github.com/GrandOrgue/grandorgue/issues/2258

oleg68:
Resolves: #2258

This PR
1. Moves initial metronome settings from the "Options" tab to the new "Metronome" tab.
2. Adds capability to choose from the old "Bell" metronome sound, the new "Click" one and a custom sound where a user provides two wave files for the first beat and for other beats.

### DIFF
--- a/src/core/settings/GOSettingFile.cpp
+++ b/src/core/settings/GOSettingFile.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2024 GrandOrgue contributors (see AUTHORS)
+ * Copyright 2009-2026 GrandOrgue contributors (see AUTHORS)
  * License GPL-2.0 or later
  * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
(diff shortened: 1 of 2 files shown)


oleg68:

larspalo:
This PR seems to work ok. It's a bit clumsy to have to re-load a sample set when changing the sounds, but I guess an user will not really change the metronome sound so often but rather set it to what they want and then just use that... Caches will need to be updated too if a change of meteronome sound is made. Somehow, I almost feel that currently the meteronome is "in between" as a global sound but still require the windchest of a sample set to function. Perhaps it would have been better if the meteronome choice was either saved with the sample set, or completely independent of the sample set as a pure global choice. It's perhaps much beyond the scope of this PR though...

When selecting the custom files, it would have been nice if the path would have been remembered for the other file chooser after having set (any) one of them instead of having to browse through all the same directories again for the files. At least, once a path have been specified for each of them, GO will remember that for the next choice.

rousseldenis (approved):

PR #2356, opened 2026-01-10
Fixed unability to check for updates on slow internet connections

oleg68:
Sometimes GrandOrgue couldn't check for updates with a wrong json format error.

The reason was a timeout occured and only a part of responce json was fetched.

This PR

1. Adds check for errors after finishing fetching?
2. Replaces a timeout with a minimum of fetching speed.
3. Reduces of number of last releases read by GrandOrgue from 100 to 3

So now check for update works more stable.


oleg68:

PR #2367, opened 2026-01-18
When importing yaml midi settings, set device to 0 if not present.

tim-br:
When importing midi settings into midi objects, if the device is not present it comes up as missing and has to manually be changed to the correct device. This PR sets any midi object to any device if the device is not present.

Also changes import behaviour so that existing midi objects are not overwritten unless explicitly set in the import yaml file. The means we can import midi objects while maintaining existing midi objects.

  - path: manuals/000/switches/001
    name: Sousbasse 16'
    receive:
      - event_type: Note
        channel: 16
        key: 36


oleg68:
@tim-br Thank your for your contribution.

You put three different changes in one PR. Some of them are valid, some are not.

It'd be better to post three different PRs instead one complex.

PR #2373, opened 2026-01-24
Removed file GOUtil

oleg68:
This PR removed custom implementation of conversions between string and numbers in flavor to the standard one.

It also moved the check for trailing spaces from different methods of GOConfigReader to `Read`.

@larspalo I reworked the previous PR. The issue with old ODF was related to empty values, ex. `ShortcutKey=`. Now they are processed correctly.

### DIFF
--- a/src/core/config/GOConfigReader.cpp
+++ b/src/core/config/GOConfigReader.cpp
@@ -362,12 +362,12 @@ int GOConfigReader::ReadInteger(
     return defaultValue;
 
   std::string strValue = value.ToStdString();
-  std::size_t nParcedChars;
+  std::size_t nParsedChars;
   long retval;
   wxString errMsg;
 
   try {
-    retval = std::stol(strValue, &nParcedChars);
+    retval = std::stol(strValue, &nParsedChars);
   } catch (const std::exception &exc) {
     throw wxString::Format(
       _("Invalid integer value '%s' at section '%s' entry '%s': %s"),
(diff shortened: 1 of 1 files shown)


oleg68:

src/core/config/GOConfigReader.cpp
```diff
@@ -356,41 +358,34 @@ int GOConfigReader::ReadInteger(
   bool required,
   int defaultValue) {
   wxString value;
-  if (!Read(type, group, key, required, value))
+  if (!Read(type, group, key, required, true, value))
     return defaultValue;
 
+  std::string strValue = value.ToStdString();
+  std::size_t nParcedChars;
```
larspalo:
```suggestion
  std::size_t nParsedChars;
```

src/core/config/GOConfigReader.cpp
```diff
@@ -356,41 +358,34 @@ int GOConfigReader::ReadInteger(
...
-      wxLogWarning(
-        _("Trailing whitespace at section '%s' entry '%s': %s"),
-        group.c_str(),
-        key.c_str(),
-        value.c_str());
-    value.Trim();
+  wxString errMsg;
+
+  try {
+    retval = std::stol(strValue, &nParcedChars);
```
larspalo:
```suggestion
    retval = std::stol(strValue, &nParsedChars);
```

Issue #2348, opened 2025-12-30
Convolution reverb randomly stops working after a few minutes of activating it


ashervahalik:
I have some sample sets that are quite dry, so I have a set of convolution reverb files that I use with them to improve the sound. But since about 2 versions ago, the convolution reverb has been randomly deciding to quit working after a while, and I have to open Settings/Reverb and click OK to restart the reverb. I checked all of the logs, and everything looks normal; I also checked the gain levels on each of the affected organs because that can be an issue too, but that did not appear to be the case here.

oleg68:
What does "Reverb stop working" mean? Does GrandOrgue produce no sound at all? Or it produces sound but it is dry?

ashervahalik:
What I mean is that occasionally GO will randomly act as if the “Enable Convolution Reverb” toggle has been turned off, playing just the raw, dry samples without the convolution reverb. Hitting the panic button does temporarily fix it.

oleg68:
@ashervahalik 

1. does it depend on a convolution response file? 
2. Could you post one of it here?
3. What operation system do you use?
3. Post here the screenshot of the Audio settings tab.
4.

PR #2380, opened 2026-01-26
Fixed switching the convolution reverb off when the system is overloaded https://github.com/GrandOrgue/grandorgue/issues/2348

oleg68:
#Resolves: #2348

The Zita convolver has a counter of number of times when it was late to process. When it became 5, the convolution stops working.

This PR sets an option of convolver that forces itself to continue even it is late many times.


oleg68:

PR #2394, opened 2026-02-08
Added GOSoundBuffer and it's subclasses for the future use

oleg68:
GrandOrgue processes sound in float[] buffers. It passes them across different classes. There are some difficulries:

- The metadata (channel count, samples count) sometimes are also passed, sometimes aren't and it is stored independently. So the call order impacts to the right processing, and it's changing may bring errors.
- There are lots of similar low-level code (filling with 0, copying, adding) in the code

This PR introduces a new abstraction `GOSoundBuffer` that consists of a pointer to a float[] buffer and the metadata. Thee provide basic low-level operations with buffers.

The class hierarchy:

- GOSoundBuffer
    - GOSoundBufferMutable
        - GOSoundBufferMutableMono
        - GOSoundBufferManaged
        
The unit tests (AI-generated) are also present here.

This is just adding of new code that is not in use currently. No GO behavior should be changed.

### DIFF
--- a/src/tests/testing/sound/buffer/GOTestPerfSoundBufferMutable.cpp
+++ b/src/tests/testing/sound/buffer/GOTestPerfSoundBufferMutable.cpp
@@ -38,10 +38,10 @@ struct Baseline {
 // Baselines are set ~10% below the minimum observed value across all CI runs.
 static constexpr Baseline BASELINE_FILL_WITH_SILENCE[] = {
 #ifdef NDEBUG
-  {32, 7500},   // 7500 Mframes/sec (raised for modern hardware)
-  {128, 5900},  // 5900 Mframes/sec (measured: 6636.1, -10% margin)
-  {512, 5500},  // 5500 Mframes/sec (lowered: min observed 6112.9, -10% margin)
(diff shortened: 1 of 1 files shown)


oleg68:

oleg68:
Thank you, @larspalo.

> SoundItem, nItems, OffsetInItems then for all individual floats.

SoundItem and SoundFrame are good names. I will use the shorten form of them `Item` and `Frame` in variable and function names in the context of SoundBuffer, because Sound is already present in the class name.

I will replace the names in the code.

Another concern is using the `Offset` term. I would replace it with `Index`. What do you thing about it?

larspalo:
> Another concern is using the Offset term. I would replace it with Index. What do you thing about it?

Index is absolute in its meaning so I think it's easy to understand. Offset has more of the feeling of being relative to something.

oleg68:
@larspalo @rousseldenis 

I have fixed all the comments you made. There are lots of other changes based on this PR, but I cannot submit them now because this PR is not merged.

Could you approve this PR as soon as possible?

larspalo (approved):

rousseldenis (approved):

PR #2452, opened 2026-03-13
GOSoundStream: extracted common code to helpers

oleg68:
This PR introduces the new GOSoundStream functions: getDecodeBlockFunctionFor, GoToStartSegment, InitFromSection. It simplifies the code of InitStream, InitAlignedStream and ReadBlock.

It also adds some unit tests for these functions.

### DIFF
--- a/src/grandorgue/sound/playing/GOSoundStream.cpp
+++ b/src/grandorgue/sound/playing/GOSoundStream.cpp
@@ -127,7 +127,7 @@ void GOSoundStream::DecodeBlock(float *pOut, unsigned nOutSamples) {
 
 GOSoundStream::DecodeBlockFunction GOSoundStream::getDecodeBlockFunction(
   uint8_t nChannels,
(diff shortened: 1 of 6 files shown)


larspalo:
@oleg68 Not nice to remove older comments... This PR is still is not working. The loops are still having clicks and are actually cancelled one after another so that even if the key is held it will be silence! Not acceptable!

larspalo:
@oleg68 This is what the debug log window show at such situations:

Skärmbild från 2026-05-24 21-51-41

oleg68:
@larspalo I fixed the regression and added a test that would catch it. Could you test [the new version](https://github.com/GrandOrgue/grandorgue/actions/runs/26680269537?pr=2452)?

larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2460, opened 2026-03-28
Reordered getters and setters in GOSoundOrganEngine

oleg68:
Now GOSoundOrganEngine starting up and shutting down are very fragile: there are lots of setters that affect to the result of other setters. So their calling order is important, and calling setters in a wrong order may cause incorrect operatings or even a crash.

I'm going to make starting/stopping more robust. The first step is documenting existing setters and the required calling order.

This PR splits the setters into three groups: Independent, Dependent and State getters. It also adds comments to setters: what of them depend on other.

It also reorders their declaration and their code in the intended calling order.

No GrandOrgue methods code are changed. So no GO behavior should be changed.


oleg68:

rousseldenis (approved):

larspalo (approved):

PR #2463, opened 2026-03-31
Bugfix/open sound asan

oleg68:
Earlier:
 - Debug symbols on Linux were never shipped with GitHub build artifacts, making it impossible to diagnose crash backtraces
 - Windows PDB generation (cv2pdb) was unconditional — no way to disable it
 - All GitHub builds (release and intermediate) used the `Release` build type
 - ASAN was not supported in CI or build scripts

 This PR does:
 - Adds `RELEASE_FLAG` CMake option — unifies release/debug build type selection across all platform build scripts
 - Adds `GO_SPLIT_DEBUG_SYMBOLS` CMake option (default: ON on Windows always; on Linux = RELEASE_FLAG) — extracts `.dbg` via `objcopy --only-keep-debug` on Linux, gates PDB generation on Windows
 - Now all intermediate GitHub builds use the `Debug` build type; release builds use the `Release` build type
 - Adds `GO_BUILD_ASAN` CMake option with `-fsanitize=address -fno-omit-frame-pointer -static-libasan`, wired through build scripts and CI via `workflow_dispatch` `asan` boolean input
 - Adds capability of triggering ASAN builds from the GitHub Actions UI

This is a GitHub CI/build-only change. GO runtime behavior is unchanged. Intermediate builds now use the `Debug` build type, but the performance impact is negligible thanks to [c4ec548f](https://github.com/GrandOrgue/grandorgue/commit/c4ec548f).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,13 @@
 name: Build
 
-on: [push]
+on:
+  push:
+  workflow_dispatch:
+    inputs:
+      asan:
+        description: 'Build with ASAN'
+        type: boolean
+        default: false
 jobs:
   tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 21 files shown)


oleg68:
Is replaced with #2476

PR #2464, opened 2026-03-31
Moved some classes to gui/, renamed classes

oleg68:
- GODocument → GOGuiOrgan (src/grandorgue/gui/GOGuiOrgan.h/cpp)
- GOApp → GOGuiApp (src/grandorgue/gui/GOGuiApp.h/cpp)
- GOView → GODocumentView (src/grandorgue/gui/document-base/GODocumentView.h/cpp)
- document-base/ → gui/document-base/
- Renamed GOFrame::mp_doc → mp_organ, GOFrame::r_app type updated.
- GOLog to gui/, renamed to GOGuiLog
- GOImageCache to gui/, renamed to GOGuiImageCache
- Renamed GOFrame to GOToolbarWindow

It is just moving and renaming. No GO behavior should be changed.

### DIFF
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -104,7 +104,7 @@ gui/dialogs/GOSplash.cpp
 gui/dialogs/go-message-boxes.cpp
 gui/document-base/GODocumentBase.cpp
 gui/document-base/GODocumentView.cpp
-gui/frames/GOToolbarWindow.cpp
+gui/frames/GOAppWindow.cpp
 gui/frames/GOLogWindow.cpp
 gui/frames/GOMainWindowData.cpp
 gui/frames/GOStopsWindow.cpp
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -65,8 +65,9 @@ control/GOElementCreator.cpp
 control/GOEventDistributor.cpp
 control/GOLabelControl.cpp
 control/GOPistonControl.cpp
-document-base/GODocumentBase.cpp
-document-base/GOView.cpp
+gui/GOGuiImageCache.cpp
+gui/GOGuiLog.cpp
+gui/GOGuiOrgan.cpp
 gui/dialogs/common/GODialogCloser.cpp
 gui/dialogs/common/GODialogSizeSet.cpp
 gui/dialogs/common/GODialogTab.cpp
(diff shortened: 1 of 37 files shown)


larspalo:
@oleg68 Sorry for being absent for a while. One thing I'm a bit concerned about is the renaming of GOFrame to GOToolbarWindow. The main frame of the application is more than just the toolbar so I would suggest a different naming that reflect that this is actually the main application window instead of just a window with a toolbar.

oleg68:
@larspalo Is `GOMainApplicationWindow` better?

larspalo:
> Is GOMainApplicationWindow better?

Yes, I think so.

oleg68:
> > Is GOMainApplicationWindow better?
> 
> Yes, I think so.

But GrandOrgue users usualy do not understand that the main window is one with a menu and a toolbar. They mention the largest organ panel as a "Main window". So "Toolbar Window" is more understandable for users.

Maybe there is another name that would be quite precise and would not confuse users?

AI promised "GOControlPanelWindow" and "GOAppControlWindow". It also suggested "GOMainMenuWindow", but I disagree, because this window has not only menu and the main thing for user is not the same as the main thing for a programmer.

larspalo:
@oleg68 MainAppWindow or even MainAppFrame would be most precise depending if you take the user or developer perspective.

larspalo:
> They mention the largest organ panel as a "Main window".

Only with a certain sample set loaded. Otherwise it won't even be visible! Thus, the main panel of a sample set is not the same as the app main panel.

larspalo:
Or, in this case actually the frame...

oleg68:
> Or, in this case actually the frame...

Frame is not understandable by a user at all.

What do you think on `GOAppControlWindow`? Is it acceptable?

larspalo:
@oleg68 What about just GOAppWindow? If you're not happy with that I can live with GOAppControlWindow. But GOAppFrame would no doubt describe the file content best.

> Frame is not understandable by a user at all

A normal user isn't expected to bother with the file naming anyway since they most likely never even will look in the code or the source tree. The naming should make it easy for developers, and the main point here is that it's rather important to know where the main frame of the application actually is.

oleg68:
> What about just GOAppWindow? 

OK. Let's GO with `GOAppWindow`.

oleg68:
@larspalo I renamed it to `GOAppWindow`

larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2469, opened 2026-04-04
Fix Turkish İ issue

### LINKED ISSUE #2401: Turkish related uppercase error
I was testing the application and got the following error.

It happens because the application is uppercasing the `medium` string based on the system locale. 

In Turkish, we have two letters, I and İ (ı and i)

`ı` = `I`

`i` = `İ`

Therefore, the `medium` string becomes `MEDİUM` instead of `MEDIUM`, and causes some unwanted errors. Like the following

Image

ghost:
Fix the letter I became İ in Turkish

Fixes: https://github.com/GrandOrgue/grandorgue/issues/2401

### DIFF
--- a/src/core/config/GOConfigReader.cpp
+++ b/src/core/config/GOConfigReader.cpp
@@ -290,6 +290,8 @@ GOLogicalColour GOConfigReader::ReadColor(
     value = defaultValue;
 
   value.MakeUpper();
+  // Fix I became İ in Turkish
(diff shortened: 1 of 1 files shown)


oleg68 (changes requested):
It is not a correct fix.

The right way is to find where the Turkish I appears in cfg.

oleg68:
Replaced with #2471

PR #2470, opened 2026-04-05
Introduced GOProgressMonitor interface to decouple loading from GUI

oleg68:
This PR Replaces GOProgressDialog* parameters in GOOrganController::Load() and UpdateCache() with GOProgressMonitor& so model-layer code no longer depends on a GUI dialog class. GOProgressDialog now inherits from GOProgressMonitor and overrides all three methods.

This is just refactoring. No GO behavior should be changed.


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2481, opened 2026-04-26
Updated performance test baselines based on recent CI failures

oleg68:
Lowered 13 baseline thresholds to reflect lower throughput observed on Azure VM runners (April 2026). Each value is set to min-observed × 0.9 (-10% margin) per the existing convention.

It affects only tests. No GO behavior should bec changed.


larspalo (approved):

oleg68:
@larspalo @rousseldenis could you approve this PR as soon as possible? Otherwise I receive a lot of spam that performance tests fail.

PR #2496, opened 2026-05-23
Fixed expression pedal malfunction when CC value exceeds configured high value #2494

### LINKED ISSUE #2494: Expression pedal response to continuous controller is wrong in 3.17.11
I recently upgraded from 3.16.? to 3.17.1-1 and I find that the expression shoes respond differently to a continuous controller input.  Previously  if I set the midi settings as shown:

Image

in 3.16 the expression shoe would be fully on at continuous controller values of 115 and above, and start closing once the cc value falls below 115. this is useful if the maximum value of the cc varies with temperature or supply voltage, and also allows one to prevent random CC messages when setting the midi settings for other components.

In 3.17.1-1 when the cc controller value rises above 115 the expression shoe suddenly closes completely as though GrandOrgue had some sort of internal overflow problem.

oleg68:
## Summary

- Fixed `int_fast8_t` overflow in `GOMidiEventPattern::convertValueBetweenRanges` that caused enclosures to close completely when a CC value exceeded the configured `high_value`
- Replaced manual clamp logic with `int dstValue` + `std::clamp` to prevent overflow before clamping
- Added changelog entry

## Root cause

Commit bb9b80e1 (#2311) changed the local variable `dstValue` from `int` to `int_fast8_t`. When `srcValue > srcHigh`, the intermediate interpolation result can exceed 127, which overflowed `int_fast8_t` (wrapping to −128) before the clamp could correct it — causing the enclosure to snap fully closed.

## Test plan

- [ ] Configure an enclosure to respond to CC range 0–115
- [ ] Send CC values above 115 — enclosure should remain fully open, not close
- [ ] Send CC values below 0 — enclosure should remain fully closed
- [ ] Verify normal range (0–127) still maps correctly

Fixes https://github.com/GrandOrgue/grandorgue/issues/2494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
 # 3.17.2 (2026-05-24)
 - Fixed matching MIDI and audio device names containing regex metacharacters (e.g. parentheses) https://github.com/GrandOrgue/grandorgue/issues/2491
 - Fixed installing the .deb package on Ubuntu with PipeWire https://github.com/GrandOrgue/grandorgue/issues/2457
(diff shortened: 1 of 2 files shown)


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2520, opened 2026-06-09
Removed usage of GOOrganController from GOMainWindowData

oleg68:
Removes the direct dependency on GOOrganController from GOMainWindowData so the class depends only on the model layer (GOOrganModel), which is a prerequisite for moving GOMainWindowData ownership to GOGuiOrgan.


oleg68:

oleg68:
@larspalo @rousseldenis Will you look at this PR?

rousseldenis (approved):

oleg68:
Excuse me, @larspalo , but I'm merging this PR because you do not answer.

PR #2527, opened 2026-06-19
Fixed wxWidgets assertion fail (and crash of intermediate builds) when opening Settings

oleg68:
** Root case

GODialog::Init() unconditionally enabled wxDIALOG_ADAPTATION_MODE_ENABLED for all dialogs, including GOTabbedDialog (based on wxPropertySheetDialog). On Show(), wxStandardDialogLayoutAdapter created a wxScrolledWindow and called SetContainingWindow() on the main sizer, which asserted that every child window's parent equals that wxScrolledWindow — but notebook page children are parented to their pages, not to the adapter's window.

** Fix

Guard the call with if constexpr so layout adaptation is only enabled for plain wxDialog subclasses, not wxPropertySheetDialog ones.

### DIFF
--- a/src/grandorgue/gui/dialogs/common/GODialog.h
+++ b/src/grandorgue/gui/dialogs/common/GODialog.h
@@ -1,6 +1,6 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2024 GrandOrgue contributors (see AUTHORS)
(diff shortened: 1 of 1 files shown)


larspalo (approved):

oleg68:

rousseldenis (approved):

PR #2528, opened 2026-06-20
Fix assert(nearestFilledCell) in GOSoundReleaseAlignTable Phase 2

oleg68:
This PR fixes crash mentioned at https://github.com/GrandOrgue/grandorgue/issues/2494#issuecomment-4699382332

## Summary

- **Bug fixed**: `GOSoundReleaseAlignTable::ComputeTable` Phase 2 could fire `assert(nearestFilledCell)` when Phase 1 filled only one cell and an unfilled cell sat directly in the same derivative row. The old bit-trick formula never generated `derivDelta=0`, so same-row neighbors were unreachable.
- **Fix**: replaced the formula with `signed_offset()` producing `0, +1, −1, +2, −2, …` for both dimensions — a correct separable outward scan with derivative as the outer loop.
- **Refactor**: extracted the search into `GOSoundReleaseAlignTable::findNearestFilledCell` (public static member, `std::optional<CellCoords>` return, documented in header); added `CellCoords` as a public nested struct.
- **Tests**: new `GOTestReleaseAlignTable` with 6 cases: regression for the original assert, same-row preference, distant single cell, empty table, self-find, and amplitude ordering.

## Test plan

- [x] `LANG=C make -k -j16` — clean build, no warnings in changed files
- [x] `./bin/GOTestExe --no-perf` — all 12 functional tests pass including `GOTestReleaseAlignTable`
- [ ] Load a real organ with release samples and verify release tails sound as before

### DIFF
--- a/src/grandorgue/sound/playing/GOSoundReleaseAlignTable.cpp
+++ b/src/grandorgue/sound/playing/GOSoundReleaseAlignTable.cpp
@@ -12,8 +12,6 @@
 #include "loader/cache/GOCache.h"
 #include "loader/cache/GOCacheWriter.h"
 
-#include "GOSoundAudioSection.h"
-
 #ifndef NDEBUG
 #ifdef PALIGN_DEBUG
 #include <stdio.h>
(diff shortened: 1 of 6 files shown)


oleg68:

rousseldenis (approved):

PR #2532, opened 2026-06-27
Fix heap buffer overflow in GOSoundOutputTask meter info update

### LINKED ISSUE #2531: Heap buffer overflow in GOSoundOutputTask::Run() meter info update (since 3.17.1)
## Summary

`GOSoundOutputTask::Run()` overflows `m_MeterInfo` by one element on roughly every other channel wrap-around, writing 4 bytes past the end of its heap allocation. Confirmed as a genuine heap corruption via Application Verifier's Page Heap (Windows): `Corrupted suffix pattern for heap block`, block size 8 bytes, corruption address = block start + 8 - exactly matching a 2-channel `m_MeterInfo` (2 floats = 8 bytes) overrun by one `float`.

## Root cause

In `src/grandorgue/sound/tasks/GOSoundOutputTask.cpp`, `Run()` tracks the current channel slot in `m_MeterInfo` with two redundant variables: the index `channelI` (correctly wrapped inside the loop body) and a raw pointer `pMeterInfo` (also reset inside the body on wrap-around, but *additionally* incremented unconditionally by the `for` loop's own increment clause):

```cpp
for (unsigned nItems = GetNItems(), itemI = 0, channelI = 0; itemI < nItems;
     itemI++, pData++, pMeterInfo++) {           // <-- pMeterInfo++ runs every iteration, unconditionally
  ...
  if (absF > *pMeterInfo)
    *pMeterInfo = absF;

  channelI++;
  if (channelI >= nChannels) {
    channelI = 0;
    pMeterInfo = m_MeterInfo.data();             // <-- reset inside the body
  }
}
```

A C++ `for(init; cond; step) body;` runs `step` *after* `body` on every iteration, regardless of what `body` did. So right after `pMeterInfo` is reset to `m_MeterInfo.data()` (index 0) inside the body, the loop's own `pMeterInfo++` immediately bumps it to index 1 - one step ahead of `channelI`. From there it drifts further out of sync each cycle, and every second wrap-around it ends up at `m_MeterInfo.data() + nChannels`, i.e. one element past the end of the vector.

This bug was introduced in 4f848c81f / #2418 ("GOSoundBufferTask", 3.17.1), which converted index-based access (`m_MeterInfo[c]`) to pointer-based access (`*pMeterInfo`) as part of a larger refactor. The pre-existing index variable `c` (now `channelI`) was never incremented by the loop header - only the new pointer got that extra, conflicting increment.

## Impact

This runs on every audio buffer, for every `GOSoundOutputTask`, whenever an organ is loaded and producing sound - so the overflow itself happens continuously and deterministically during playback. Whether it's ever *visible* depends entirely on what the heap allocator happens to place adjacent to this specific `m_MeterInfo` allocation at the time, which varies with the loaded organ, memory layout, and timing. This is consistent with sporadic, hard-to-reproduce crashes, including `STATUS_HEAP_CORRUPTION` (0xC0000374) seen via Windows Event Viewer, that don't reproduce reliably and appear unrelated to any specific user action.

## How it was found

While debugging an unrelated audio-device-robustness feature, a real `STATUS_HEAP_CORRUPTION` crash (reproducible specifically when triggering an audio engine restart, e.g. via Audio Panic, while an organ was loaded) turned out to reproduce identically on an unmodified release build too. Application Verifier's Heaps layer (Page Heap) on a `-Og -g` debug build pinpointed the exact corrupting write to this loop.

## Fix

See #2532 - index `m_MeterInfo` via `channelI` directly instead of maintaining a second, separately-advanced pointer.

sebion7125:
## Summary

Fixes #2531

`GOSoundOutputTask::Run()` indexed `m_MeterInfo` (one float per channel) through a raw pointer that was incremented in two independent places: unconditionally by the `for` loop's own increment clause, and additionally reset inside the loop body on channel wrap-around. The loop header's increment always runs *after* the body, so right after a wrap-around reset, the pointer immediately gets bumped one element past where it should be - drifting it out of sync with the (correctly wrapped) `channelI` index. Every second wrap-around this overflows `m_MeterInfo` by one element, writing past the end of its heap allocation.

Confirmed as a real heap corruption via Application Verifier's Page Heap on Windows (`Corrupted suffix pattern for heap block`, 8-byte block = a 2-channel `m_MeterInfo`, corruption address = block start + 8).

## Fix

Index `m_MeterInfo` via `channelI` directly (`pMeterInfo[channelI]`) instead of maintaining a second, separately-advanced pointer. `channelI` is now the only thing that determines the position, removing the possibility of drift.

```diff
   for (unsigned nItems = GetNItems(), itemI = 0, channelI = 0; itemI < nItems;
-       itemI++, pData++, pMeterInfo++) {
+       itemI++, pData++) {
     float f = std::clamp(*pData, CLAMP_MIN, CLAMP_MAX);
     float absF = std::abs(f);

     if (f != *pData)
       *pData = f;
-    if (absF > *pMeterInfo)
-      *pMeterInfo = absF;
+    if (absF > pMeterInfo[channelI])
+      pMeterInfo[channelI] = absF;

     // Move to next channel (circular: after last channel, wrap to first)
     channelI++;
-    if (channelI >= nChannels) {
+    if (channelI >= nChannels)
       channelI = 0;
-      pMeterInfo = m_MeterInfo.data();
-    }
   }
```

Introduced in 4f848c81f / #2418 (3.17.1).

## Test plan

- [x] Compiles cleanly (`-fsyntax-only` against the win64 cross-toolchain)
- [x] Root cause confirmed by reproducing the corruption under Application Verifier's Page Heap on the *unfixed* code (Audio Panic with a loaded organ, `-Og -g` build) - the report's address/size matched this exact allocation
- [ ] Re-running the same repro with this fix applied, to confirm the corruption no longer occurs, is still pending

### DIFF
--- a/src/grandorgue/sound/tasks/GOSoundOutputTask.cpp
+++ b/src/grandorgue/sound/tasks/GOSoundOutputTask.cpp
@@ -79,21 +79,19 @@ void GOSoundOutputTask::Run(GOSoundThread *pThread) {
   float *pMeterInfo = m_MeterInfo.data();
 
   for (unsigned nItems = GetNItems(), itemI = 0, channelI = 0; itemI < nItems;
-       itemI++, pData++, pMeterInfo++) {
+       itemI++, pData++) {
     float f = std::clamp(*pData, CLAMP_MIN, CLAMP_MAX);
     float absF = std::abs(f);
 
     if (f != *pData)
       *pData = f;
-    if (absF > *pMeterInfo)
-      *pMeterInfo = absF;
+    if (absF > pMeterInfo[channelI])
+      pMeterInfo[channelI] = absF;
 
     // Move to next channel (circular: after last channel, wrap to first)
     channelI++;
-    if (channelI >= nChannels) {
+    if (channelI >= nChannels)
       channelI = 0;
-      pMeterInfo = m_MeterInfo.data();
-    }
   }
 
   m_Done.store(true);


oleg68:
Prrformance is an important in real time sound processing. That is why the indexed expressions were replaced with pointer expressions. Restoring indexing is a step backwards, not forwards.

#2533 fixes the bug without a performance penalty. Moreover it replaces the counting-up loop with a countdown loop because the counting-down requires fewer registers.

PR #2538, opened 2026-06-30
Fix heap corruption in wxImage::Scale() when alpha and mask are both set

### LINKED ISSUE #2536: Crash (heap corruption) on panel resize when an image has both alpha and a mask
## Description

GrandOrgue can crash with a heap corruption when resizing a panel window, if the panel's ODF graphics include an image that has both a native alpha channel (e.g. a PNG with transparency) *and* a separate `Mask...` file.

## Root cause

`GOGuiImageCache::LoadImage()` calls `image.SetMaskFromImage(...)` to apply the mask file, which sets `HasMask() == true` regardless of whether the image already had alpha. The existing cleanup `if (bitmap->HasMask()) bitmap->InitAlpha();` is meant to normalize this, but `wxImage::InitAlpha()` is internally guarded against images that already have alpha (`wxCHECK_RET(!HasAlpha())`) and silently no-ops in that case. The result is a `wxImage` that simultaneously reports `HasAlpha() == true` and `HasMask() == true` — a combination wxWidgets does not officially support.

When `GOBitmap::BuildBitmapFrom()` later calls `wxImage::Scale(..., wxIMAGE_QUALITY_BICUBIC)` on such an image during panel resize, the dual alpha+mask state corrupts the heap and crashes the process.

This logic (and the bug) predates the 2026 GOBitmap refactors (#2390, #2433) - it traces back essentially unchanged to a 2015 commit ("Always use alpha instead of a mask").

## Steps to reproduce

1. Load an organ whose panel uses an image with native PNG alpha *and* a `Mask...` ODF entry for the same control.
2. Resize the panel window (or open it at a non-native CMB-specified size, which also triggers a rescale).
3. Process crashes with heap corruption inside `wxImage::Scale()`.

## Fix

PR incoming: fold the mask into the alpha channel at load time (set alpha to 0 for mask-colour pixels) instead of carrying both `HasAlpha()` and `HasMask()` simultaneously, and only call `InitAlpha()` when alpha isn't already present. Also hardens `GOBitmap::BuildTileBitmap()` against degenerate (zero/negative) source or target dimensions, which could independently construct a zero-area `wxImage` and crash similarly during resize.

sebion7125:
Fixes #2536

## Summary
- `GOGuiImageCache::LoadImage()` could leave a `wxImage` with both `HasAlpha()` and `HasMask()` set simultaneously, a combination wxWidgets does not officially support. `wxImage::InitAlpha()` silently no-ops when alpha is already present, so the pre-existing `if (HasMask()) InitAlpha();` cleanup never actually resolved this for images that had native alpha before a mask was layered on via `SetMaskFromImage()`.
- Scaling such a dual-state image during panel resize (`GOBitmap::BuildBitmapFrom`, BICUBIC) corrupts the heap and crashes.
- Fold the mask into the alpha channel at load time instead of carrying both states, and only call `InitAlpha()` when alpha isn't already present.
- Also harden `GOBitmap::BuildTileBitmap()` against degenerate (zero/negative) source or target dimensions, which could independently construct a zero-area `wxImage` and crash similarly during resize.

## Test plan
- [ ] Load an organ with an ODF image that has native PNG alpha and a `Mask...` file
- [ ] Resize the panel window / open it at a non-native CMB size
- [ ] Confirm no crash and the masked region still renders correctly

### DIFF
--- a/src/grandorgue/gui/GOGuiImageCache.cpp
+++ b/src/grandorgue/gui/GOGuiImageCache.cpp
@@ -9,12 +9,20 @@
 
 #include <wx/intl.h>
 #include <wx/mstream.h>
+#include <wx/stopwatch.h>
 
 #include "files/GOOpenedFile.h"
 #include "loader/GOLoaderFilename.h"
 
 #include "GOBuffer.h"
 #include "GOGuiLog.h"
+#ifndef LOG_TIMING
+#define LOG_TIMING(...)
+#endif
+#ifndef LOG_GUI_GAP
+#define LOG_GUI_GAP(...)
+#endif
+#define GUI_BITMAP_TIMING 1
 #include "Images.h"
 
 #define BITMAP_LIST                                                            \
@@ -223,6 +231,10 @@ const wxImage *GOGuiImageCache::LoadImage(
   const wxString &filename, const wxString &maskName) {
   const wxImage *pImage = FindImage(filename, maskName);
 
+#if GUI_BITMAP_TIMING
+  wxStopWatch swBmp;
+#endif
+
   if (!pImage) {
     wxImage image, maskimage;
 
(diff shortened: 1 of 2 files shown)


oleg68:
Replaced with #2540

PR #2539, opened 2026-06-30
Debounce expensive control re-render during panel resize, fix flicker https://github.com/GrandOrgue/grandorgue/issues/2537

### LINKED ISSUE #2537: Panel resize is extremely slow and flickers when dragged past native size
## Description

Two related usability problems with resizing a panel window:

1. **Resizing/opening is very slow.** Every `WM_SIZE` tick during a live drag (and the initial open of a panel at a non-native, CMB-specified size) runs the full `PrepareDraw()`/`OnUpdate()` cascade, which re-rescales every single key, stop and texture bitmap with `wxIMAGE_QUALITY_BICUBIC`. With realistic panel control counts this makes both live dragging and opening a panel at a non-native size unusably slow.

2. **Flicker/jump when dragging past the native size.** Once the window is resized larger than the panel's native (CMB) size in a way that doesn't match its aspect ratio, the panel briefly appears to jump to the top-left corner before settling into its centered position - visible as a flicker on every resize tick.

## Root cause

1. `GOGUIPanelWidget::UpdateSize()` had no debouncing - the expensive per-control BICUBIC re-render ran unconditionally on every resize event.
2. `GOGUIPanelView::OnSize()` repositions/resizes the panel widget through up to three separate native window calls per tick (`UpdateSize()`'s internal `SetSize()`, then `SetPosition(0,0)`, then conditionally `CentreOnParent()`). Without batching, the window manager can paint the in-between geometry states, which is seen as the flicker/jump described above - specifically only once centering becomes non-trivial, i.e. once the window is dragged past the size where it matches the panel's native aspect ratio.

## Fix

PR incoming:
- Debounce the expensive full redraw behind a short (150ms) timer; every resize tick in between does a cheap, smooth bilinear rescale of the last fully-rendered bitmap instead (always rescaling from that pristine bitmap rather than a previously-rescaled one, to avoid compounding blur across a multi-tick drag).
- Wrap the panel widget's multiple geometry changes in `OnSize()` in `Freeze()`/`Thaw()` so only the final, settled position/size is ever painted.

sebion7125:
Fixes #2537

## Summary
- Every `WM_SIZE` tick during a live panel resize re-ran `PrepareDraw()`/`OnUpdate()`, rescaling every key/stop/texture bitmap with BICUBIC quality - making live dragging and opening a panel at a non-native CMB size unusably slow.
- `GOGUIPanelWidget` now does a cheap bilinear rescale of the last sharp, fully-rendered bitmap on every resize tick for immediate feedback, and only re-runs the expensive per-control BICUBIC redraw once via a 150ms debounce timer after resizing goes idle. Rescaling always starts from that pristine last-full-redraw bitmap (not the already-rescaled preview), so quality doesn't degrade further across a multi-tick drag.
- Separately, `GOGUIPanelView::OnSize()` moved/resized the panel widget through up to three separate native calls per tick, which could be painted mid-sequence and looked like a flicker/jump whenever the panel was dragged past its native size and centering kicked in. Wrapped the sequence in `Freeze()`/`Thaw()` so only the final, settled geometry is ever painted.

## Test plan
- [ ] Drag-resize a panel window and confirm smooth, responsive dragging (no multi-second stalls)
- [ ] Open a panel at a non-native CMB-specified size and confirm it opens promptly
- [ ] Drag a panel window larger than its native size (aspect mismatch) and confirm no flicker/jump

### DIFF
--- a/src/grandorgue/gui/panels/GOGUIPanelView.cpp
+++ b/src/grandorgue/gui/panels/GOGUIPanelView.cpp
@@ -129,6 +129,16 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
     // Scale out the panel according the new size
 
     const wxSize newSize = event.GetSize();
+
+    // UpdateSize(), SetPosition() and CentreOnParent() below each move/resize
+    // m_panelwidget natively on their own. Without freezing, the window
+    // manager can paint the in-between states (e.g. still at the old
+    // centering position right after the size already changed), which is
+    // seen as a brief flicker/jump whenever the panel is dragged past its
+    // native size and centering kicks in. Freeze() suppresses repaints until
+    // all the geometry changes below are settled.
+    m_panelwidget->Freeze();
+
     const wxSize maxSize = m_panelwidget->UpdateSize(newSize);
 
     this->SetVirtualSize(maxSize);
@@ -156,6 +166,8 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
       m_panelwidget->CentreOnParent(wxHORIZONTAL);
     if (actualsize.GetHeight() > maxSize.GetHeight())
       m_panelwidget->CentreOnParent(wxVERTICAL);
+
+    m_panelwidget->Thaw();
   }
   event.Skip();
 }
--- a/src/grandorgue/gui/panels/GOGUIPanelView.cpp
+++ b/src/grandorgue/gui/panels/GOGUIPanelView.cpp
@@ -11,6 +11,7 @@
 #include <wx/frame.h>
 #include <wx/image.h>
 #include <wx/toplevel.h>
+#include <wx/wupdlock.h>
 
 #include "gui/wxcontrols/go_gui_utils.h"
 
@@ -67,7 +68,7 @@ GOGUIPanelView::GOGUIPanelView(
   // At this point, the new window may fill the whole display and still not be
   // large enough to show all contents. So, before showing anything, lets see
   // what scaling is needed to fit the content completely to the window.
-  wxSize scaledsize = m_panelwidget->UpdateSize(topWindow->GetClientSize());
+  wxSize scaledsize = m_panelwidget->SetInitialSize(topWindow->GetClientSize());
 
   topWindow->Show();
   topWindow->Update();
@@ -130,16 +131,16 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
 
     const wxSize newSize = event.GetSize();
 
-    // UpdateSize(), SetPosition() and CentreOnParent() below each move/resize
-    // m_panelwidget natively on their own. Without freezing, the window
-    // manager can paint the in-between states (e.g. still at the old
+    // UpdatePreviewSize(), SetPosition() and CentreOnParent() below each
+    // move/resize m_panelwidget natively on their own. Without freezing, the
+    // window manager can paint the in-between states (e.g. still at the old
     // centering position right after the size already changed), which is
     // seen as a brief flicker/jump whenever the panel is dragged past its
-    // native size and centering kicks in. Freeze() suppresses repaints until
-    // all the geometry changes below are settled.
-    m_panelwidget->Freeze();
+    // native size and centering kicks in. The update locker suppresses
+    // repaints until all the geometry changes below are settled.
+    wxWindowUpdateLocker lock(m_panelwidget);
 
-    const wxSize maxSize = m_panelwidget->UpdateSize(newSize);
+    const wxSize maxSize = m_panelwidget->UpdatePreviewSize(newSize);
 
     this->SetVirtualSize(maxSize);
 
@@ -166,8 +167,6 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
       m_panelwidget->CentreOnParent(wxHORIZONTAL);
     if (actualsize.GetHeight() > maxSize.GetHeight())
       m_panelwidget->CentreOnParent(wxVERTICAL);
-
-    m_panelwidget->Thaw();
   }
   event.Skip();
 }
--- a/src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
+++ b/src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
@@ -19,9 +19,15 @@
 
 DEFINE_LOCAL_EVENT_TYPE(wxEVT_GOCONTROL)
 
+namespace {
+enum { ID_PANEL_RESIZE_TIMER = wxID_HIGHEST + 1 };
+constexpr int RESIZE_SETTLE_MS = 150;
+} // namespace
+
 BEGIN_EVENT_TABLE(GOGUIPanelWidget, wxPanel)
 EVT_ERASE_BACKGROUND(GOGUIPanelWidget::OnErase)
 EVT_PAINT(GOGUIPanelWidget::OnPaint)
+EVT_TIMER(ID_PANEL_RESIZE_TIMER, GOGUIPanelWidget::OnResizeTimer)
 EVT_COMMAND(0, wxEVT_GOCONTROL, GOGUIPanelWidget::OnGOControl)
 EVT_MOTION(GOGUIPanelWidget::OnMouseMove)
 EVT_LEFT_DOWN(GOGUIPanelWidget::OnMouseLeftDown)
@@ -43,6 +49,7 @@ GOGUIPanelWidget::GOGUIPanelWidget(
     m_BGInit(false),
     m_Scale(1),
     m_FontScale(1),
+    m_ResizeTimer(this, ID_PANEL_RESIZE_TIMER),
     m_PressedPoint(default_point) {
   m_Background.SetSourceImage(&m_BGImage);
   initFont();
@@ -51,13 +58,14 @@ GOGUIPanelWidget::GOGUIPanelWidget(
     m_panel->GetWidth() * m_Scale, m_panel->GetHeight() * m_Scale);
   m_panel->PrepareDraw(m_Scale, NULL);
   OnUpdate();
+  m_LastFullBitmap = m_ClientBitmap;
   m_BGImage = m_ClientBitmap.ConvertToImage();
   m_Background.BuildScaledBitmap(m_Scale, wxRect(0, 0, 0, 0), NULL);
   m_BGInit = true;
   SetCanFocus(m_panel->IsKeyboardInputUsed());
 }
 
-GOGUIPanelWidget::~GOGUIPanelWidget() {}
+GOGUIPanelWidget::~GOGUIPanelWidget() { m_ResizeTimer.Stop(); }
 
 void GOGUIPanelWidget::initFont() {
   wxMemoryDC dc;
@@ -92,10 +100,43 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
                       // this. Without this limit, sizing too a too small value
                       // causes a crash!
     m_Scale = 0.25;
+
+  const int targetWidth = m_panel->GetWidth() * m_Scale + 0.5;
+  const int targetHeight = m_panel->GetHeight() * m_Scale + 0.5;
+
+  // Cheap immediate feedback for a live drag: rescale the last sharply
+  // redrawn bitmap instead of re-running PrepareDraw()/OnUpdate(), which
+  // rescale every key/stop/texture with BICUBIC quality and made dragging a
+  // panel border unusably slow when done on every single WM_SIZE tick. The
+  // sharp, full-quality redraw happens once in OnResizeTimer(), after
+  // resizing has been idle for a moment.
+  //
+  // Always rescale from m_LastFullBitmap, never from m_ClientBitmap: during
+  // a drag m_ClientBitmap is itself the result of this same preview rescale,
+  // so chaining off it would compound quality loss tick after tick instead
+  // of staying as sharp as a single rescale from the original allows.
+  //
+  // wxDC::StretchBlit was tried here first, but on wxMSW it maps to GDI
+  // StretchBlt in COLORONCOLOR (nearest-neighbour) mode, which duplicates or
+  // drops whole rows/columns at non-integer scale factors instead of
+  // interpolating - visibly blocky. wxImage::Scale() with NORMAL (bilinear)
+  // quality looks smooth and is still cheap here because it runs once on the
+  // already-composed panel bitmap, not once per control like BICUBIC mode
+  // does in OnUpdate()/PrepareDraw().
+  if (m_LastFullBitmap.IsOk() && targetWidth > 0 && targetHeight > 0)
+    m_ClientBitmap = (wxBitmap)m_LastFullBitmap.ConvertToImage().Scale(
+      targetWidth, targetHeight, wxIMAGE_QUALITY_NORMAL);
+  SetSize(targetWidth, targetHeight);
+  Refresh();
+  m_ResizeTimer.StartOnce(RESIZE_SETTLE_MS);
+  return GetSize();
+}
+
+void GOGUIPanelWidget::OnResizeTimer(wxTimerEvent &WXUNUSED(event)) {
   m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
   OnUpdate();
+  m_LastFullBitmap = m_ClientBitmap;
   Refresh();
-  return GetSize();
 }
 
 void GOGUIPanelWidget::OnDraw(wxDC *dc) {
--- a/src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
+++ b/src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
@@ -19,10 +19,8 @@
 
 DEFINE_LOCAL_EVENT_TYPE(wxEVT_GOCONTROL)
 
-namespace {
 enum { ID_PANEL_RESIZE_TIMER = wxID_HIGHEST + 1 };
 constexpr int RESIZE_SETTLE_MS = 150;
-} // namespace
 
 BEGIN_EVENT_TABLE(GOGUIPanelWidget, wxPanel)
 EVT_ERASE_BACKGROUND(GOGUIPanelWidget::OnErase)
@@ -47,7 +45,9 @@ GOGUIPanelWidget::GOGUIPanelWidget(
   : wxPanel(parent, id),
     m_panel(panel),
     m_BGInit(false),
+    m_ResizeInProgress(false),
     m_Scale(1),
+    m_LastRedrawScale(1),
     m_FontScale(1),
     m_ResizeTimer(this, ID_PANEL_RESIZE_TIMER),
     m_PressedPoint(default_point) {
@@ -58,7 +58,6 @@ GOGUIPanelWidget::GOGUIPanelWidget(
     m_panel->GetWidth() * m_Scale, m_panel->GetHeight() * m_Scale);
   m_panel->PrepareDraw(m_Scale, NULL);
   OnUpdate();
-  m_LastFullBitmap = m_ClientBitmap;
   m_BGImage = m_ClientBitmap.ConvertToImage();
   m_Background.BuildScaledBitmap(m_Scale, wxRect(0, 0, 0, 0), NULL);
   m_BGInit = true;
@@ -85,7 +84,7 @@ void GOGUIPanelWidget::Focus() {
     SetFocus();
 }
 
-wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
+void GOGUIPanelWidget::ComputeScale(const wxSize &size) {
   double scaleX = size.GetWidth() / (double)m_panel->GetWidth();
   double scaleY = size.GetHeight() / (double)m_panel->GetHeight();
   if (scaleX < scaleY) // To fit all, we need to use the SMALLEST of the two
@@ -100,9 +99,26 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
                       // this. Without this limit, sizing too a too small value
                       // causes a crash!
     m_Scale = 0.25;
+}
 
-  const int targetWidth = m_panel->GetWidth() * m_Scale + 0.5;
-  const int targetHeight = m_panel->GetHeight() * m_Scale + 0.5;
+wxSize GOGUIPanelWidget::GetScaledPanelSize() const {
+  return wxSize(
+    m_panel->GetWidth() * m_Scale + 0.5, m_panel->GetHeight() * m_Scale + 0.5);
+}
+
+wxSize GOGUIPanelWidget::SetInitialSize(const wxSize &size) {
+  ComputeScale(size);
+  m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
+  OnUpdate();
+  m_LastRedrawScale = m_Scale;
+  Refresh();
+  return GetSize();
+}
+
+wxSize GOGUIPanelWidget::UpdatePreviewSize(const wxSize &size) {
+  ComputeScale(size);
+
+  const wxSize scaledSize = GetScaledPanelSize();
 
   // Cheap immediate feedback for a live drag: rescale the last sharply
   // redrawn bitmap instead of re-running PrepareDraw()/OnUpdate(), which
@@ -111,8 +127,11 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
   // sharp, full-quality redraw happens once in OnResizeTimer(), after
   // resizing has been idle for a moment.
   //
-  // Always rescale from m_LastFullBitmap, never from m_ClientBitmap: during
-  // a drag m_ClientBitmap is itself the result of this same preview rescale,
+  // The snapshot is captured lazily at the start of each resize series so
+  // that it reflects whatever m_ClientBitmap shows right now - including
+  // single-control redraws (OnGOControl) done since the last full redraw.
+  // Always rescale from that snapshot, never from m_ClientBitmap: during a
+  // drag m_ClientBitmap is itself the result of this same preview rescale,
   // so chaining off it would compound quality loss tick after tick instead
   // of staying as sharp as a single rescale from the original allows.
   //
@@ -123,19 +142,37 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
   // quality looks smooth and is still cheap here because it runs once on the
   // already-composed panel bitmap, not once per control like BICUBIC mode
   // does in OnUpdate()/PrepareDraw().
-  if (m_LastFullBitmap.IsOk() && targetWidth > 0 && targetHeight > 0)
-    m_ClientBitmap = (wxBitmap)m_LastFullBitmap.ConvertToImage().Scale(
-      targetWidth, targetHeight, wxIMAGE_QUALITY_NORMAL);
-  SetSize(targetWidth, targetHeight);
+  if (!m_ResizeInProgress) {
+    m_ResizeInProgress = true;
+    m_PreResizeBitmap = m_ClientBitmap;
+  }
+  if (
+    m_PreResizeBitmap.IsOk() && scaledSize.GetWidth() > 0
+    && scaledSize.GetHeight() > 0)
+    m_ClientBitmap = (wxBitmap)m_PreResizeBitmap.ConvertToImage().Scale(
+      scaledSize.GetWidth(), scaledSize.GetHeight(), wxIMAGE_QUALITY_NORMAL);
+  SetSize(scaledSize.GetWidth(), scaledSize.GetHeight());
   Refresh();
   m_ResizeTimer.StartOnce(RESIZE_SETTLE_MS);
   return GetSize();
 }
 
 void GOGUIPanelWidget::OnResizeTimer(wxTimerEvent &WXUNUSED(event)) {
+  m_ResizeInProgress = false;
+
+  /* The series ended at the very scale that is already sharply rendered
+   * (the user dragged out and back, or size events delivered while a long
+   * redraw was running re-armed the timer). m_ClientBitmap already shows
+   * the 1:1 preview of the sharp bitmap plus any control redraws done on
+   * top of it, so there is nothing a full redraw could improve. Without
+   * this check a slow redraw ran up to three times back to back for
+   * identical output. */
+  if (m_Scale == m_LastRedrawScale)
+    return;
+
   m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
   OnUpdate();
-  m_LastFullBitmap = m_ClientBitmap;
+  m_LastRedrawScale = m_Scale;
   Refresh();
 }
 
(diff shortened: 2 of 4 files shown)


oleg68 (commented):
Pulled this locally, built it, and reviewed the debounce/preview mechanism in depth. The core idea (cache-and-rescale preview + settle timer) is solid and the perf win is real. I found two correctness issues and a few smaller cleanup items — left as inline suggestions below, all implemented and locally tested together:

**Correctness**
1. `OnGOControl()` mutates `m_ClientBitmap` for single-control redraws but never updates `m_LastFullBitmap`, so a control whose state changed since the last full redraw (a stop engaged, a piston pressed) visibly reverts to its old state in the live-drag preview until the next full redraw. Fixed by capturing the preview snapshot lazily at the start of each resize series (from whatever `m_ClientBitmap` currently holds) instead of eagerly maintaining a separate cache in three places.
2. Opening a panel always goes through the same debounced/preview path as a live resize tick, since `UpdateSize()` can't tell the two apart — so opening a panel that needs scaling shows a blurry frame before popping to the sharp render ~150ms later, and opening at native scale pays for a wasted duplicate full render. Fixed by giving one-off sizing its own method (`SetInitialSize()`) that skips the preview/debounce entirely.

**Minor / cleanup** (see inline comments): swapped the manual `Freeze()`/`Thaw()` pairing for `wxWindowUpdateLocker` (RAII), dropped the anonymous namespace around the timer constants (only such usage in `src/grandorgue`; `constexpr` already has internal linkage), and extracted the `GetWidth()*m_Scale+0.5`/`GetHeight()*m_Scale+0.5` formula (duplicated between `UpdateSize()`/`OnUpdate()`) into a shared `GetScaledPanelSize()` helper. Also renamed `UpdateSize()` → `UpdatePreviewSize()` to make its "cheap, provisional" nature explicit now that `SetInitialSize()` exists alongside it, and `m_LastFullBitmap` → `m_PreResizeBitmap` to reflect its new lazy-snapshot semantics.

**Noted but not changed** (see inline comments): the `wxImage` round-trip in the preview rescale can lose/mishandle alpha on some backends (worth a look if a skin with transparent edges ever looks wrong specifically during a live-drag preview, but correct once settled); the destructor's `m_ResizeTimer.Stop()` relies on wx discarding any already-queued timer event on all backends; this is now the third independent hand-rolled debounce-timer in the codebase (after `GOSplash` and `GOMidiEventRecvTab`) — left a `TODO` rather than generalizing it here, out of scope for this PR.

**A few pieces of the diff aren't part of this PR's diff and GitHub won't let me anchor an inline suggestion to them (they're pre-existing lines this PR doesn't touch, or lines outside any hunk's context), so here they are directly:**

In `GOGUIPanelWidget.h`, the public declarations of the two split-out sizing methods (replaces the old `wxSize UpdateSize(wxSize size);`):
```cpp
  /**
   * Sets the panel's size immediately, with a full sharp redraw and no
   * live-drag preview/debounce. Intended for one-off sizing (e.g. when a
   * panel window is first opened), as opposed to UpdatePreviewSize() which
   * is for live window resizing. Returns the actual size of the scaled
   * panel.
   */
  wxSize SetInitialSize(const wxSize &size);
  /**
   * Handles a live resize tick: computes the new scale, shows a cheap
   * bilinear preview immediately, and (re)starts the debounce timer that
   * will trigger a full sharp redraw once resizing settles (see
   * OnResizeTimer()). Returns the actual size of the scaled panel.
   */
  wxSize UpdatePreviewSize(const wxSize &size);
```

In `GOGUIPanelWidget.cpp`, the old `UpdateSize()` signature and the top of its scale-clamping code becomes the new `ComputeScale()`'s signature (the tail of this same function is covered by the inline suggestion further down):
```cpp
void GOGUIPanelWidget::ComputeScale(const wxSize &size) {
  double scaleX = size.GetWidth() / (double)m_panel->GetWidth();
  double scaleY = size.GetHeight() / (double)m_panel->GetHeight();
  if (scaleX < scaleY) // To fit all, we need to use the SMALLEST of the two
    m_Scale = scaleX;
  else
    m_Scale = scaleY;
  if (m_Scale > 2) // Actually useless, because max frame size is set to
                   // initial size
    m_Scale = 2;
  if (m_Scale < 0.25) // Let's not make the window content TOO small. There
                      // will be scrollbars if the user sizes to smaller than
```
(continues into the inline suggestion on `OnUpdate()`'s tail below)

And `OnUpdate()`'s body, using the new `GetScaledPanelSize()` helper instead of duplicating the scale formula:
```cpp
void GOGUIPanelWidget::OnUpdate() {
  const wxSize scaledSize = GetScaledPanelSize();

  if (m_BGInit)
    m_ClientBitmap = (wxBitmap)m_BGImage.Scale(
      scaledSize.GetWidth(), scaledSize.GetHeight(), wxIMAGE_QUALITY_BICUBIC);
  else
    m_ClientBitmap = wxBitmap(scaledSize.GetWidth(), scaledSize.GetHeight());
```

In `GOGUIPanelView.cpp`, add the include needed for `wxWindowUpdateLocker`:
```cpp
#include <wx/wupdlock.h>
```
and switch the constructor's one-off sizing call from `UpdateSize()` to `SetInitialSize()`:
```cpp
  wxSize scaledsize
    = m_panelwidget->SetInitialSize(topWindow->GetClientSize());
```

**CHANGELOG.md**: this PR doesn't add an entry. Since it fixes a user-visible issue (#2537), worth adding a line at the top of the unreleased section, matching the existing style, e.g.:
```
- Fixed slow/flickery panel resizing while dragging a panel window's border https://github.com/GrandOrgue/grandorgue/issues/2537
```
(Not left as an inline suggestion since this PR doesn't touch that file, so GitHub won't let me anchor a suggestion there.)

**Testing**: built locally (ASan + leak checks enabled) and manually verified — toggling a stop then live-dragging a panel border no longer reverts the stop's visual state during the drag; opening panels at both native and non-native sizes shows the sharp render immediately; live resize is still smooth with no flicker/jump on centering; clean shutdown, no crashes.

src/grandorgue/gui/panels/GOGUIPanelWidget.h
```diff
@@ -24,9 +25,29 @@ class GOGUIPanelWidget : public wxPanel {
   bool m_BGInit;
   GOBitmap m_Background;
   wxBitmap m_ClientBitmap;
+
+  /**
+   * The bitmap produced by the last full, sharp redraw (OnUpdate() right
+   * after PrepareDraw()). UpdateSize() always rescales from this one for its
+   * cheap live-drag preview rather than from m_ClientBitmap - which during a
+   * drag may itself already be a rescaled preview - to avoid repeatedly
+   * rescaling an already-rescaled (and thus progressively softer) image.
```
oleg68: Renamed to reflect the new lazy-snapshot semantics (see the comment below and the field rename further down).

```suggestion
   * A snapshot of m_ClientBitmap captured lazily at the start of a resize
   * series (see UpdatePreviewSize()). UpdatePreviewSize() rescales from this
   * one for its cheap live-drag preview rather than from m_ClientBitmap -
   * which during a drag may itself already be a rescaled preview - to avoid
   * repeatedly rescaling an already-rescaled (and thus progressively softer)
   * image. Captured lazily (rather than eagerly updated wherever
   * m_ClientBitmap changes) so it always reflects the latest state,
   * including partial redraws done by OnGOControl().
```

src/grandorgue/gui/panels/GOGUIPanelWidget.h
```diff
@@ -24,9 +25,29 @@ class GOGUIPanelWidget : public wxPanel {
...
+   */
+  wxBitmap m_LastFullBitmap;
```
oleg68: Renamed to reflect the new lazy-snapshot semantics (see the comment above and UpdatePreviewSize() below).

```suggestion
  wxBitmap m_PreResizeBitmap;
```

src/grandorgue/gui/panels/GOGUIPanelWidget.h
```diff
@@ -24,9 +25,29 @@ class GOGUIPanelWidget : public wxPanel {
...
   double m_FontScale;
 
+  /**
+   * Fires once resizing has been idle for a short while. UpdateSize() only
+   * does a cheap rescale of m_LastFullBitmap on every WM_SIZE tick (a live
+   * drag can fire dozens of those); the expensive, sharp re-render of every
+   * control bitmap only runs once here, after the user stops moving the
+   * border - otherwise every pixel of mouse movement triggered a full
+   * BICUBIC rescale of every key/stop/texture, making resizing unusably
+   * slow.
```
oleg68: Documents when the timer actually starts/stops and how many times it can fire per resize series.

```suggestion
   * Fires once resizing has been idle for a short while.
   * UpdatePreviewSize() only does a cheap rescale of m_PreResizeBitmap on
   * every WM_SIZE tick (a live drag can fire dozens of those); the
   * expensive, sharp re-render of every control bitmap only runs once here,
   * after the user stops moving the border - otherwise every pixel of mouse
   * movement triggered a full BICUBIC rescale of every key/stop/texture,
   * making resizing unusably slow.
   *
   * Started (or restarted, since it is one-shot) by every
   * UpdatePreviewSize() call, i.e. on every live-drag resize tick.
   * wxTimer::Start()/StartOnce() simply restarts an already-running timer
   * rather than queuing another firing, so in the common case of ticks
   * arriving faster than RESIZE_SETTLE_MS apart (a normal mouse drag), it
   * does not fire even once until the ticks stop - it fires exactly once,
   * RESIZE_SETTLE_MS after the last tick. Only if ticks are spaced out by
   * more than RESIZE_SETTLE_MS (a slow/stepped resize) can it fire more than
   * once per resize series. Stops itself once it fires, running
   * OnResizeTimer(); also explicitly stopped in the destructor so it cannot
   * fire after the widget is destroyed. SetInitialSize() never touches it -
   * the initial sizing path is fully synchronous and skips the timer
   * entirely.
   *
   * TODO: this is the third independent hand-rolled debounce-timer
   * implementation in the codebase (alongside GOSplash and
   * GOMidiEventRecvTab::m_Timer). If a fourth case appears, extract a shared
   * "settle timer" helper instead of writing a fourth copy.
```

src/grandorgue/gui/panels/GOGUIPanelWidget.h
```diff
@@ -24,9 +25,29 @@ class GOGUIPanelWidget : public wxPanel {
...
+   */
+  wxTimer m_ResizeTimer;
```
oleg68, line 90: Adds the private helpers used to split scale computation from the two redraw paths (see the .cpp comments). Exact placement within the private section doesn't matter - tacking them on after `m_ResizeTimer` here just happens to land in a spot GitHub can anchor a suggestion to.

```suggestion
  wxTimer m_ResizeTimer;

  /**
   * Computes m_Scale to fit the panel into the given size, clamped to the
   * [0.25, 2] range
   */
  void ComputeScale(const wxSize &size);
  /**
   * Returns the size m_ClientBitmap should have at the current m_Scale
   */
  wxSize GetScaledPanelSize() const;
  /**
   * Runs the expensive, sharp redraw of every control at the current
   * m_Scale: PrepareDraw()+OnUpdate(), then repaints. Used both for the
   * initial full-quality render and for the settled redraw after a resize
   * series ends (OnResizeTimer()).
   */
  void FullRedraw();
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -19,9 +19,15 @@
 
 DEFINE_LOCAL_EVENT_TYPE(wxEVT_GOCONTROL)
 
+namespace {
+enum { ID_PANEL_RESIZE_TIMER = wxID_HIGHEST + 1 };
+constexpr int RESIZE_SETTLE_MS = 150;
+} // namespace
```
oleg68: This is the only anonymous-namespace usage in `src/grandorgue` (everywhere else uses file-scope `static`), and it isn't needed here: `constexpr` already has internal linkage by default, and enumerators aren't linked symbols at all.

```suggestion
static constexpr int ID_PANEL_RESIZE_TIMER = wxID_HIGHEST + 1;
static constexpr unsigned RESIZE_SETTLE_MS = 150;
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -51,13 +58,14 @@ GOGUIPanelWidget::GOGUIPanelWidget(
     m_panel->GetWidth() * m_Scale, m_panel->GetHeight() * m_Scale);
   m_panel->PrepareDraw(m_Scale, NULL);
   OnUpdate();
+  m_LastFullBitmap = m_ClientBitmap;
```
oleg68: No longer needed - `m_PreResizeBitmap` (renamed from `m_LastFullBitmap`) is now captured lazily on the first `UpdatePreviewSize()` call of each resize series, from whatever `m_ClientBitmap` holds at that point (see below). This also fixes it staying in sync with `OnGOControl()`'s partial redraws, which never updated the old `m_LastFullBitmap`.

```suggestion
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -51,13 +58,14 @@ GOGUIPanelWidget::GOGUIPanelWidget(
...
   m_BGImage = m_ClientBitmap.ConvertToImage();
   m_Background.BuildScaledBitmap(m_Scale, wxRect(0, 0, 0, 0), NULL);
   m_BGInit = true;
   SetCanFocus(m_panel->IsKeyboardInputUsed());
 }
 
-GOGUIPanelWidget::~GOGUIPanelWidget() {}
+GOGUIPanelWidget::~GOGUIPanelWidget() { m_ResizeTimer.Stop(); }
```
oleg68: Documents the assumption this relies on (not verified against every wx backend's timer implementation - flagged in case it's ever worth checking).

```suggestion
GOGUIPanelWidget::~GOGUIPanelWidget() {
  /* Relies on wx discarding any wxTimerEvent already queued for this timer
  at the moment of Stop()/destruction, on all supported backends. */
  m_ResizeTimer.Stop();
}
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -92,10 +100,43 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
...
+  // already-composed panel bitmap, not once per control like BICUBIC mode
+  // does in OnUpdate()/PrepareDraw().
+  if (m_LastFullBitmap.IsOk() && targetWidth > 0 && targetHeight > 0)
+    m_ClientBitmap = (wxBitmap)m_LastFullBitmap.ConvertToImage().Scale(
+      targetWidth, targetHeight, wxIMAGE_QUALITY_NORMAL);
+  SetSize(targetWidth, targetHeight);
+  Refresh();
+  m_ResizeTimer.StartOnce(RESIZE_SETTLE_MS);
+  return GetSize();
+}
```
oleg68, line 166: **Correctness fix (stale control state) + one-off vs. live-drag split.** GitHub won't let me anchor a suggestion above this point since lines 88-99 (the old `UpdateSize()` signature and the top of the scale-clamping code) predate this PR and aren't part of its diff - see the general comment for the full before/after including that part. From here on:
- `ComputeScale()`/`GetScaledPanelSize()` extract logic that was inline here so it can be shared with the new `SetInitialSize()`/`FullRedraw()`.
- `SetInitialSize()` is a new, fully synchronous path (no preview, no timer) intended for one-off sizing - `GOGUIPanelView`'s constructor now calls this instead of the old `UpdateSize()`.
- `UpdatePreviewSize()` (renamed from `UpdateSize()`) now captures `m_PreResizeBitmap` **lazily**, only when `m_ResizeTimer` is not yet running, from whatever `m_ClientBitmap` currently holds - fixing the staleness bug where `OnGOControl()`'s partial redraws were never reflected in the old eagerly-maintained `m_LastFullBitmap`.

```suggestion
                      // this. Without this limit, sizing too a too small value
                      // causes a crash!
    m_Scale = 0.25;
}

wxSize GOGUIPanelWidget::GetScaledPanelSize() const {
  return wxSize(
    m_panel->GetWidth() * m_Scale + 0.5, m_panel->GetHeight() * m_Scale + 0.5);
}

void GOGUIPanelWidget::FullRedraw() {
  m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
  OnUpdate();
  Refresh();
}

wxSize GOGUIPanelWidget::SetInitialSize(const wxSize &size) {
  ComputeScale(size);
  FullRedraw();
  return GetSize();
}

wxSize GOGUIPanelWidget::UpdatePreviewSize(const wxSize &size) {
  ComputeScale(size);
  const wxSize targetSize = GetScaledPanelSize();

  /* Cheap immediate feedback for a live drag: rescale a snapshot of the last
  sharply redrawn bitmap instead of re-running PrepareDraw()/OnUpdate(),
  which rescale every key/stop/texture with BICUBIC quality and made
  dragging a panel border unusably slow when done on every single WM_SIZE
  tick. The sharp, full-quality redraw happens once in OnResizeTimer(),
  after resizing has been idle for a moment.

  The snapshot (m_PreResizeBitmap) is captured lazily here, only at the
  start of a resize series (when m_ResizeTimer is not yet running), from
  whatever m_ClientBitmap currently holds. Never rescale from
  m_ClientBitmap directly: during a drag it is itself the result of this
  same preview rescale, so chaining off it would compound quality loss
  tick after tick instead of staying as sharp as a single rescale from the
  pre-drag frame allows. */
  if (!m_ResizeTimer.IsRunning())
    m_PreResizeBitmap = m_ClientBitmap;

  /* wxDC::StretchBlit was tried here first, but on wxMSW it maps to GDI
  StretchBlt in COLORONCOLOR (nearest-neighbour) mode, which duplicates or
  drops whole rows/columns at non-integer scale factors instead of
  interpolating - visibly blocky. wxImage::Scale() with NORMAL (bilinear)
  quality looks smooth and is still cheap here because it runs once on the
  already-composed panel bitmap, not once per control like BICUBIC mode
  does in OnUpdate()/PrepareDraw().

  Note: converting through wxImage can lose/mishandle alpha masks on some
  wx backends; if a skin with transparent control edges ever looks wrong
  specifically during the live-drag preview (but correct once settled),
  this round-trip is the first place to check. */
  if (
    m_PreResizeBitmap.IsOk() && targetSize.GetWidth() > 0
    && targetSize.GetHeight() > 0)
    m_ClientBitmap = (wxBitmap)m_PreResizeBitmap.ConvertToImage().Scale(
      targetSize.GetWidth(), targetSize.GetHeight(), wxIMAGE_QUALITY_NORMAL);
  SetSize(targetSize);
  Refresh();
  m_ResizeTimer.StartOnce(RESIZE_SETTLE_MS);
  return GetSize();
}
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -92,10 +100,43 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
...
+
+void GOGUIPanelWidget::OnResizeTimer(wxTimerEvent &WXUNUSED(event)) {
   m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
   OnUpdate();
+  m_LastFullBitmap = m_ClientBitmap;
   Refresh();
-  return GetSize();
 }
```
oleg68, line 180: Now just calls the shared `FullRedraw()` helper (see above).

```suggestion
void GOGUIPanelWidget::OnResizeTimer(wxTimerEvent &WXUNUSED(event)) {
  /* Reaching here means resizing has either finished or is happening slowly
  enough (ticks more than RESIZE_SETTLE_MS apart) that a full sharp redraw
  is affordable now. */
  FullRedraw();
}
```

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -129,6 +129,16 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
...
     const wxSize newSize = event.GetSize();
+
+    // UpdateSize(), SetPosition() and CentreOnParent() below each move/resize
+    // m_panelwidget natively on their own. Without freezing, the window
+    // manager can paint the in-between states (e.g. still at the old
+    // centering position right after the size already changed), which is
+    // seen as a brief flicker/jump whenever the panel is dragged past its
+    // native size and centering kicks in. Freeze() suppresses repaints until
+    // all the geometry changes below are settled.
+    m_panelwidget->Freeze();
```
oleg68: Swapped the manual `Freeze()` for `wxWindowUpdateLocker` (RAII) - see the paired comment further down where the matching `Thaw()` is removed. Keeps the pair balanced even if an early return is added here later. Needs `#include <wx/wupdlock.h>` added near the top of the file (see the general comment - that line isn't part of this PR's diff so I can't anchor a suggestion to it).

```suggestion
    /* UpdatePreviewSize(), SetPosition() and CentreOnParent() below each
    move/resize m_panelwidget natively on their own. Without freezing, the
    window manager can paint the in-between states (e.g. still at the old
    centering position right after the size already changed), which is
    seen as a brief flicker/jump whenever the panel is dragged past its
    native size and centering kicks in. wxWindowUpdateLocker suppresses
    repaints (via Freeze()/Thaw()) until all the geometry changes below
    are settled, and keeps them balanced even if an early return is added
    here later. */
    wxWindowUpdateLocker locker(m_panelwidget);
```

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -129,6 +129,16 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
...
+
     const wxSize maxSize = m_panelwidget->UpdateSize(newSize);
```
oleg68: Renamed call site, matching the rename below.

```suggestion
    const wxSize maxSize = m_panelwidget->UpdatePreviewSize(newSize);
```

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -156,6 +166,8 @@ void GOGUIPanelView::OnSize(wxSizeEvent &event) {
       m_panelwidget->CentreOnParent(wxHORIZONTAL);
     if (actualsize.GetHeight() > maxSize.GetHeight())
       m_panelwidget->CentreOnParent(wxVERTICAL);
+
+    m_panelwidget->Thaw();
```
oleg68: No longer needed - `wxWindowUpdateLocker` (introduced above) calls `Thaw()` automatically when it goes out of scope at the end of the `if` block.

```suggestion
```

oleg68:
@sebion7125 Are you going to change this PR along my review?

sebion7125:
@oleg68 Thank you for the thorough review - all points are applied in a953b41:

- **Lazy preview snapshot**: `m_PreResizeBitmap` is now captured at the start of each resize series from whatever `m_ClientBitmap` currently shows, so single-control redraws done since the last full redraw stay visible in the preview. The three eager maintenance sites are gone.
- **`SetInitialSize()` / `UpdatePreviewSize()` split** as suggested, including your doc comments; the panel-open path no longer shows a blurry frame or pays a duplicate full render.
- **Cleanups**: `wxWindowUpdateLocker`, `ComputeScale()` + `GetScaledPanelSize()`, anonymous namespace dropped, CHANGELOG entry added.

One addition on top of the review, in the same commit: `OnResizeTimer()` now skips the full redraw when the series ends at the scale that is already sharply rendered (`m_LastRedrawScale`). While profiling a complex panel I saw size events delivered *during* a slow redraw re-arm the timer, producing up to three identical back-to-back redraws; `m_ClientBitmap` at that point already shows the 1:1 preview of the sharp bitmap plus any control redraws on top, so there is nothing a redraw could improve. Happy to split it out if you prefer it as a separate commit.

Your "noted but not changed" points (preview alpha round-trip, timer-stop semantics, third hand-rolled debounce) all make sense to me as notes - left as they are.

Status: builds cleanly on Linux/GTK3 (no warnings, clang-format clean). The review items match the implementation you already tested locally; the only functional addition beyond that is the same-scale skip described above.

src/grandorgue/gui/panels/GOGUIPanelWidget.h
```diff
@@ -24,20 +25,69 @@ class GOGUIPanelWidget : public wxPanel {
...
    * A point where the mouse has been pressed. Used for deduplication
    */
   wxPoint m_PressedPoint;
 
   void initFont();
+
+  /** Computes and stores m_Scale so that the panel fits into `size`. */
+  void ComputeScale(const wxSize &size);
+  /** The panel's pixel size at the current m_Scale. */
+  wxSize GetScaledPanelSize() const;
```
oleg68, line 102: `SetInitialSize()` and `OnResizeTimer()` both duplicate the same `PrepareDraw()`/`OnUpdate()`/`Refresh()` sequence. Worth extracting into a shared helper (as originally suggested in https://github.com/GrandOrgue/grandorgue/pull/2539#discussion_r3521758520, which only got the correctness-fix part applied):

```suggestion
  /** The panel's pixel size at the current m_Scale. */
  wxSize GetScaledPanelSize() const;
  /**
   * Runs the expensive, sharp redraw of every control at the current
   * m_Scale: PrepareDraw()+OnUpdate(), then repaints. Used both for the
   * initial full-quality render and for the settled redraw after a resize
   * series ends (OnResizeTimer()).
   */
  void FullRedraw();
```

src/grandorgue/gui/panels/GOGUIPanelWidget.cpp
```diff
@@ -92,12 +99,83 @@ wxSize GOGUIPanelWidget::UpdateSize(wxSize size) {
...
+   * this check a slow redraw ran up to three times back to back for
+   * identical output. */
+  if (m_Scale == m_LastRedrawScale)
+    return;
+
+  m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
+  OnUpdate();
+  m_LastRedrawScale = m_Scale;
+  Refresh();
+}
```
oleg68: `SetInitialSize()` and `OnResizeTimer()` both duplicate the same `PrepareDraw()`/`OnUpdate()`/`Refresh()`/`m_LastRedrawScale` sequence. Extract into a shared `FullRedraw()` helper (declaration suggested separately on `GOGUIPanelWidget.h`) so there's a single place to change if this sequence ever needs to change again.

```suggestion
void GOGUIPanelWidget::FullRedraw() {
  m_panel->PrepareDraw(m_Scale, m_BGInit ? &m_Background : NULL);
  OnUpdate();
  m_LastRedrawScale = m_Scale;
  Refresh();
}

wxSize GOGUIPanelWidget::SetInitialSize(const wxSize &size) {
  ComputeScale(size);
  FullRedraw();
  return GetSize();
}

wxSize GOGUIPanelWidget::UpdatePreviewSize(const wxSize &size) {
  ComputeScale(size);

  const wxSize scaledSize = GetScaledPanelSize();

  // Cheap immediate feedback for a live drag: rescale the last sharply
  // redrawn bitmap instead of re-running PrepareDraw()/OnUpdate(), which
  // rescale every key/stop/texture with BICUBIC quality and made dragging a
  // panel border unusably slow when done on every single WM_SIZE tick. The
  // sharp, full-quality redraw happens once in OnResizeTimer(), after
  // resizing has been idle for a moment.
  //
  // The snapshot is captured lazily at the start of each resize series so
  // that it reflects whatever m_ClientBitmap shows right now - including
  // single-control redraws (OnGOControl) done since the last full redraw.
  // Always rescale from that snapshot, never from m_ClientBitmap: during a
  // drag m_ClientBitmap is itself the result of this same preview rescale,
  // so chaining off it would compound quality loss tick after tick instead
  // of staying as sharp as a single rescale from the original allows.
  //
  // wxDC::StretchBlit was tried here first, but on wxMSW it maps to GDI
  // StretchBlt in COLORONCOLOR (nearest-neighbour) mode, which duplicates or
  // drops whole rows/columns at non-integer scale factors instead of
  // interpolating - visibly blocky. wxImage::Scale() with NORMAL (bilinear)
  // quality looks smooth and is still cheap here because it runs once on the
  // already-composed panel bitmap, not once per control like BICUBIC mode
  // does in OnUpdate()/PrepareDraw().
  if (!m_ResizeInProgress) {
    m_ResizeInProgress = true;
    m_PreResizeBitmap = m_ClientBitmap;
  }
  if (
    m_PreResizeBitmap.IsOk() && scaledSize.GetWidth() > 0
    && scaledSize.GetHeight() > 0)
    m_ClientBitmap = (wxBitmap)m_PreResizeBitmap.ConvertToImage().Scale(
      scaledSize.GetWidth(), scaledSize.GetHeight(), wxIMAGE_QUALITY_NORMAL);
  SetSize(scaledSize.GetWidth(), scaledSize.GetHeight());
  Refresh();
  m_ResizeTimer.StartOnce(RESIZE_SETTLE_MS);
  return GetSize();
}

void GOGUIPanelWidget::OnResizeTimer(wxTimerEvent &WXUNUSED(event)) {
  m_ResizeInProgress = false;

  /* The series ended at the very scale that is already sharply rendered
   * (the user dragged out and back, or size events delivered while a long
   * redraw was running re-armed the timer). m_ClientBitmap already shows
   * the 1:1 preview of the sharp bitmap plus any control redraws done on
   * top of it, so there is nothing a full redraw could improve. Without
   * this check a slow redraw ran up to three times back to back for
   * identical output. */
  if (m_Scale == m_LastRedrawScale)
    return;

  FullRedraw();
}
```

sebion7125: Done in aff9b139 - `FullRedraw()` extracted exactly as suggested (it also owns the `m_LastRedrawScale` update, so the skip logic and the redraw stay in one place). Rebased on your master merge; builds cleanly.

oleg68:
> @oleg68 Thank you for the thorough review - all points are applied

@sebion7125 Thank you for your changes. But there are some notes left. Could you apply them?
- Your code duplicates the same sequence in SetInitialSize() and OnResizeTimer(). Could you get rid the duplication and extract the common code in a new helper `FullRedraw` as suggested above?
-  GO code does not use ``namespace {}``. Could you replace it with `static` as suggested above?

> One addition on top of the review, in the same commit: `OnResizeTimer()` now skips the full redraw when the series ends at the scale that is already sharply rendered (`m_LastRedrawScale`). While profiling a complex panel I saw size events delivered _during_ a slow redraw re-arm the timer, producing up to three identical back-to-back redraws; `m_ClientBitmap` at that point already shows the 1:1 preview of the sharp bitmap plus any control redraws on top, so there is nothing a redraw could improve. Happy to split it out if you prefer it as a separate commit.

This change is good here, but any changed code of GrandOrgue **must not** contain early returns. So could your replace "if (condition) return;" with "if (!condition) { the rest code; }"?

sebion7125:
@oleg68 All three applied:

- `FullRedraw()` was extracted in aff9b139 (our comments crossed - it also owns the `m_LastRedrawScale` update so the skip logic and the redraw stay in one place)
- f8974a0d replaces the enum/`constexpr` pair with `static constexpr` exactly as in your suggestion, and
- inverts the skip check in `OnResizeTimer()` so the changed code contains no early return.

I checked the rest of the diff for further early returns - the remaining `return` statements are either pre-existing code outside this PR's changes or value returns at the end of a function. Builds cleanly, clang-format clean.

Good to know about both conventions (no `namespace {}`, no early returns in changed code) - I'll follow them in upcoming PRs.

oleg68 (approved):

oleg68:
Thank you, @sebion7125 , for quick fixes. I've extended the comment text and added another comment. Now this PR is ready for merge

oleg68:

rousseldenis (approved):
Tested. Great !

PR #2540, opened 2026-06-30
Fixed crash when loading PNG images with embedded alpha channel used with mask images

### LINKED ISSUE #2535: GOGuiImageCache: InitAlpha() called unconditionally may crash on PNG images with embedded alpha channel
## Description

In `src/grandorgue/gui/GOGuiImageCache.cpp`, `wxImage::InitAlpha()` is called without checking `HasAlpha()` first:

```cpp
// GOGuiImageCache.cpp ~line 263
if (pNewImage->HasMask())
    pNewImage->InitAlpha();   // ← crashes if image already has an alpha channel
```

If a PNG image file has an embedded alpha channel **and** a mask is applied via `SetMaskFromImage()`, the call to `InitAlpha()` triggers a wxWidgets assertion failure:

```
image.cpp(2218) assert HasAlpha() failed in initAlpha(): image already has an alpha channel
```

In **debug builds** this shows an assert dialog and then continues with a corrupted image state. In **release builds** the assert is silently skipped but `InitAlpha()` re-allocates the alpha buffer, losing the original alpha data → the image is corrupted → crash when the panel tries to render it.

## How to reproduce

Load a sample set whose ODF panel references a PNG bitmap that has both an embedded alpha channel and is used with a mask image. GrandOrgue will crash after the cache is built, when the panel windows are being opened.

## Fix

`GOBitmap.cpp` already does this correctly — add the same `HasAlpha()` guard here:

```cpp
// Before:
if (pNewImage->HasMask())
    pNewImage->InitAlpha();

// After:
if (pNewImage->HasMask() && !pNewImage->HasAlpha())
    pNewImage->InitAlpha();
```

This matches the existing pattern in `src/grandorgue/gui/panels/primitives/GOBitmap.cpp` lines 46–47.

## Environment

- GrandOrgue: current master branch (confirmed present as of 2026-06-30)
- wxWidgets: reproducible on at least one Windows build

oleg68:
Fixes #2535

## Summary

- `GOGuiImageCache::LoadImage()` called `wxImage::InitAlpha()` unconditionally when `HasMask()` was true, without checking `HasAlpha()` first. If a PNG image already had an embedded alpha channel and a mask was applied via `SetMaskFromImage()`, this triggered a wxWidgets assertion in debug builds and corrupted the alpha buffer in release builds, leading to a crash during panel rendering.
- When the image already has an alpha channel, `InitAlpha()` must not be called. Instead, the mask is applied manually: pixels matching the mask color are set to `alpha=0`; all other pixels keep their existing alpha value. Relying on the wxImage mask (chroma-key) alone is not sufficient because `GOBitmap::BuildBitmapFrom()` copies `img.GetAlpha()` directly into the compositing result and never consults the mask.

## Test plan

- [ ] Load an organ with an ODF image that has native PNG alpha and a `Mask...` file
- [ ] Confirm no assertion dialog (debug build) and no crash (release build)
- [ ] Confirm the masked region renders correctly (transparency preserved)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed crash when loading PNG images with embedded alpha channel used with mask images https://github.com/GrandOrgue/grandorgue/issues/2535
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
 - Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
--- a/src/grandorgue/gui/GOGuiImageCache.cpp
+++ b/src/grandorgue/gui/GOGuiImageCache.cpp
@@ -248,8 +248,46 @@ const wxImage *GOGuiImageCache::LoadImage(
 
     wxImage *pNewImage = new wxImage(image);
 
-    if (pNewImage->HasMask())
-      pNewImage->InitAlpha();
+    if (pNewImage->HasMask()) {
+      if (!pNewImage->HasAlpha()) {
+        /* No alpha channel yet: convert mask to alpha.
+          InitAlpha() sets mask-colored pixels to alpha=0, all others to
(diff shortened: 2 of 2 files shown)


oleg68:

rousseldenis (approved):

PR #2519, opened 2026-06-08
Print stacktrace and cpptrace integration for debug builds

oleg68:
## Problem statement

After making intermediate builds as debug builds some user started complaining thet GrandOrgue terminated on assertion failure. 

When the crash was not reproducible, getting stacktrace from the user was difficult. This PR enables printing stacktrace on assertion failure.

## Summary

- Override `GOGuiApp::OnAssertFailure` to print assertion info + a full cpptrace stacktrace to `stderr` before delegating to the standard wxWidgets handler. This makes debug builds survivable when non-fatal wx assertions fire (e.g. `CheckExpectedParentIs` on Ubuntu 24.04) and provides a stacktrace to find the root cause.
- Integrate [cpptrace](https://github.com/jeremy-rifkin/cpptrace) as a git submodule (`submodules/cpptrace`, tag v1.0.4) instead of a downloaded/installed external package.
- `USE_INTERNAL_CPPTRACE=OFF` allows building against a system-installed cpptrace.
- cpptrace is linked conditionally (`if(TARGET cpptrace::cpptrace)`) so the build degrades gracefully on platforms without it.
- Windows cross-compilation uses `dbghelp` (Windows system library) instead of `libdwarf` to avoid FetchContent cross-compilation issues with MinGW.

## Test plan
- [ ] Linux: `cmake .. && make -k -j16` builds cleanly; `GO_HAS_CPPTRACE` is defined; `OnAssertFailure` prints a stack trace on debug assertion.
- [ ] Windows (cross-compile): `dbghelp` backend used; no PE binary corruption.
- [ ] `USE_INTERNAL_CPPTRACE=OFF` (with system cpptrace installed): builds and links against system library.
- [ ] CI: all existing tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2542, opened 2026-07-02
fix: keyboard shortcuts work in detached organ panel windows

### LINKED ISSUE #2541: Keyboard shortcuts don't work when a detached organ panel has focus
## Problem

When an organ panel is opened as a detached floating window (all panels except the one embedded in the main application window), keyboard shortcuts stop working:

- **Escape** (Panic / stop all sound)
- **Ctrl+S** (Save)
- **Ctrl+L / Ctrl+O** (Load/Open organ)
- **Ctrl+I** (Install organ package)
- **Ctrl+P** (Load MIDI)
- **F1** (Help)

The reason is that wxWidgets' accelerator processing is scoped to the top-level window that currently has focus. Detached panel frames have no menu bar and no accelerator table, so key presses are silently dropped.

## Expected behaviour

All keyboard shortcuts defined in the main menu should work regardless of which panel (main or detached) currently has keyboard focus.

## Steps to reproduce

1. Open GrandOrgue with an organ that has multiple panels.
2. Click on a detached panel window so it gets focus.
3. Press **Escape** or **Ctrl+S** — neither triggers the expected action.

## Environment

- GrandOrgue (current master)
- Linux / Windows / macOS (all platforms affected)

sebion7125:
Fixes #2541

## Problem

When a detached organ panel frame has keyboard focus, shortcuts like
**Escape** (Panic), **Ctrl+S**, **F1** etc. do nothing. The root
cause: wxWidgets accelerator processing is scoped to the focused
top-level window, and detached panel frames carry no menu bar.

### `GOGUIPanelView` — forward key events to the main window

Each detached panel frame gets an `EVT_CHAR_HOOK` binding that
forwards every key event to the main window's event handler:

```cpp
topWindow->Bind(wxEVT_CHAR_HOOK, [mainWin](wxKeyEvent &e) {
    wxKeyEvent copy(e);
    copy.SetEventObject(mainWin);
    if (!mainWin->GetEventHandler()->ProcessEvent(copy))
        e.Skip();
});
```

`ProcessEvent()` returns `true` when `OnKeyCommand` consumed the
event (recognized shortcut). In that case `e.Skip()` is suppressed so
the organ panel does not simultaneously play a note for the same key.
For unrecognized keys (organ notes, Shift for setter) `Skip()` is
called and `GOGUIPanelWidget` handles them as before.

### `GOAppWindow` — single shortcut table in `OnKeyCommand`

The existing `OnKeyCommand` only handled Escape. Menu shortcuts (Ctrl+S
etc.) were handled by the OS/wxWidgets accelerator mechanism — which is
not reachable via `ProcessEvent`. Therefore `OnKeyCommand` is extended
to handle **all** shortcuts through a single `s_shortcuts[]` table:

```cpp
const ShortcutEntry s_shortcuts[] = {
  {wxACCEL_NORMAL, WXK_ESCAPE, ID_AUDIO_PANIC},
  {wxACCEL_NORMAL, WXK_F1,     wxID_HELP},
  {wxACCEL_CTRL,   'L',        ID_FILE_LOAD},
  {wxACCEL_CTRL,   'O',        ID_FILE_OPEN},
  {wxACCEL_CTRL,   'S',        ID_FILE_SAVE},
  {wxACCEL_CTRL,   'I',        ID_FILE_INSTALL},
  {wxACCEL_CTRL,   'P',        ID_MIDI_LOAD},
};
```

`OnKeyCommand` iterates the table and calls `ProcessCommand(id)`,
returning **without** `Skip()` on a match. This prevents the menu
system from double-firing. For unrecognized keys `Skip()` is called as
before.

**Adding a new shortcut** requires exactly one line in `s_shortcuts[]`
plus the `\tXxx` annotation in the menu label. Nothing else.

## Shortcuts covered

| Key | Action |
|-----|--------|
| Escape | Panic (stop all sound) |
| F1 | Help |
| Ctrl+L | Load organ |
| Ctrl+O | Open organ |
| Ctrl+S | Save |
| Ctrl+I | Install organ package |
| Ctrl+P | Load MIDI |

## Test plan

- [ ] Open an organ with multiple panels; detach at least one panel
- [ ] Click the detached panel to give it focus
- [ ] Press **Escape** → Panic fires
- [ ] Press **Ctrl+S** → Save occurs
- [ ] Press **F1** → Help opens
- [ ] Press **Ctrl+L / Ctrl+O** → Load/Open dialog appears
- [ ] Press **Ctrl+I** → Install dialog appears
- [ ] Press **Ctrl+P** → Load MIDI dialog appears
- [ ] Press organ keys (letters) → notes still play normally
- [ ] Press **Shift** → setter (Memory Set) still activates
- [ ] Verify all shortcuts still work when the main window has focus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Fixed keyboard shortcuts (Panic, Help, Load/Open/Save/Install organ, MIDI player load) not working when a detached organ panel window has focus
+- Fixed keyboard shortcuts (Panic, Help, Load/Open/Save/Install organ, MIDI player load) not working when a detached organ panel window has focus https://github.com/GrandOrgue/grandorgue/issues/2541
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
 - Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed keyboard shortcuts (Panic, Help, Load/Open/Save/Install organ, MIDI player load) not working when a detached organ panel window has focus
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
 - Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
--- a/src/grandorgue/gui/frames/GOAppWindow.cpp
+++ b/src/grandorgue/gui/frames/GOAppWindow.cpp
@@ -55,7 +55,6 @@
 #include "go_limits.h"
 #include "go_path.h"
 
-namespace {
 struct ShortcutEntry {
   int mod, key, id;
 };
@@ -63,7 +62,7 @@ struct ShortcutEntry {
 // To add a shortcut: append one line here. The menu label's \tXxx annotation
 // handles display; this table handles dispatch (for both GOAppWindow focus and
 // forwarded events from detached panel frames).
-const ShortcutEntry s_shortcuts[] = {
+static const ShortcutEntry s_shortcuts[] = {
   {wxACCEL_NORMAL, WXK_ESCAPE, ID_AUDIO_PANIC},
   {wxACCEL_NORMAL, WXK_F1,     wxID_HELP},
   {wxACCEL_CTRL,   'L',        ID_FILE_LOAD},
@@ -72,7 +71,6 @@ const ShortcutEntry s_shortcuts[] = {
   {wxACCEL_CTRL,   'I',        ID_FILE_INSTALL},
   {wxACCEL_CTRL,   'P',        ID_MIDI_LOAD},
 };
-} // namespace
 
 BEGIN_EVENT_TABLE(GOAppWindow, wxFrame)
 EVT_MSGBOX(GOAppWindow::OnMsgBox)
@@ -1311,20 +1309,26 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
 
 void GOAppWindow::OnKeyCommand(wxKeyEvent &event) {
   int flags = wxACCEL_NORMAL;
+
   if (event.ControlDown())
     flags |= wxACCEL_CTRL;
   if (event.AltDown())
     flags |= wxACCEL_ALT;
   if (event.ShiftDown())
     flags |= wxACCEL_SHIFT;
+
   const int k = event.GetKeyCode();
+  bool isProcessed = false;
+
   for (const auto &s : s_shortcuts) {
     if (s.mod == flags && s.key == k) {
       ProcessCommand(s.id);
-      return;
+      isProcessed = true;
+      break;
     }
   }
-  event.Skip();
+  if (!isProcessed)
+    event.Skip();
 }
 
 void GOAppWindow::OnMidiEvent(const GOMidiEvent &event) {
--- a/src/grandorgue/gui/frames/GOAppWindow.cpp
+++ b/src/grandorgue/gui/frames/GOAppWindow.cpp
@@ -55,6 +55,25 @@
 #include "go_limits.h"
 #include "go_path.h"
 
+namespace {
+struct ShortcutEntry {
+  int mod, key, id;
+};
+// Single source of truth for all keyboard shortcuts.
+// To add a shortcut: append one line here. The menu label's \tXxx annotation
+// handles display; this table handles dispatch (for both GOAppWindow focus and
+// forwarded events from detached panel frames).
+const ShortcutEntry s_shortcuts[] = {
+  {wxACCEL_NORMAL, WXK_ESCAPE, ID_AUDIO_PANIC},
+  {wxACCEL_NORMAL, WXK_F1,     wxID_HELP},
+  {wxACCEL_CTRL,   'L',        ID_FILE_LOAD},
+  {wxACCEL_CTRL,   'O',        ID_FILE_OPEN},
+  {wxACCEL_CTRL,   'S',        ID_FILE_SAVE},
+  {wxACCEL_CTRL,   'I',        ID_FILE_INSTALL},
+  {wxACCEL_CTRL,   'P',        ID_MIDI_LOAD},
+};
+} // namespace
+
 BEGIN_EVENT_TABLE(GOAppWindow, wxFrame)
 EVT_MSGBOX(GOAppWindow::OnMsgBox)
 EVT_RENAMEFILE(GOAppWindow::OnRenameFile)
@@ -1291,13 +1310,18 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
 }
 
 void GOAppWindow::OnKeyCommand(wxKeyEvent &event) {
-  int k = event.GetKeyCode();
-  if (!event.AltDown()) {
-    switch (k) {
-    case WXK_ESCAPE: {
-      ProcessCommand(ID_AUDIO_PANIC);
-      break;
-    }
+  int flags = wxACCEL_NORMAL;
+  if (event.ControlDown())
+    flags |= wxACCEL_CTRL;
+  if (event.AltDown())
+    flags |= wxACCEL_ALT;
+  if (event.ShiftDown())
+    flags |= wxACCEL_SHIFT;
+  const int k = event.GetKeyCode();
+  for (const auto &s : s_shortcuts) {
+    if (s.mod == flags && s.key == k) {
+      ProcessCommand(s.id);
+      return;
     }
   }
   event.Skip();
--- a/src/grandorgue/gui/frames/GOAppWindow.cpp
+++ b/src/grandorgue/gui/frames/GOAppWindow.cpp
@@ -64,12 +64,12 @@ struct ShortcutEntry {
 // forwarded events from detached panel frames).
 static const ShortcutEntry s_shortcuts[] = {
   {wxACCEL_NORMAL, WXK_ESCAPE, ID_AUDIO_PANIC},
-  {wxACCEL_NORMAL, WXK_F1,     wxID_HELP},
-  {wxACCEL_CTRL,   'L',        ID_FILE_LOAD},
-  {wxACCEL_CTRL,   'O',        ID_FILE_OPEN},
-  {wxACCEL_CTRL,   'S',        ID_FILE_SAVE},
-  {wxACCEL_CTRL,   'I',        ID_FILE_INSTALL},
-  {wxACCEL_CTRL,   'P',        ID_MIDI_LOAD},
+  {wxACCEL_NORMAL, WXK_F1, wxID_HELP},
+  {wxACCEL_CTRL, 'L', ID_FILE_LOAD},
+  {wxACCEL_CTRL, 'O', ID_FILE_OPEN},
+  {wxACCEL_CTRL, 'S', ID_FILE_SAVE},
+  {wxACCEL_CTRL, 'I', ID_FILE_INSTALL},
+  {wxACCEL_CTRL, 'P', ID_MIDI_LOAD},
 };
 
 BEGIN_EVENT_TABLE(GOAppWindow, wxFrame)
--- a/src/grandorgue/gui/panels/GOGUIPanelView.cpp
+++ b/src/grandorgue/gui/panels/GOGUIPanelView.cpp
@@ -99,12 +99,14 @@ GOGUIPanelView::GOGUIPanelView(
   // called and GOGUIPanelWidget handles them normally.
   // Panels embedded directly inside GOAppWindow are skipped because
   // GOAppWindow's own EVT_CHAR_HOOK already covers them.
-  wxWindow *mainWin = wxTheApp->GetTopWindow();
-  if (mainWin && topWindow != mainWin) {
-    topWindow->Bind(wxEVT_CHAR_HOOK, [mainWin](wxKeyEvent &e) {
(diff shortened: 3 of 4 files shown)


oleg68:
@sebion7125 could you add a changelog entry related to this fix?

oleg68:
@sebion7125 Could you install pre-commit and check ``pre-commit run -a``?

src/grandorgue/gui/frames/GOAppWindow.cpp
```diff
@@ -55,6 +55,25 @@
 #include "go_limits.h"
 #include "go_path.h"
 
+namespace {
```
oleg68: Using namespaces contradicts the GrandOrgue code style. Use `static` instead.

src/grandorgue/gui/frames/GOAppWindow.cpp
```diff
@@ -1291,13 +1310,18 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
...
 void GOAppWindow::OnKeyCommand(wxKeyEvent &event) {
-  int k = event.GetKeyCode();
-  if (!event.AltDown()) {
-    switch (k) {
-    case WXK_ESCAPE: {
-      ProcessCommand(ID_AUDIO_PANIC);
-      break;
-    }
+  int flags = wxACCEL_NORMAL;
+  if (event.ControlDown())
```
oleg68, line 1355: Please enter a empty line between variable declaration and other code

src/grandorgue/gui/frames/GOAppWindow.cpp
```diff
@@ -1291,13 +1310,18 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
...
+    flags |= wxACCEL_CTRL;
+  if (event.AltDown())
+    flags |= wxACCEL_ALT;
+  if (event.ShiftDown())
+    flags |= wxACCEL_SHIFT;
+  const int k = event.GetKeyCode();
```
oleg68, line 1362: Please enter a empty line between variable declaration and other code

src/grandorgue/gui/frames/GOAppWindow.cpp
```diff
@@ -1291,13 +1310,18 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
...
+  for (const auto &s : s_shortcuts) {
```
oleg68, line 1365: Please enter a empty line between variable declaration and other code

src/grandorgue/gui/frames/GOAppWindow.cpp
```diff
@@ -1291,13 +1310,18 @@ void GOAppWindow::OnChangeVolume(wxCommandEvent &event) {
...
+    if (s.mod == flags && s.key == k) {
+      ProcessCommand(s.id);
+      return;
```
oleg68: Please avoid early returns. Use break with setting a bool value that the event has been processed

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,25 @@ GOGUIPanelView::GOGUIPanelView(
...
+  // keyboard shortcuts (e.g. Escape/Panic, Ctrl+S, F2-F12) work regardless of
+  // which panel has focus. GOAppWindow's OnKeyCommand returns without Skip()
+  // for recognized shortcuts, causing ProcessEvent() to return true; in that
+  // case we suppress e.Skip() so the panel widget does not also process the key
+  // (e.g. organ note). For unrecognized keys (organ notes, Shift) Skip() is
+  // called and GOGUIPanelWidget handles them normally.
+  // Panels embedded directly inside GOAppWindow are skipped because
+  // GOAppWindow's own EVT_CHAR_HOOK already covers them.
+  wxWindow *mainWin = wxTheApp->GetTopWindow();
+  if (mainWin && topWindow != mainWin) {
```
oleg68: Please enter a empty line between variable declaration and other code

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,25 @@ GOGUIPanelView::GOGUIPanelView(
...
+    topWindow->Bind(wxEVT_CHAR_HOOK, [mainWin](wxKeyEvent &e) {
+      wxKeyEvent copy(e);
+      copy.SetEventObject(mainWin);
```
oleg68: Please enter a empty line between variable declaration and other code

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,25 @@ GOGUIPanelView::GOGUIPanelView(
...
+  // for recognized shortcuts, causing ProcessEvent() to return true; in that
+  // case we suppress e.Skip() so the panel widget does not also process the key
+  // (e.g. organ note). For unrecognized keys (organ notes, Shift) Skip() is
+  // called and GOGUIPanelWidget handles them normally.
+  // Panels embedded directly inside GOAppWindow are skipped because
+  // GOAppWindow's own EVT_CHAR_HOOK already covers them.
+  wxWindow *mainWin = wxTheApp->GetTopWindow();
+  if (mainWin && topWindow != mainWin) {
+    topWindow->Bind(wxEVT_CHAR_HOOK, [mainWin](wxKeyEvent &e) {
+      wxKeyEvent copy(e);
```
oleg68:
```suggestion
      wxKeyEvent eCopied(e);
```

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,25 @@ GOGUIPanelView::GOGUIPanelView(
...
+  // Forward key events from detached panel frames to the main window so that
+  // keyboard shortcuts (e.g. Escape/Panic, Ctrl+S, F2-F12) work regardless of
+  // which panel has focus. GOAppWindow's OnKeyCommand returns without Skip()
+  // for recognized shortcuts, causing ProcessEvent() to return true; in that
+  // case we suppress e.Skip() so the panel widget does not also process the key
+  // (e.g. organ note). For unrecognized keys (organ notes, Shift) Skip() is
+  // called and GOGUIPanelWidget handles them normally.
+  // Panels embedded directly inside GOAppWindow are skipped because
+  // GOAppWindow's own EVT_CHAR_HOOK already covers them.
+  wxWindow *mainWin = wxTheApp->GetTopWindow();
```
oleg68:
```suggestion
  wxWindow *pMainAppWin = wxTheApp->GetTopWindow();
```

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,25 @@ GOGUIPanelView::GOGUIPanelView(
...
+  if (mainWin && topWindow != mainWin) {
+    topWindow->Bind(wxEVT_CHAR_HOOK, [mainWin](wxKeyEvent &e) {
+      wxKeyEvent copy(e);
+      copy.SetEventObject(mainWin);
+      if (!mainWin->GetEventHandler()->ProcessEvent(copy))
```
oleg68:
```suggestion
      if (!pMainAppWin->ProcessWindowEvent(eCopied))
```

oleg68 (changes requested):

sebion7125:
Thanks for the review! Addressed:

- Anonymous namespace → `static` (project convention)
- Empty lines between variable declarations and following code (both files)
- Avoided the early return in `OnKeyCommand`: now uses `break` + an `isProcessed` flag
- Renamed `mainWin`/`copy` → `pMainAppWin`/`eCopied`, and switched to `pMainAppWin->ProcessWindowEvent(eCopied)` instead of `GetEventHandler()->ProcessEvent()`

Also rebased onto current `upstream/master` (this branch predated the cpptrace submodule integration in #2519) so the diff only shows the two relevant files.

oleg68:
@sebion7125 

1. Every user-visible change must ship with a changelog entry
2. Your code does not pass the pre-commit validation. Could you install pre-commit and run ``pre-commit run -a`` before committing changes?

sebion7125:
Both addressed:

- Installed pre-commit and ran `pre-commit run -a` — clang-format flagged the manually aligned columns in the shortcut table (my formatting, not project style); reformatted and now passes cleanly.
- Added a changelog entry for the user-visible fix (keyboard shortcuts not working when a detached panel window has focus).

sebion7125:
Done. 

Greetings from Germany ;)

Am 4. Juli 2026 20:42:15 MESZ schrieb Oleg Samarin ***@***.***>:
>oleg68 left a comment (GrandOrgue/grandorgue#2542)
>
>@sebion7125 
>
>1. Every user-visible change must ship with a changelog entry
>2. Your code does not pass the pre-commit validation. Could you install pre-commit and run ``pre-commit run -a`` before committing changes?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/GrandOrgue/grandorgue/pull/2542#issuecomment-4883435218
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

src/grandorgue/gui/panels/GOGUIPanelView.cpp
```diff
@@ -88,6 +89,27 @@ GOGUIPanelView::GOGUIPanelView(
...
+
+  if (pMainAppWin && topWindow != pMainAppWin) {
+    topWindow->Bind(wxEVT_CHAR_HOOK, [pMainAppWin](wxKeyEvent &e) {
+      wxKeyEvent eCopied(e);
+
+      eCopied.SetEventObject(pMainAppWin);
+      if (!pMainAppWin->ProcessWindowEvent(eCopied))
+        e.Skip();
+    });
+  }
```
oleg68: Suggest replacing the lambda with a named handler method plus an explicit `Bind`/`Unbind` pair, following the pattern already used elsewhere in the project (`GOSettingsPorts.cpp:81` + `~GOSettingsPorts()`), where a `Bind` on another object's window is always paired with an `Unbind` in the destructor.

**Reasoning:**
1. This lambda is the only place in `src/grandorgue` where `Bind` is called with anything other than `&Class::Method, this`. The rest of the codebase (`GOSettingsMidiDeviceList`, `GOSettingsPorts`) always uses a named handler.
2. There is no `Bind`/`Unbind` symmetry here — the binding lives as long as `topWindow` and is never explicitly removed. `wxEvtHandler` does clean up its own bindings on window destruction, but skipping the explicit `Unbind` departs from the pattern the project already uses for this exact kind of cross-window binding.
3. The handler doesn't need to store the main window pointer as a member at all: it can be looked up again via `wxTheApp->GetTopWindow()` inside the handler, which is the dominant pattern in this codebase (`GOSetter.cpp`, `GOGuiOrgan.cpp`, `GOEvent.cpp`, `GOSoundSystem.cpp` all call `wxTheApp->GetTopWindow()->GetEventHandler()->...` freshly each time, without caching the pointer). This also removes any question about the lifetime of a cached main-window pointer.

```suggestion
  /* Forward key events from detached panel frames to the main window so
     that keyboard shortcuts (e.g. Escape/Panic, Ctrl+S, F2-F12) work
     regardless of which panel has focus. GOAppWindow's OnKeyCommand
     returns without Skip() for recognized shortcuts, causing
     ProcessEvent() to return true; in that case we suppress e.Skip() so
     the panel widget does not also process the key (e.g. organ note).
     For unrecognized keys (organ notes, Shift) Skip() is called and
     GOGUIPanelWidget handles them normally. Panels embedded directly
     inside GOAppWindow are skipped because GOAppWindow's own
     EVT_CHAR_HOOK already covers them. */
  if (topWindow != wxTheApp->GetTopWindow())
    topWindow->Bind(wxEVT_CHAR_HOOK, &GOGUIPanelView::OnCharHook, this);
```

This suggestion only covers the local call site. It also needs:

```cpp
// GOGUIPanelView.h
private:
  void OnCharHook(wxKeyEvent &e);
```

```cpp
// GOGUIPanelView.cpp, destructor
GOGUIPanelView::~GOGUIPanelView() {
  if (m_TopWindow && m_TopWindow != wxTheApp->GetTopWindow())
    m_TopWindow->Unbind(wxEVT_CHAR_HOOK, &GOGUIPanelView::OnCharHook, this);
  ...
}

void GOGUIPanelView::OnCharHook(wxKeyEvent &e) {
  wxWindow *pMainAppWin = wxTheApp->GetTopWindow();
  bool isProcessed = false;

  if (pMainAppWin) {
    /* Copy the event: SetEventObject() below must not mutate the
       original e, because e is still routed further via e.Skip() to
       topWindow's own handlers (e.g. GOGUIPanelWidget playing a note)
       when the key is not a recognized shortcut, and those expect
       GetEventObject() to stay the panel/frame, not the main window. */
    wxKeyEvent eCopied(e);

    eCopied.SetEventObject(pMainAppWin);
    isProcessed = pMainAppWin->ProcessWindowEvent(eCopied);
  }
  if (!isProcessed)
    e.Skip();
}
```

This mirrors the `isProcessed` flag pattern already used in `GOAppWindow::OnKeyCommand` in this same PR, keeping a single `Skip()` call instead of two separate branches.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Fixed keyboard shortcuts (Panic, Help, Load/Open/Save/Install organ, MIDI player load) not working when a detached organ panel window has focus
```
oleg68: Please append the issue url to the line

sebion7125:
Thanks, makes sense — applied exactly as suggested:

- Replaced the lambda with a named `GOGUIPanelView::OnCharHook` method
- `Bind` in the constructor / `Unbind` in the destructor, matching `GOSettingsPorts`'s pattern (guarded by `m_TopWindow &&` since the constructor's own guard already skips the embedded-panel case, but the destructor needs its own null/self check)
- Main window looked up fresh via `wxTheApp->GetTopWindow()` inside the handler instead of captured
- Kept the `isProcessed` flag pattern for the single `Skip()` call, and added a comment explaining why the event must be copied (the original `e` is still routed onward via `Skip()` to the panel's own handlers, which expect `GetEventObject()` to stay the panel/frame)

sebion7125:
Added, thanks: https://github.com/GrandOrgue/grandorgue/issues/2541

oleg68 (approved):

oleg68:

rousseldenis (approved):

PR #2544, opened 2026-07-04
fix: off-by-one in compressed ring-buffer skip-ahead Seek()

sebion7125:
## Summary

- `StreamCacheReadAheadWindow::Seek()` (`GOSoundStream.cpp`) decides whether a freshly decoded sample should be stored with `if (r_cache.m_position >= index)`. But `DecompressionStep()` increments `m_position` *after* decoding, so the sample it just produced has index `m_position - 1`, not `m_position`.
- Using `>=` stores one sample too early on any skip-ahead `Seek()` where the cache starts behind the requested index — shifting every subsequent FIR tap in the polyphase resampler's read-ahead window by one, until it self-heals after `windowLen` (8) samples.

## Root cause

This is a frequent, everyday trigger, not an exotic edge case: `GOSoundAudioSection::Compress()` pins a precomputed cache snapshot at each loop start segment's `start_offset`, and after a loop wrap the resampling position typically lands a few samples further in (the wrap remainder) than that snapshot — a skip-ahead `Seek()` on an otherwise-fresh cache on *every* such wrap, for any compressed, looped sample.

```cpp
// Wrong — off by one: the sample just decoded has index (m_position - 1),
// so this fires one iteration too early:
if (r_cache.m_position >= index) { ... }

// Fixed:
if (r_cache.m_position > index) { ... }
```

## Test plan

- [x] Added `TestCompressedLoopWrapMatchesUncompressed` to the existing `GOTestSoundStream` suite: with `LINEAR` interpolation and factor 1.0 (identity), decoding is lossless, so a compressed and an uncompressed section with identical PCM data and loop layout must produce bit-identical `ReadBlock()` output for every frame, including immediately after a wrap. No assumption about "realistic" sample values is needed — any divergence is the bug.
- [x] Verified locally with a standalone harness against the real production `GOSoundCompressionCache`/`GOSoundResample` headers: with the `>=` bug, skip-ahead reads deviate from ground truth for exactly `windowLen` taps after every fresh `Seek()`, healing afterwards; with `>`, output is bit-identical to the uncompressed reference in all cases tested.
- [ ] New test not yet run in this local environment (toolchain lacks C++20 `<format>`, GCC 12); should pass in CI (`ubuntu-latest`).

### DIFF
--- a/src/grandorgue/sound/playing/GOSoundStream.cpp
+++ b/src/grandorgue/sound/playing/GOSoundStream.cpp
@@ -89,9 +89,13 @@ public:
       while (r_cache.m_position < readAheadIndexTo) {
         r_cache.DecompressionStep(nChannels, format16);
 
-        /* fill the read ahead buffer. If r_cache.position > index we assume
-          that the previous samples already present */
-        if (r_cache.m_position >= index) {
+        /* DecompressionStep increments m_position AFTER decoding, so the
+         * sample it just produced has index (m_position - 1). Store it only
+         * if that index is >= the requested index, i.e. m_position > index
+         * (not >=): on a Seek() to a fresh/skip-ahead position, using ">="
+         * stored one sample too early, shifting every subsequent tap in the
+         * window by one until it self-healed after windowLen samples. */
+        if (r_cache.m_position > index) {
           const int *pRead = r_cache.m_value;
 
           for (uint8_t i = nChannels; i > 0; i--)
--- a/src/grandorgue/sound/playing/GOSoundStream.cpp
+++ b/src/grandorgue/sound/playing/GOSoundStream.cpp
@@ -87,15 +87,16 @@ public:
       int *pWrite2 = pWrite1 + WINDOW_SAMPLES;
 
       while (r_cache.m_position < readAheadIndexTo) {
+        /* The sample about to be decoded has this index; DecompressionStep
+         * increments m_position afterwards, so m_position itself no longer
+         * identifies it. */
+        unsigned lastDecompressedPos = r_cache.m_position;
+
         r_cache.DecompressionStep(nChannels, format16);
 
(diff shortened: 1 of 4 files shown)


src/grandorgue/sound/playing/GOSoundStream.cpp
```diff
@@ -89,9 +89,13 @@ class GOSoundStream::StreamCacheReadAheadWindow
...
-        /* fill the read ahead buffer. If r_cache.position > index we assume
-          that the previous samples already present */
-        if (r_cache.m_position >= index) {
+        /* DecompressionStep increments m_position AFTER decoding, so the
+         * sample it just produced has index (m_position - 1). Store it only
+         * if that index is >= the requested index, i.e. m_position > index
+         * (not >=): on a Seek() to a fresh/skip-ahead position, using ">="
+         * stored one sample too early, shifting every subsequent tap in the
+         * window by one until it self-healed after windowLen samples. */
+        if (r_cache.m_position > index) {
```
oleg68: It is not obvious. It is better to remember r_cache.m_position to a temporary variable `lastDecompressedPos` before calling to r_cache.DecompressionStep

oleg68 (changes requested):

oleg68:
@sebion7125 Could you add a changelog entry?

sebion7125:
Thanks for the review! Both addressed:

- Capture `r_cache.m_position` into `lastDecompressedPos` before calling `DecompressionStep()`, so the store condition reads `lastDecompressedPos >= index` again (obvious, matching the original code's intent) instead of reasoning about `m_position - 1` in a comment.
- Added a changelog entry.

Re-verified with the same standalone harness against the real production headers — bit-identical output, no regressions.

oleg68 (approved):

oleg68:

rousseldenis (approved):

PR #2546, opened 2026-07-09
Moved Start/Stop orchestration from GOSoundSystem to GOAppWindow

oleg68:
When opening or closing sound and when Panic buttton is pressed, lots of GrandOrgue components are involved: GOSoundSystem, GOSoundEngine, GOAppWindow, GOOrganController.

Earlier GOSoundSystem was responsible for orchestration of all necessary actions between components. But GOAppWindow already has an orchestration code for other cases, so it is better to move the open/close sound orchestration there.

This PR:
- Adds GOSoundCloseListener: new interface for pre-close notification
- GOAppWindow: implements GOSoundCloseListener; added EnsureOrganStartedIfReady (builds+starts engine, calls ConnectToEngine+PreparePlayback) and EnsureOrganStopped (calls Abort+DisconnectFromEngine+StopAndDestroy); registers itself as close listener in constructor
- GOSoundSystem: added SetCloseListener/ConnectToEngine/DisconnectFromEngine/ IsOpen/GetAudioRecorder; removed m_IsRunning (use engine->IsUsed() instead); removed Start/Stop/NotifySound* helpers; AssignOrganFile now only creates/ destroys the engine object (no start/stop); AssureSoundIsOpen only opens ports; AssureSoundIsClosed calls listener before closing

This is just refactoring, no GO behavior should be changed.

### DIFF
--- a/src/grandorgue/gui/frames/GOAppWindow.cpp
+++ b/src/grandorgue/gui/frames/GOAppWindow.cpp
@@ -163,6 +163,7 @@ GOAppWindow::GOAppWindow(
     m_AfterSettingsEventType(wxEVT_NULL),
     m_AfterSettingsEventId(0),
     p_AfterSettingsEventOrgan(NULL) {
+  r_SoundSystem.SetCloseListener(this);
   SetIcon(get_go_icon());
 
   wxArrayString choices;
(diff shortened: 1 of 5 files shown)


oleg68:
@larspalo @rousseldenis Could you approve this PR? It is a base of lots of future work,

rousseldenis (approved):

PR #2548, opened 2026-07-13
Fix organ volume not saved into organ settings

### LINKED ISSUE #2547: Organ volume is no longer saved into the organ settings (regression from #2464)
**What happens:** Changing the volume with the toolbar spinner and then saving
the organ settings writes the volume the organ was *loaded* with, not the
current one. After reloading, the volume change is lost.

**Cause:** The class-move/rename commit from #2464 accidentally dropped a
functional line from `GODocument::SyncState()` when it became
`GOGuiOrgan::SyncState()`:

```cpp
m_OrganController->SetVolume(m_sound.GetEngine().GetVolume());
```

This line was the only path that ever wrote the live engine volume back into
the controller before saving (it had survived in this form since 2011, r339).
Without it, `GOOrganController::Save()` persists the stale value from load
time; `GOOrganController::SetVolume` now has no callers at all.

**Steps to reproduce:** Load an organ, change the volume spinner, save the
organ settings, reload the organ. The volume is back to its previous value.

**Minimal fix:** Restore the line at the top of `GOGuiOrgan::SyncState()`.

**Related design question:** Volume changes have never set the organ-modified
flag - the value was silently pulled at save time, so a volume-only change
never triggered the "save settings?" dialog (and was silently lost if the
user never saved for another reason). Is that the intended behaviour, or
should the volume spinner mark the organ as modified like transpose does?
Happy to adjust a PR either way.

oleg68:
## Summary
- Fixes #2547: changing the volume spinner and saving the organ settings persisted the volume the organ was loaded with, not the current one, because commit 26f2cdf2 (#2464) dropped the line that pulled the live engine volume back into `GOOrganController` before `Save()`.
- Instead of restoring that pull-back step in `GOGuiOrgan::SyncState()` (which should stay a GUI-only sync point), `GOOrganController::SetVolume()` is now the single entry point that updates the persisted `m_volume` field and pushes the value to the live sound engine, matching the existing `SetTemperament()`/`SetTranspose()` pattern already used for other organ attributes.
- `PreparePlayback()` now applies the cached volume to the engine as soon as it becomes attached, making the separate volume-set call after `Load()` redundant (removed).
- `GOAppWindow::OnSettingsVolume` now goes through `GOOrganController::SetVolume()` instead of touching the sound engine directly.

## How volume is persisted in the preset file

`GOOrganController` keeps the volume in its own `m_volume` member. It is loaded from the preset (CMB) file in `GOOrganController::Load()`:

```cpp
m_volume = cfg.ReadInteger(
  CMBSetting, WX_ORGAN, wxT("Volume"), -120, 100, false, m_config.Volume());
```

and written back to it in `GOOrganController::Export()` (called by both `Save()` and `Export()`):

```cpp
cfg.WriteInteger(WX_ORGAN, wxT("Volume"), m_volume);
```

Since `Export()` always reads straight from `m_volume`, saving is correct as long as `m_volume` itself is kept current. That's what this PR fixes: `SetVolume()` is now the single place that updates `m_volume` and forwards the value to the live sound engine (when attached), instead of the volume only being pushed to the engine at load time while `m_volume` went stale.

## Test plan
- [ ] Load an organ, move the volume spinner, confirm the audio level changes live
- [ ] Save the organ settings, reload the organ, confirm the volume spinner shows the saved value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed organ volume not saved into organ settings https://github.com/GrandOrgue/grandorgue/issues/2547
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
 - Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -861,9 +861,16 @@ void GOOrganController::PreconfigRecorder() {
   }
 }
 
+void GOOrganController::SetVolume(int volume) {
+  m_volume = volume;
+  if (m_soundengine)
+    m_soundengine->SetVolume(volume);
+}
+
 void GOOrganController::PreparePlayback(
   GOSoundOrganEngine *engine, GOMidiSystem *midi, GOSoundRecorder *recorder) {
   m_soundengine = engine;
+  m_soundengine->SetVolume(m_volume);
   m_midi = midi;
   m_MidiRecorder->SetOutputDevice(m_config.MidiRecorderOutputDevice());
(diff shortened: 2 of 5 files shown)


oleg68:
This bug has already fixed in another way with #2559, so the only thing to do is to add the changelog entry.

PR #2550, opened 2026-07-13
Fixed skipping the start of the release sample for silent-attack (BlankLoop) ranks https://github.com/GrandOrgue/grandorgue/issues/2521

### LINKED ISSUE #2521: SetupStreamAlignment skips start of release sample for silent-attack ranks (BlankLoop pattern)
## Summary

In organ sample sets that use a **silent loop** (commonly called \"BlankLoop\") as a dummy attack placeholder — with the actual key-action noise as standalone release files — `SetupStreamAlignment` / `ComputeTable` can cause the **beginning of the release sample to be skipped** at playback.

## Pattern in affected sample sets

Many Hauptwerk-derived sets (e.g. Obervellach) model key/stop action noises as:

```ini
Percussive=N
Pipe001=Noises/BlankLoop.wav       ; silent loop — nothing audible on key-down
Pipe001LoadRelease=N
Pipe001Release001=Noises/KeyNoise_R0.wav
Pipe001Release002=Noises/KeyNoise_R1.wav
```

The BlankLoop WAV contains only ±1 LSB values (essentially silence with quantisation noise).

## Root cause

`SetupStreamAlignment` computes `max_amplitude` / `max_derivative` as the maximum across the joinable attacks **and** the release itself. Because the release (noise WAV) dominates, the condition `(max_amplitude != 0) && (max_derivative != 0)` is always true, so `ComputeTable` runs.

`ComputeTable` scans the **release** WAV for positions whose amplitude/derivative falls within the range derived from the BlankLoop attack (amplitude ≈ 1 LSB). That match may be somewhere in the **middle** of the noise WAV, not at sample 0. At playback the BlankLoop's near-zero amplitude/derivative maps to that position → **the initial transient of the key noise is cut off**.

## Expected behaviour

For releases whose only joinable attacks are silent, `ComputeTable` should either be skipped or default to start position 0, so the release always plays from its natural beginning.

## Suggested fix

In `GOSoundAudioSection::SetupStreamAlignment`, check the amplitude of the joinable attacks separately from the release. If all joinables are essentially silent (max joinable amplitude < release amplitude / 1000, or below an absolute LSB threshold), skip `ComputeTable` and leave `m_ReleaseAligner` null — the release then starts at position 0 without any amplitude-based seek.

## Affected sample sets

Any set using the BlankLoop pattern for release noises. Common in professionally produced Hauptwerk-compatible sets.

oleg68:
## Summary
- Sample sets using a near-silent "BlankLoop" placeholder attack alongside a real release noise WAV caused `GOSoundAudioSection::SetupStreamAlignment` to build a release-alignment table from quantization noise, matching a position in the middle of the release and cutting off its true onset.
- `GOSoundProvider::ComputeReleaseAlignmentInfo` now excludes essentially-silent attacks (new `GOSoundAudioSection::IsEssentiallySilent()`) from the joinables passed to `SetupStreamAlignment`, so the existing empty-joinables guard leaves the release aligner null and playback falls back to the release's natural start position. `SetupStreamAlignment`/`ComputeTable` themselves are unchanged.

Resolves: https://github.com/GrandOrgue/grandorgue/issues/2521

## Test plan
- [x] Added `GOTestSoundStream::TestInitAlignedStreamWithSilentAttack`, verifying `IsEssentiallySilent()` for both a BlankLoop-like near-zero section and a normal section, and that the release aligner stays null when the only joinable attack is silent.
- [x] Built with `-DGO_BUILD_TESTING=ON` and ran `./bin/GOTestExe`: `GOTestSoundStream` passes. The only failing suite (`GOTestPerfSoundBufferMutable`) fails identically on unmodified `master` — pre-existing debug-build perf-baseline flakiness, unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fixed keyboard shortcuts (Panic, Help, Load/Open/Save/Install organ, MIDI player load) not working when a detached organ panel window has focus https://github.com/GrandOrgue/grandorgue/issues/2541
 - Fixed a rare glitch right after a loop wrap in compressed samples, caused by an off-by-one in the read-ahead ring buffer
 - Fixed crash on unknown command-line argument
+- Fixed skipping the start of the release sample for ranks with a silent-attack (BlankLoop) pattern https://github.com/GrandOrgue/grandorgue/issues/2521
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
 - Fixed expression pedal malfunction when CC value exceeds configured high value https://github.com/GrandOrgue/grandorgue/issues/2494
--- a/src/grandorgue/sound/playing/GOSoundAudioSection.h
+++ b/src/grandorgue/sound/playing/GOSoundAudioSection.h
@@ -28,6 +28,12 @@ class GOSampleStatistic;
 
 class GOSoundAudioSection {
 public:
+  /* A section whose max single-channel amplitude never exceeds this many raw
+   * PCM units (regardless of bit depth, since 1 LSB is always 1 in these raw
+   * units) is considered essentially silent, e.g. a "BlankLoop" placeholder
+   * attack containing only quantization noise. */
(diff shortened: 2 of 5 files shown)


oleg68:

oleg68:
I'm merging this pr because the reviewers do not answer.

PR #2553, opened 2026-07-15
Introduced GOOrganReader/GOLoadedOrganInfo to isolate ODF/CMB resolution

oleg68:
## Summary
- Extracts the archive/ODF/CMB resolution logic out of `GOOrganController::Load()` into a new `GOOrganReader` class (owning its own `GOConfigReaderDB`/`GOConfigReader` pair) and a `GOLoadedOrganInfo` struct holding the paths/hash/customization flag it discovers.
- `GOOrganController` now stores `GOOrgan m_ConfiguredOrgan` and `GOLoadedOrganInfo m_LoadedOrganInfo` instead of the separate `m_ArchiveID`/`m_ArchivePath`/`m_odf`/`m_SettingFilename`/`m_CacheFilename`/`m_ODFHash`/`m_b_customized` members.
- `GOOrgan` now self-normalizes its ODF path on construction (`SetODFPath()`), so `GOOrganReader` can take it as a plain in-parameter with no separate normalization step.
- This is the first step (PR A) of a larger plan to move GUI-related state out of `GOOrganController` into `GOGuiOrgan`; `Load()`/`ReadOrganFile()` orchestration itself is not split yet (that's a later PR).

Also fixes a latent bug in the shared test fixture `GOCommonControllerTest::setUp()`: it called `mkdtemp()` on an invalid template (`"."`, missing the required `XXXXXX` placeholder) stored in a `setUp()`-local array, so `organ_directory` was always null (and would have dangled after `setUp()` returned even had it succeeded). It now uses a member `std::string` buffer, which lets the new `GOTestOrganReader` test create real fixture files on disk.

## Test plan
- [x] `ctest` (both `GOTestFunctional` and `GOTestPerf`) passes
- [x] `./bin/GOTestExe` passes all 13 tests, including the new `GOTestOrganReader`
- [x] Full `make -k -j$(nproc)` build succeeds cleanly

### DIFF
--- a/src/core/GOOrgan.cpp
+++ b/src/core/GOOrgan.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright 2006 Milan Digital Audio LLC
- * Copyright 2009-2025 GrandOrgue contributors (see AUTHORS)
+ * Copyright 2009-2026 GrandOrgue contributors (see AUTHORS)
  * License GPL-2.0 or later
  * (https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).
  */
@@ -16,6 +16,7 @@
 #include "GOHash.h"
 #include "GOOrganList.h"
 #include "archive/GOArchiveFile.h"
+#include "go_path.h"
 
 GOOrgan::GOOrgan(
   const wxString &odf,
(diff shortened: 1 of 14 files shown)


oleg68:

oleg68:

PR #2554, opened 2026-07-16
Fix dangling GOConfig reference in GOCommonControllerTest

oleg68:
## Summary
- `GOCommonControllerTest::setUp()` constructed `GOOrganController` with a `GOConfig` local to `setUp()`, leaving `controller->m_config` dangling as soon as `setUp()` returned. Any test that exercises `Load()`/`Export()`/`Save()` through `controller` would hit undefined behavior. Store the `GOConfig` as a deferred-constructed member (`mp_config`, via `std::optional` since `GOConfig` has no default constructor and needs `GetName()`, only valid once the derived test's constructor has run) that outlives `controller`.
- `tearDown()` now actually `delete`s `controller` instead of leaking it, and resets `mp_config` afterward.
- `GOTestWindchest` registered `enclosure`/`rank` only via the non-owning `GOWindchest::AddEnclosure()`/`AddRank()` instead of the owning `GOOrganController::AddEnclosure()`/`AddRank()`, leaking both (and the rank's pipe) now that `controller` is actually destroyed.

## Test plan
- [x] `GOTestExe --no-perf`: all 11 functional tests pass
- [ ] Note: a pre-existing, separate leak remains and is out of scope here — `GOOrganController::m_setter` is always `new`'d in the constructor but only registered (and thus freed) via `m_elementcreators` inside `Load()`; tests that never call `Load()` (all `GOCommonControllerTest`-based tests) leak it. Tracked separately.

### DIFF
--- a/src/tests/common/GOTest.cpp
+++ b/src/tests/common/GOTest.cpp
@@ -38,16 +38,18 @@ bool GOCommonControllerTest::setUp() {
   GOTest::setUp();
   char path[] = ".";
   this->organ_directory = mkdtemp(path);
-  GOConfig settings(GetName(), "");
-  this->controller = new GOOrganController(settings);
+  mp_config.emplace(GetName(), "");
+  this->controller = new GOOrganController(*mp_config);
   this->controller->InitOrganDirectory(this->organ_directory);
   return true;
 }
 
 bool GOCommonControllerTest::tearDown() {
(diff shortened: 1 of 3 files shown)


oleg68:
@larspalo @rousseldenis could you approve this PR as soon as possible. Otherwise my new tests fail with segfault

rousseldenis (approved):

Issue #2523, opened 2026-06-12
Bug with a MIDI-Button REC + Incorrect Channel Interpretation When Playing MIDI Files


essemo:
Dear technical developers,

I would like to begin by thanking you for all the effort you put into developing GrandOrgue, free of charge.

I would like to report two small problems I have encountered.

1) Behavior of the MIDI REC button in the Record/Player panel.

I have noticed some issues:

I had configured several 32-byte Hauptwerk SysEx sends, specifically for:

Label-organ name,
Label-temperament,
Label-transpose,
Label-general bank,
Label-sequencer position

However, very strangely, when I click the MIDI-Button REC:

it triggers the sending of all the SysEx sends listed above.

It also triggers the sending configured for the Normal button and the sending configured for the Limiter button.

I noticed these SysEx transmissions thanks to KMidimon.

These issues only appear on the MIDI-Button REC "MIDI Recorder" but not on REC "AUDIO Recorder".

However, these SysEx transmissions are clearly causing issues...

2) Problem reading MIDI files: incorrectly identified channels.

Knowing that my ODF is correctly configured (MIDIInputNumber=1 for the foot controller, 2 for Manual I, 3 for Manual II, and 4 for Manual III);

When I load a MIDI file (for example, downloaded from https://viscountorgans.net/midi-file-downloads/),

there's an error because the notes received on channel 1 are transcribed onto Manual I instead of the Pedal,
the notes received on channel 2 are transcribed onto Manual II instead of Manual I,
and so on...

I verified this by viewing the MIDI event log:
all the notes from the MIDI file sent on channels 1, 2, and 3 are sent to the keyboards designated by channels 2, 3, and 4.
Therefore, there's a delay of '1'.

Thank you very much for your help!

Best regards, Jean-Bernard
Ubuntu Studio 24.04
grandorgue-3.17.2-1.x86_64 appimage
Ryzen 3900X

sebion7125:
Hi Jean-Bernard,

sorry that this issue did not get a reply earlier, and thank you for the detailed report. I had a look at the relevant code paths, and I think both observations are plausible.

[...]

PR #2557, opened 2026-07-18
Fixed MIDI REC resending external MIDI sends on recording start https://github.com/GrandOrgue/grandorgue/issues/2523

oleg68:
## Summary
- Starting MIDI recording broadcasts `GOEventDistributor::PrepareRecording()`, which makes every MIDI-sending object (buttons, labels, SysEx sends) resend its current value. This is required so the new recording starts from a known state and can be played back correctly, but it was also unexpectedly resending that state to external MIDI outputs (Hauptwerk-style SysEx labels, physical console panels/LCDs), even though nothing had actually changed. `AUDIO REC` does not exhibit this because it never reaches `PrepareRecording()`.
- `GOMidiSendProxy` now exposes two independent, positively-named routing flags: `IsToSendMidi()`/`SetToSendMidi()` (external MIDI ports) and `IsToRecordMidi()`/`SetToRecordMidi()` (the internal MIDI recorder stream), plus a `GOMidiSendProxy::StateRestorer` RAII helper to save/restore both.
- `GOOrganController::PrepareRecording()` now disables `ToSendMidi` (keeping `ToRecordMidi` enabled) only while broadcasting `PrepareRecording()`, so the initial state is still written into the `.mid` file but is not re-sent to external hardware. The direct `GOEventDistributor::PrepareRecording()` call made when starting/loading an organ is untouched, so external console synchronization on startup still works as before.
- This addresses only the "MIDI REC resends external sends" half of https://github.com/GrandOrgue/grandorgue/issues/2523. The separate standard-MIDI-file channel-mapping issue reported in the same issue is not addressed here.

## Test plan
- [x] `GOTestMidiSendProxy` unit test added (default state, independence of the two flags, `StateRestorer` behavior) — all 12 functional tests pass (`bin/GOTestExe --no-perf`)
- [x] Verified the same pre-existing ASan leak count appears on `master` without this change (unrelated to `GOTestWindchest`)
- [ ] Manual verification with a MIDI monitor (e.g. `aseqdump`) recommended: confirm no configured external send fires on MIDI REC, while the recorded `.mid` file still contains the initial GO setup/control state and plays back correctly; confirm organ startup still resyncs external console state

#2523 (partial — MIDI REC external sends only)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed MIDI REC resending external MIDI sends on recording start https://github.com/GrandOrgue/grandorgue/issues/2523
 - Fixed crash on unknown command-line argument
 - Fixed on-screen keyboard appearing when tapping organ panels on touch-enabled desktops https://github.com/GrandOrgue/grandorgue/issues/2492
 - Removed default keyboard shortcut assignments from MIDI objects; shortcuts must now be assigned manually in the MIDI event dialog
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -898,6 +898,17 @@ void GOOrganController::PrepareRecording() {
   m_MidiRecorder->SetSamplesetId(m_SampleSetId1, m_SampleSetId2);
   PreconfigRecorder();
 
+  // The broadcasted PrepareRecording() below makes every MIDI-sending object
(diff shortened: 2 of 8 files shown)


oleg68:

oleg68:

rousseldenis (approved):
LGTM

PR #2563, opened 2026-07-22
Added a choice of MIDI channel mapping schemes (pedal first, pedal last, or by the organ's MIDIInputNumber) when playing back MIDI files without a GrandOrgue setup header, configurable via a per-load dialog and MIDI Devices settings https://github.com/GrandOrgue/grandorgue/issues/2523

### LINKED ISSUE #2523: Bug with a MIDI-Button REC + Incorrect Channel Interpretation When Playing MIDI Files

oleg68:
Resolves: #2523 

## Summary
- Playing back a MIDI file that lacks GrandOrgue's own channel setup header previously assumed a fixed pedal-last channel scheme, which mismatches files recorded by other software (e.g. Viscount organs, which put the pedal on channel 1).
- Adds a "Choose MIDI channel mapping" dialog offering three schemes — pedal first (Viscount style), pedal last (old GrandOrgue style, default), or by the organ's MIDIInputNumber — shown when loading such a file.
- The dialog is skippable via a new "Ask channel mapping on MIDI file load" checkbox on the MIDI Devices settings page, with two configurable default schemes (with/without a usable MIDIInputNumber mapping), both defaulting to "Pedal last" for full backward compatibility.
- Help documentation updated to describe the new settings and dialog, with a link from the "Load MIDI" menu item help section.

Fixes part 2 of #2523 (part 1 was already fixed by #2557).

## Test plan
- [x] `GrandOrgue` and `GOTestExe` build clean (Debug, `GO_BUILD_TESTING=ON`)
- [x] New unit tests in `GOTestMidiPlayerContent` pass (`isMidiInputNumberMappingUsable`, `computeManualChannels` for all three mapping modes with/without pedal, `hasNativeHeader`)
- [x] Manually verified Settings → MIDI Devices layout and the "Choose MIDI channel mapping" dialog


oleg68:

src/grandorgue/midi/GOMidiPlayerContent.cpp
```diff
@@ -26,23 +29,53 @@ void GOMidiPlayerContent::Reset() { m_Pos = 0; }
...
+    const unsigned pedalChannel = isPedalFirst ? 1 : nManuals + 1;
+    unsigned nextManualChannel = isPedalFirst && hasPedal ? 2 : 1;
+
+    if (hasPedal)
+      channels.push_back(pedalChannel);
+    for (unsigned manualI = nManuals; manualI > 0; manualI--)
+      channels.push_back(nextManualChannel++);
   }
+
+  assert(channels.size() == midiInputNumbers.size());
```
rousseldenis, line 76: This makes me think to introduce (if it does not exist) specific exceptions like `GOError` or ... in order to manage the way they are displayed to the user. @oleg68 What do you think ?

oleg68: assert makes the program termination in debug builds and it is ignored in release builds. 

Normally the assert condition must be satisfied (and the code above guards it), so this termination should be considered as a bug.

rousseldenis: Oooh ok, understood.

rousseldenis (approved):
LGTM

Issue #2566, opened 2026-07-31
Windows: dbghelp refuses to load symbols from cv2pdb-patched exe/dll (no function names/lines in crash reports)


oleg68:
## Summary

On Windows, symbols for `GrandOrgue.exe`/`libGrandOrgueCore.dll` (and other cv2pdb-processed
binaries) cannot be resolved by `dbghelp.dll` at runtime, even though the `.pdb` produced by
`cv2pdb` sits right next to the matching binary and the CodeView (RSDS) debug directory entry in
the binary looks well-formed to tools like `objdump`.

This surfaced while adding a crash-report handler (`GOCrashHandler`) that
uses `cpptrace`'s dbghelp backend. On real Windows 11 (`dbghelp.dll` 10.0.22621.5471), a fatal
crash produces a stack trace with raw addresses and module names only — no function names or
file/line info — for frames inside `GrandOrgue.exe`/`libGrandOrgueCore.dll`.

## What was ruled out

- **Packaging order bug (fixed separately, unrelated to this issue):** `cv2pdb` writes its
  patched binary (with the CodeView record) to a throwaway `*-tmp` file that was never copied
  over the actually shipped `.exe`/`.dll`, so the shipped binary had an empty Debug Directory.
  This has been fixed in `cmake/BuildExecutable.cmake` / `cmake/BuildLibrary.cmake`. After the
  fix, the shipped binary's Debug Directory correctly points at the matching `.pdb`
  (verified with `objdump -p`), but the problem described here persists.
- **Embedded pdb path:** by default `cv2pdb` embeds the absolute path passed on its command line
  (in our CI/build-container setup, a Wine-internal path like
  `z:\mnt\host\home\...\GrandOrgue.pdb`, meaningless on the target machine). Tried `cv2pdb -p
  GrandOrgue.pdb` to embed a bare filename instead — no change in behavior.
- **`SLMFLAG_VIRTUAL`:** tried passing this flag to `SymLoadModuleExW` — no change (and it also
  breaks loading a known-good binary like `notepad.exe` in the same test harness, so it isn't
  applicable here anyway).

## Root cause (isolated)

Wrote a small PowerShell script that P/Invokes `dbghelp.dll` directly (`SymInitialize` +
`SymLoadModuleExW`) on `win11go` (real Windows, not Wine) to compare:

- `C:\Windows\System32\notepad.exe` (a normal MSVC-built PE): `SymLoadModuleExW` succeeds,
[...]

PR #2567, opened 2026-07-31
Added crash reporting for fatal signals/exceptions on all platforms

oleg68:
## Summary

- Adds `GOCrashHandler` (`src/core/`): a cross-platform fatal-error handler for crashes that
  cannot be caught as C++ exceptions (access violations, SIGSEGV, SIGBUS, ...). Without it, a
  crash in a release build leaves no diagnostics at all, and on Windows the process stays alive
  forever inside Windows Error Reporting instead of terminating — looking hung rather than
  crashed (this is what originally motivated the investigation, see
  https://github.com/GrandOrgue/grandorgue/issues/2566 for the underlying crash that started it).
- On installation (`GOGuiApp::OnInit`, right after `GOStdPath::InitLocaleDir()`) it installs a
  `sigaction`-based handler on Unix (SIGSEGV/SIGBUS/SIGFPE/SIGILL/SIGABRT, with an alternate
  stack so stack overflows are reported too) or `SetUnhandledExceptionFilter` on Windows, writes
  a `cpptrace` stack trace to stderr and to a report file under
  `<Documents>/GrandOrgue/CrashReports/`, then lets the process terminate immediately.
  `cpptrace` linkage moved to `GrandOrgueCore` (was previously duplicated per-executable) so all
  three executables (`GrandOrgue`, `GrandOrgueTool`, `GrandOrguePerfTest`) get crash reporting.
- Fixes a pre-existing Windows packaging bug in `cmake/BuildExecutable.cmake` /
  `cmake/BuildLibrary.cmake`: `cv2pdb` writes its patched exe/dll (with the CodeView record
  pointing at the matching `.pdb`) to a throwaway `*-tmp` file that was never copied over the
  actually shipped binary, so the shipped `.exe`/`.dll` had an empty PE Debug Directory and
  `dbghelp` could never even attempt to load the `.pdb`. Fixed by keeping the linker's raw output
  in the build tree (like the existing `OBJECT_FIXUP_REQUIRED`/macOS path already does) and
  having `cv2pdb` write its patched copy directly into `BINDIR`/`LIBDIR`.

## Known limitation (tracked separately)

Even after the packaging fix, Windows crash reports currently show raw addresses and module
names but not resolved function names/lines, because `dbghelp.dll` refuses to register
`cv2pdb`-patched modules for symbol lookup at all (isolated and written up in
[#2566](https://github.com/GrandOrgue/grandorgue/issues/2566); confirmed unrelated to this PR's
packaging fix or to the embedded pdb path). This does not affect Linux/macOS, where `cpptrace`
resolves DWARF/Mach-O directly, and does not affect the core benefit of this PR: the process no
longer hangs in WER on a fatal error.

## Test plan

- [x] Linux Debug build (`build/current`, `GO_BUILD_TESTING=ON`): builds clean, `ctest` — both
      `GOTestFunctional` and `GOTestPerf` pass.
- [x] Standalone crash test linked against `libGrandOrgueCore.so` (null dereference and stack
      overflow): process terminates with the original signal (core dump still produced), full
      symbolicated stack trace written to stderr and to
      `~/Documents/GrandOrgue/CrashReports/crash-<timestamp>-<pid>.txt`.
- [x] Windows cross-build (MinGW, official `build-scripts/for-win64/build-on-linux.sh`): builds
      and packages clean; verified via `objdump -p` that `GrandOrgue.exe` /
      `libGrandOrgueCore.dll` now carry a non-empty PE Debug Directory pointing at the matching
      `.pdb`.
- [x] Reproduced the original hang scenario on a real Windows 11 VM with the freshly built
      `GrandOrgue.exe`: launched without a loaded organ, triggered *Settings → Audio → Change
      audio device*. The process now terminates immediately instead of hanging in WER, and a
      crash report is written under `CrashReports`.


oleg68:

oleg68:

PR #2568, opened 2026-08-01
Moved audio callback processing from GOSoundSystem to GOSoundOrganEngine

oleg68:
## Summary

Previously, the double-buffered/multi-device audio callback synchronization logic (waiting for all output devices to fill their buffer before advancing to the next period, waking worker threads, etc.) lived in `GOSoundSystem`, in a private nested `GOSoundOutput` struct plus a set of atomic counters (`m_WaitCount`, `m_CalcCount`) and per-output mutex/condition pairs. `GOSoundSystem::AudioCallback()` directly manipulated this per-output locking/waiting state and called `GOSoundOrganEngine::GetAudioOutput()` / `NextPeriod()` / `WakeupThreads()` as separate steps.

After this PR, that synchronization logic has moved into `GOSoundOrganEngine` itself:
- `GOSoundSystem` now only owns the audio ports (`std::vector<std::unique_ptr<GOSoundPort>> mp_SoundPorts`, replacing the old `m_AudioOutputs` vector of `GOSoundOutput`) and calls a single new entry point, `GOSoundOrganEngine::ProcessAudioCallback(outputIndex, outBuffer)`, which fills the buffer for one output and, once all outputs have been filled in the current period, advances to the next period and wakes up the worker threads.
- `GOSoundOrganEngine` gained a private `OutputState` struct (task + mutex/condition + "already processed this period" flag) per output, and two per-period atomic counters (`m_NCallbacksEnteredCurrPeriod`, `m_NCallbacksFinishedCurrPeriod`).
- The engine's lifecycle gained a new `STREAMING` state (`IDLE -> BUILT -> WORKING -> USED -> STREAMING`), with a new `SetStreaming(bool)` method. `GOSoundSystem::ConnectToEngine`/`DisconnectFromEngine` now call `SetStreaming(true/false)` instead of directly resetting counters and manipulating output mutexes/conditions. `SetStreaming(false)` also broadcasts all output conditions so that any callback blocked waiting for the next period is unblocked immediately, fixing a deadlock that could otherwise occur when disconnecting while an output was mid-wait (xrun scenario).
- Reconnecting after a mid-period disconnect now correctly resets the per-period counters and per-output "already processed" flags, instead of carrying over dirty state that could deadlock a subsequent streaming session.
- `GOPerfTest` was adapted to call the engine through the same new API.

Net effect for callers: `GOSoundSystem` is now a thinner audio-port/backend wrapper, and all of the audio-thread synchronization/period-advancing logic (previously split across two classes) is consolidated in one place, `GOSoundOrganEngine`, alongside the rest of the engine's per-period processing.

## Tests added

New test suite `GOTestSoundOrganEngine` (with a shared `GOTestSoundOrganEngineBase` providing engine build/stop helpers), registered in `GOTestExe.cpp` / `CMakeLists.txt`:

- **Single-output lifecycle**: build → `SetUsed` → `SetStreaming` → 5 `ProcessAudioCallback` calls → `SetStreaming(false)` → `SetUsed(false)` → stop, checking state transitions throughout.
- **Multi-output lifecycle** (`TestTwoOutputsLifecycle`, `TestTwoGroupsTwoOutputsLifecycle`): 5 periods of calling `ProcessAudioCallback` for each output and checking that the period only advances once every output has been processed.
- **`SetUsed` transitions**: 3 cycles of WORKING↔USED, checking `IsWorking() && !IsUsed()` at each step.
- **Concurrent-callback stress tests**: 100 build/stop cycles with two threads concurrently calling `ProcessAudioCallback` for the same output (simulating an xrun), and 100 cycles across 8 engine configurations with one thread per output.
- **Deadlock regression test** (`TestDisconnectWithXrunDeadlock`): with 2 outputs, forces an xrun on output 0 so that its callback blocks; verifies `SetStreaming(false)` unblocks it so the disconnect wait completes without timing out.
- **Reconnect regression test** (`TestReconnectAfterMidPeriodDisconnect`): verifies that disconnecting mid-period and reconnecting resets the per-period counters/flags correctly, instead of deadlocking on the next streaming session due to leftover dirty state.

https://claude.ai/code/session_01WA6kJDFXW3FD4qW2TP2naq

### DIFF
--- a/src/grandorgue/sound/GOSoundOrganEngine.cpp
+++ b/src/grandorgue/sound/GOSoundOrganEngine.cpp
@@ -91,6 +91,16 @@ std::vector<GOSoundOrganEngine::AudioOutputConfig> GOSoundOrganEngine::
  * Constructors and destructors
  */
 
+GOSoundOrganEngine::OutputState::OutputState()
+  : condition(mutex), isFinishedCurrPeriod(false) {}
+
+GOSoundOrganEngine::OutputState::OutputState(OutputState &&other) noexcept
+  : mp_task(std::move(other.mp_task)),
+    condition(mutex),
+    isFinishedCurrPeriod(other.isFinishedCurrPeriod) {}
+
+GOSoundOrganEngine::OutputState::~OutputState() = default;
+
 GOSoundOrganEngine::GOSoundOrganEngine(
   GOOrganModel &organModel, GOMemoryPool &memoryPool)
   : r_OrganModel(organModel),
@@ -105,7 +115,9 @@ GOSoundOrganEngine::GOSoundOrganEngine(
     m_ReverbConfig(GOSoundReverb::CONFIG_REVERB_DISABLED),
     m_NSamplesPerBuffer(1),
     m_LifecycleState(LifecycleState::IDLE),
-    p_AudioRecorder(nullptr) {
+    p_AudioRecorder(nullptr),
+    m_NCallbacksEnteredCurrPeriod(0),
+    m_NCallbacksFinishedCurrPeriod(0) {
   SetVolume(-15);
   mp_ReleaseTask
     = std::make_unique<GOSoundReleaseTask>(m_SamplerPlayer, mp_AudioGroupTasks);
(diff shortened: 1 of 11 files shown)


oleg68:

oleg68:

PR #2569, opened 2026-08-01
Moved owning GOMidiSystem from GOSoundSystem to GOGuiApp

oleg68:
## Summary
- Moved ownership of `GOMidiSystem` from `GOSoundSystem` to `GOGuiApp`, so the sound subsystem no longer owns MIDI setup/lifecycle.

## Test plan
- [x] Builds cleanly (`GrandOrgue`, `GrandOrguePerfTest`)

### DIFF
--- a/src/grandorgue/GOOrganController.cpp
+++ b/src/grandorgue/GOOrganController.cpp
@@ -804,7 +804,8 @@ void GOOrganController::PreconfigRecorder() {
   }
 }
 
-void GOOrganController::StartOrgan(GOSoundSystem &soundSystem) {
+void GOOrganController::StartOrgan(
+  GOSoundSystem &soundSystem, GOMidiSystem &midi) {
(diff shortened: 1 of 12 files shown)


oleg68:

rousseldenis:
> * so the sound subsystem no longer owns MIDI setup/lifecycle.

@oleg68 Why was it a problem ? I think moving to GUI is not more appropriate, no ?

oleg68:
> > * so the sound subsystem no longer owns MIDI setup/lifecycle.
> 
> @oleg68 Why was it a problem ?

The Sound system and MIDI system are two independent systems. Earlier one of them owned another one that violated the principle of symmetry. This PR restores the symmetry between Sound and MIDI.

>  I think moving to GUI is not more appropriate, no ?

GOGuiApp is not a pure GUI class. It is a root container of other systems, including GUI and sound. Now it also starts owning GOMidiSystem instead of GOSoundSystem.

rousseldenis (approved):

PR #2578, opened 2026-08-10
Uniffied GOOrganController::Save()/Export() into a single Save(path)

oleg68:
Both did near-identical work (write core+GUI data to a file, differing
only in the destination path and whether the modified flag was reset
afterward). A single method with a default-valued path parameter covers
both cases; the file-write sequence is extracted into a static
write_config_file() helper. Prepares for moving GUI ownership out of
GOOrganController in a later step.


oleg68:
@larspalo @rousseldenis  Could you approve this PR?

rousseldenis (approved):

PR #2525, opened 2026-06-18
Refactored GOSoundOrganEngine API: AudioOutputConfig, lifecycle state, factory methods

oleg68:
## Motivation

`GOSoundOrganEngine` had a fragile lifecycle: configuration was spread across
multiple setters with implicit ordering requirements. Calling them out of order
caused subtle bugs — wrong audio output, silence, or crashes. There was no way
to tell from the API alone what state the engine was in or what had already
been configured.

This refactor makes the lifecycle explicit and the configuration self-contained.

## Changes

- Added `LifecycleState` enum (`IDLE/WORKING/USED`) with `IsIdle()`, `IsWorking()`,
  `IsUsed()`, `SetUsed()`, `SetFromConfig()` — engine state is now observable
  and transitions are controlled
- Extracted nested `AudioOutputConfig` struct into `GOSoundOrganEngine`
- Added static factory methods `createAudioOutputConfigs()` and
  `createDefaultOutputConfigs()` — callers no longer build config manually
- Changed constructor to take `GOOrganModel&` and `GOMemoryPool&` directly
- Changed `BuildAndStart` to accept `vector<AudioOutputConfig>`, `nSamplesPerBuffer`,
  `sampleRate` — all config passed at once, no partial state
- Renamed `bool` member fields to start with a question word (`m_IsDownmix`,
  `m_IsPolyphonyLimiting`, etc.) per coding conventions
- Adapted `GOSoundSystem` and `GOPerfTest` to the new API

## Test plan
- [ ] Build succeeds on Linux
- [ ] Audio playback works with existing organ definitions
- [ ] `GOPerfTest` runs without errors

PR #2592, opened 2026-08-21
Fixed stereo downmix left/right channel mixup with multiple audio groups

oleg68:
## Summary
- `GOSoundOrganEngine::BuildEngine()`'s downmix `scaleFactors` used a fixed stride of 4 per audio group, but `GOSoundOutputTask::DoRun()` reads them as a row-major matrix with row stride `nAudioGroups * 2`. The two only agreed when `nAudioGroups == 1`, so with 2+ audio groups the downmix's left and right outputs both ended up as `group0.left + group1.right` — a silent, doubled/wrong stereo image in recordings, not a crash.
- Introduced in #2525 ("Refactored GOSoundOrganEngine API: AudioOutputConfig, lifecycle state, factory methods", commit 764a5435), which added the downmix task using the `groupI*4` stride while `GOSoundOutputTask::DoRun()` already used the `nAudioGroups*2` row-major layout.
- Extracted the scaleFactors construction into `GOSoundOrganEngine::createDownmixGains()` (dB-valued, matching this class's existing gain units) and unified it with `createDefaultOutputConfigs()`, which independently implemented the same group-selection pattern — removing the duplication that let the two drift apart in the first place. Also extracted the shared dB→linear conversion into `convertGainToScaleFactor()`, used by both the downmix task and the per-device output loop.

## Test plan
- [x] `GOTestExe --no-perf`: all 20 functional tests pass, including new `GOTestSoundOrganEngine` tests (`TestDownmixGainsWithOneGroup/WithTwoGroups`, `TestDefaultOutputConfigs`, `TestDefaultOutputConfigsMatchesDownmixGains`)
- [x] Confirmed `TestDownmixGainsWithTwoGroups` fails with the pre-fix `groupI*4` indexing (reproduced the exact "L == R, both group0.left + group1.right" failure by reverting locally) and passes with the fix
- [x] Debug + ASAN build: no leaks/UB in the new/touched code (a dangling-reference bug in my own new test was caught and fixed this way)
- [ ] Manual smoke test with a real multi-audio-group organ and downmix-enabled recording (not available in this environment)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
https://claude.ai/code/session_01P1mBv1HHY4xvAEnNTM2XAw

### DIFF
--- a/src/grandorgue/sound/GOSoundOrganEngine.cpp
+++ b/src/grandorgue/sound/GOSoundOrganEngine.cpp
@@ -29,6 +29,29 @@
  * Factory functions
  */
 
+std::vector<float> GOSoundOrganEngine::createDownmixGains(
+  unsigned nAudioGroups) {
+  // GOSoundOutputTask::DoRun() reads the linear form of this (see
+  // convertGainToScaleFactor()) as a row-major [outChannelI][groupI * 2 +
+  // groupChannelI] matrix with row stride nAudioGroups * 2 (not a fixed 4,
+  // which only matches that row stride when nAudioGroups == 1): row 0 (L)
+  // takes each group's left channel (even groupChannelI), row 1 (R),
+  // starting at the row-1 offset nOutputCount, takes each group's right
+  // channel.
+  const unsigned nOutputCount = nAudioGroups * 2;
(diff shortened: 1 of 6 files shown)


oleg68:
@codex review

oleg68:

oleg68:

PR #2594, opened 2026-08-23
Added Eastern European translations

oleg68:
This PR adds UI translation for four more languages: Belarusian, Lithuanian, Russian and Ukrainian.

At first, I generated them automatically. Then I checked them manually.


rousseldenis (approved):
I approve even if I don't know languages 😄

PR #2600, opened 2026-08-26
Extract GOSoundTaskBase common round life cycle for sound tasks

oleg68:
## Summary
- Introduce `GOSoundTaskBase` as a shared implementation of the round life cycle (`Run()`/`CompleteRound()`/`NewRound()`) previously duplicated across all seven sound tasks, each of which kept its own mutex, "done" flag, and double-checked-lock preamble.
- Switch all seven `GOSoundTask` implementations (`GOSoundTremulantTask`, `GOSoundWindchestTask`, `GOSoundTouchTask`, `GOSoundOutputTask`, `GOSoundGroupTask`, `GOSoundReleaseTask`, `GOSoundRecorder`) to inherit from it, keeping only what is genuinely task-specific.
- Fix a latent bug where `GOSoundGroupTask` could be reset by `NewRound()` while a worker was still mixing outside its mutex, corrupting the round counter; `CompleteRound()` now waits for the task to become quiescent first, and `NewRound()` resets state before publishing `RUN_STATE_NOT_STARTED`.
- Split resource lifecycle (reverb engine reset, open recording file) out of `DiscardContent()` so tasks can answer `IsEmpty()`, and switch the scheduler's content contract from discard-on-register to discard-on-deregister, closing a gap where a deregistered task could hold stale content.

## Test plan
- [x] `make -k -j16` — builds clean
- [x] `./bin/GOTestExe --no-perf` — 23/23 tests pass, including new `GOTestSoundTaskBase` correctness and performance coverage (single- and multi-threaded, cooperative protocol)

Claude-Session: https://claude.ai/code/session_0164PtBRJai29FPXZ8EFLg2h

### DIFF
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -234,6 +234,7 @@ sound/tasks/GOSoundGroupTask.cpp
 sound/tasks/GOSoundOutputTask.cpp
 sound/tasks/GOSoundRecorderTask.cpp
 sound/tasks/GOSoundReleaseTask.cpp
+sound/tasks/GOSoundTaskBase.cpp
 sound/tasks/GOSoundTouchTask.cpp
 sound/tasks/GOSoundTremulantTask.cpp
 sound/tasks/GOSoundWindchestTask.cpp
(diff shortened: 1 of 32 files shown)


oleg68:
Closing — splitting into smaller PRs instead.

PR #2603, opened 2026-08-28
Renamed GOSoundFilter to GOSoundOnePoleFilter and relocated it to sound/

oleg68:
## Summary
- Pure rename+move, no logic changes: `sound/playing/GOSoundFilter.{h,cpp}` -> `sound/GOSoundOnePoleFilter.{h,cpp}`, `FilterType` -> `Type`.
- `GOSoundFilter` is performance-critical, per-sampler shared code used only by `GOSoundToneBalanceFilter`, but lived under `sound/playing/` as if it were sampler-specific. Moving it to `sound/` reflects that it's shared infrastructure, and sets up a follow-up change that turns it into a genuinely shared one-pole filter core reused by a future windchest shelf-EQ effect alongside the existing tone-balance filter.
- All call sites updated for the new name/path: `GOSoundToneBalanceFilter.h/.cpp`, `GOSoundSampler.h`. `CMakeLists.txt` path updated.

## Test plan
- [x] `cmake -DGO_BUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug ...` build succeeds
- [x] `./bin/GOTestExe` run: all functional tests pass (only unrelated performance-threshold benchmarks fail, calibrated for faster reference hardware)

### DIFF
--- a/src/grandorgue/CMakeLists.txt
+++ b/src/grandorgue/CMakeLists.txt
@@ -212,7 +212,6 @@ scheduler/GOSchedulerThread.cpp
 sound/interfaces/GOSoundCallbackConnector.cpp
 sound/playing/GOSoundAudioSection.cpp
 sound/playing/GOSoundFader.cpp
-sound/playing/GOSoundFilter.cpp
 sound/playing/GOSoundReleaseAlignTable.cpp
 sound/playing/GOSoundResample.cpp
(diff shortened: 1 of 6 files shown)


oleg68:
@codex review

oleg68:

rousseldenis (approved):

PR #2616, opened 2026-09-08
Made an audio group task quiescent before NewRound() resets it

oleg68:
## Summary
- GOSoundGroupTask mixed in ProcessList() outside m_mutex, so NewRound()'s mutex never excluded a worker still mixing from the previous period. That worker could then read a just-reset m_ActiveCount, underflow it to UINT_MAX, and permanently break "last one out" detection for the group.
- The race was silent because a group whose scale factors are zero in every output is skipped by GOSoundOutputTask, and mp_DownmixTask (which references every group) only exists when recording is configured.
- GOSoundGroupTask::CompleteRound() now waits (reusing the existing EnsureBufferReady(true) path), so every group is quiescent by construction once GOScheduler::CompleteRoundList() returns.
- GOSoundTaskBase::NewRound() now resets state before publishing RUN_STATE_NOT_STARTED rather than after, so DoNewRound() always observes the ending round whole.
- DoNewRound() now asserts the invariant instead of assuming it. GOSoundReleaseTask needed no fix; the commit records why in a comment.

## Test plan
- [x] `make -k -j16` (Debug, `GO_BUILD_TESTING=ON`) builds clean
- [x] `./bin/GOTestExe --no-perf` — all tests pass, including the new `TestCooperativeCompleteRoundWaitsForInFlightWork`, verified against a negative control (fails without the fix)


oleg68:
@codex review

oleg68:

rousseldenis (approved):

Issue #1563, opened 2023-06-05
[RFC] Add an API to call GrandOrgue


rousseldenis:
I'm wondering about what could be great to provide an API to control GrandOrgue from external tools from other means than MIDI messages.

The main advantage could be : 

- Provides list of features of the loaded sample set (e.g.: dynamically build stop controls interface, ...)

This because I was thinking how Hauptwerk-based consoles are managing their specialized touch interfaces for stops.

@oleg68 @larspalo 

The idea could be providing REST services when GrandOrgue is running.

larspalo:
@rousseldenis Care to elaborate on your exact motifs and what you really wish to achieve? I have absolutely no idea what you mean by how Hauptwerk-based consoles manages things... But I sure can tell that your main advantage point has never, ever, been a prime concern in the development of GO, nor is it really in line with the base goals of the software as such.

I'll elaborate myself on my last sentence. GO is not a general sample player software. It's highly specialized to allow virtual playing of pipe organ models. Most of the time we're talking (or should be talking) about digital models of real life organs. The general "user" of the software is mostly considered, and only required to be, a consumer of a product that a sample set producer (usually someone else) has created. That it's possible for an user to "edit" that product is a result mainly of the software being open sourced, not something that is generally encouraged.

I also sense a deviation from the main direction of (historic) GO development in the desire to move to other control means than MIDI messages. It might be useful to consider that the actual GUI of GO was always considered to be of somewhat "less" importance than the actual sound producing model as such. The main goal was always to be able to play the organ sounds of an instrument you most likely couldn't visit in real life, but still wanted to enjoy the experience of getting as close to as possible. The GUI image could certainly help enriching the experience, and somewhat increase the impression of being at the original console - playing the modeled organ - that's for granted, but it was never the "main" thing.

GO, as software, was always thought of as only part of that experience. Back in our minds as developers, we always thought that there would be a physical, MIDI driven, "organ like" console connected, from which the softwares sound production would be controlled. If that console is built DIY, or just using any existing MIDI capable digital instrument was not at all important - the development of GO tried to cater to all kind of needs (with extremely flexible and powerful MIDI possibilities/setup/detection etc).

The implication of prioritizing that the user (as a consumer) should have easily accessible means to visually alter the GUI representation of an organ model is that: "looks are more important than sound".

[...]

PR #2622, opened 2026-09-12
Feature: Grand Orgue Combination Setters Web UI :tada: 

FelipeFTN:
Hello folks! 

@oleg68 Thank you so much for this project, it's really awesome and it does so much for the organists community! :heart: 

Here is a quick contribution I seek for since a long time ago.
Happens that I don't have access to any MIDI Controller or something to control (pull or push) the stops in GO.
So, highly inspired by kovid goyal's [Calibre Library](https://github.com/kovidgoyal/calibre), I just added this feature that allows us to connect our mobile phones or external computers to Grand Orgue.
With this Web UI, we simulate a MIDI device directly from network. It's pretty great in my opinion and for me, very useful!

I had some help of LLMs, specially for writing good tests and UI.
The core implementation is mine.
I hope the code is properly good and respect the project's tradition and conventions. :+1: 

PS: Almost forgot to add some screenshots I took:

GrandOrgueMIDIGrandOrgueWebUI


oleg68:
Thank you, @FelipeFTN , for your contribution.

You bring the first WebUI example to the GrandOrgue code, so  we should discuss this feature before merging it?

- Should we provide a rest API at first (#1563) and implement a Web UI on top of it at the next? Or we should to add an android application for controlling GrandOrgue with a phone/tablet?
- Why only a sequencer and not other panels? We may implement only the sequencer now, but design the servlet path that we could easy add another panels in the future. Now this feature allocates the whole tcp port only for the sequencer that makes using the same port for other feature impossible. The urls like http:<host>:<port>/api/sequencer and http:<host>:<port>/ui/sequencer might be preferable.
- The sequencer has more controls. The web UI must cover all of them.

Please open a discussion at https://github.com/GrandOrgue/grandorgue/discussions .

FelipeFTN:
Hello @oleg68! It's an honor to contribute to this project! :pray: 

I really believe no discussion for these topics are needed for now, so I will keep it simple:

The WebRemote is already a very very simple implementation of HTTP server. It supports the multi-threading handling we are used to in the project, so I think it's pretty safe to use.
Since it's a HTTP server, if someone else wants to use its endpoints without using the already built-in web page, it's open to do it.
There is a very simple example in the code on how to integrate and use the API:
```
const hex = n => n.toString(16).padStart(2, '0');
function send(bytes) {
  fetch('/midi', { method: 'POST', body: bytes.map(hex).join(' '), keepalive: true })
    .then(r => dot.className = r.ok ? 'ok' : 'err')
    .catch(() => dot.className = 'err');
}
[...]
```
This is a very clean way, written in JavaScript at [src/grandorgue/resource/web-remote.html](https://github.com/GrandOrgue/grandorgue/pull/2622/changes#diff-df9539a6db5c51cb01b5c6fe90e2223e0af41f3401168a7f072a8f60e8489c96R105-R110) . :100: 

The HTML file looks pretty clean and lean for this simple task, and could be used as an example for any other implementation of a client for this Remote API feature.

Since it's the first version of a new feature, I think we could start really this simple; then, other PR features could be added on top of it. :sparkles: 
I don't think a REST API is really needed in this case, just a simple HTTP handler with a post request processor and some parsers did the job very well as you may see 😁

Other guides and docs could be build later on, to present different ways to create some alternative client and connect to GrandOrgue's HTTP server. This could be done in some discussion soon, or even in that one you mentioned. :+1: 

> Why only a sequencer and not other panels?

The good thing is: if there is something in GrandOrgue that can be assigned with "Listen to Event" option, this feature will work properly! Since it's made as an MIDI Device interpreter, this is already functional with all the other panels in GrandOrgue - I just tested hehe :tada: 
Give it a try, and tell me your thoughts! :pray: 

Thanks a lot again, @oleg68, I really hope this feature comes to upstream soon! :heart:

PR #2623, opened 2026-09-13
Finish moving GUI ownership from GOOrganController to GOGuiOrgan

oleg68:
## Why

`GOOrganController.h` still included headers from `gui/` and owned GUI state (panels, image cache, mouse state, main window data) — an unwanted graphics dependency in a model/controller class, getting in the way of reusing `GOOrganController` without a GUI (e.g. in tests and headless scenarios). This is the next step in the series of PRs moving GUI ownership out of `GOOrganController` into `GOGuiOrgan`.

## What changed

- `GOOrganController` no longer owns any GUI data at all: the temporary `LoadOrganGuiData()`/`ClearOrganGuiData()` split introduced in the previous PR is folded into `GOGuiOrgan::LoadOrganGui()`/`ClearOrganGui()`, which now directly own the image cache, panels, main-window data, and stop-window size (`mp_ImageCache`, `mp_MainWindowData` and the controller itself become `std::optional<T>`, matching the "owns, may be empty" role the previous raw pointers played).
- The controller now overrides `BeforeSave()`/`OnSave()`, calling `SyncState()`/`SaveOrganGui()`. Because `Save()` stays non-virtual in the base class and only calls these hooks through ordinary virtual dispatch, `GOSetter`'s direct `m_OrganController->Save()` call (with no `GOGuiOrgan` in scope) now reaches GUI-aware saving too — this also fixes a latent bug where "Save Settings" from the setter menu never synced the live main-window position/size before writing, since only the `GOGuiOrgan`-mediated `Save()` path used to call `SyncState()`.
- `GOGuiOrgan::Save()` is removed as a redundant forwarder now that `GOAppWindow` already caches a `GOOrganController*` it can call `Save()` on directly; `GOAppWindow`'s panel accessors switch from `p_OrganController->GetPanel()/GetPanelCount()` to `GOGuiOrgan`'s own.
- `GOOrganController.h` no longer includes anything from `gui/`.

## Testing

- `make -k -j16` — full build succeeds with no errors.
- `ctest` — all functional tests pass (only 2 speed-calibrated buffer perf tests fail, unrelated to this change).


oleg68:
@codex review

oleg68:

rousseldenis (approved):

PR #2631, opened 2026-09-21
Add test coverage for GOSoundOutputTask's reverb-tail IsEmpty() contract

oleg68:
## Summary
- PR #2617 fixed `GOSoundOutputTask::IsEmpty()` so it also checks `GOSoundReverb::HasContent()`, not just the meter - `GOSoundOrganEngine::NextPeriod()` calls `ResetMeterInfo()` every audio period regardless of deregistration, so the meter alone could read empty while the reverb engine still held an undecayed convolution tail.
- That PR's test coverage for `GOSoundOutputTask` didn't make it into the squash-merge. This PR adds it: per-channel mixing/clamping/metering tests, a reverb-tail test for `DiscardContent()`, and a test that a mere `ResetMeterInfo()` call must not make `IsEmpty()` report true while the reverb engine may still hold a tail.

## Test plan
- [x] `cmake -DGO_BUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug` build succeeds
- [x] `./bin/GOTestExe` run - `GOTestSoundOutputTask succeeded` (all 8 tests pass, including the 6 new ones)


oleg68:
@codex review

